PR #14, opened 2015-03-20
changed shebang line in executable python files 

bahaugen:
I was having trouble running the scripts as executables because it was using #!/usr/bin/python instead of the python in my PATH. This updates the shebang line to #!/usr/bin/env python for all of the executable python files.


lukeyeager:
Yes, that's the right way to do this. Thanks for the patch!

PR #16, opened 2015-03-21
Fixed issue#1: pretrained model selection bug

Sravan2j:
1) Removed the select field for previous networks snapshots from forms.py, due to wtforms validation issues. Then included code in views.py and new.html, to display the select fields for previous networks snapshots in html and for validating these fields. These changes fix the pretrained model selection bug. Also these changes ensure the display of error message, incase user selects the job that doesn't exists (this scenario happens when a user selects the job and at the same if another user deletes this job)
2) fix to the following issue: when validation fails, the selected radio buttons in "previous networks" tab remains selected but the select input field and customize button beside the selected previous network is in hidden status, which is wrong. same issue exists in standard networks tab. Modified new.html to ensure that select field and customize button beside selected radio button will be visible even when validation fails.
3) Modified the test case related to pretrained model selection.
4) Included the changes to display the selected "pretrained model" details in show.html file

### DIFF
--- a/digits/model/forms.py
+++ b/digits/model/forms.py
@@ -16,6 +16,13 @@ class ModelForm(Form):
         super(ModelForm, self).__init__(csrf_enabled=csrf_enabled, *args, **kwargs)
 
     ### Methods
+    def selection_exists_in_choices(form, field):
+        found=False
+        for i, choice in enumerate(field.choices):
+            if choice[0] == field.data:
+                found=True
+        if found == False:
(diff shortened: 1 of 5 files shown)


lukeyeager:
Thanks for the great work, @Sravan2j! Everything works great, I just added a few nitpicky comments.

lukeyeager:
Looks great, thanks for the fixes! Just squash all that into one commit and I'll merge it in.

Issue #15, opened 2015-03-21
[Suggestion]Add change the snapshot period item


AltenLi:
I'd like to add the textbox of snapshot_epoch for changing the snapshot period.
Or my disk is going to explode...
Would you tell me which file should I change?
(I've found the digits/model/forms.py to add the box)

AltenLi:
i've also changed "solver.snapshot = self.snap_epochs" in "digits/model/tasks/caffe_train.py"

AltenLi:
Also "self.snap_epochs = snap_epochs" in "digits/model/tasks/caffe_train.py" "__init__"

lukeyeager:
Hi @AltenLi. Yes, that's a feature that I was planning on adding. It looks like you've found all the right changes to make. If you've got it working, please send a pull request and I'll get it added into the master branch!

thatguymike:
I wonder if while we are at it, we should add an option to save the last N snapshots as well, with the default being all.

AltenLi:
It seems worked, but I wonder...is the unit of snapshot should be epoch or
iteration

On Tue, Mar 24, 2015 at 3:38 AM, Luke Yeager <notifications@github.com>
wrote:

> Hi @AltenLi <https://github.com/AltenLi>. Yes, that's a feature that I
> was planning on adding. It looks like you've found all the right changes to
> make. If you've got it working, please send a pull request and I'll get it
> added into the master branch!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/NVIDIA/DIGITS/issues/15#issuecomment-85161882>.
>



-- 
Best Regards,
------------------
李千
Qian Li
HUST
Mobile:+86-18511584115
Email: liqian115@gmail.com

AltenLi:
@lukeyeager I've updated the func for snapshot on my fork

PR #25, opened 2015-03-24
added snapshot selection

AltenLi:


lukeyeager:
Thanks for the pull request!

Can we make it a FloatField instead of an IntegerField? I think more people will be interested in taking snapshots *more frequently* than once an epoch, rather than less frequently. So we should allow them to input "0.1" for 10 snapshots/epoch, for example.

We may want to add more features in the future like only saving the last N snapshots (suggested in #15), or saving more snapshots in the first epoch (suggested in BVLC/caffe#1866). But **don't worry about any of that for now** - we'll figure it out after merging this PR.

AltenLi:
>Can we make it a FloatField instead of an IntegerField? I think more people will be interested in taking snapshots more frequently than once an epoch, rather than less frequently. So we should allow them to input "0.1" for 10 snapshots/epoch, for example.
Yes! I've changed into float.

>We may want to add more features in the future like only saving the last N snapshots (suggested in #15), or saving more snapshots in the first epoch (suggested in BVLC/caffe#1866). But don't worry about any of that for now - we'll figure it out after merging this PR.
It may change from caffe or delete by DIGITS, I'll try.

AltenLi:
oh... under linux... changed all \r\n.....

PR #26, opened 2015-03-25
Added snapshot_epoch setting with validation

AltenLi:
Fixed Last mass commit...
Sorry!

(Original PR here: #25)

### DIFF
--- a/digits/model/forms.py
+++ b/digits/model/forms.py
@@ -53,6 +53,14 @@ class ModelForm(Form):
             default=30,
             )
 
+    snapshot_epochs = wtforms.FloatField('Snapshot epochs',
+            default = 1,
+            validators = [
+                validators.DataRequired(),
+                validators.NumberRange(min=0),
+                ]
+            )
+
     val_interval = wtforms.IntegerField('Validation interval (in epochs)',
             validators = [
                 validators.NumberRange(min=0)
--- a/digits/model/images/classification/views.py
+++ b/digits/model/images/classification/views.py
@@ -129,6 +129,7 @@ def image_classification_model_create():
                     job_dir         = job.dir(),
                     dataset         = datasetJob,
                     train_epochs    = form.train_epochs.data,
+                    snapshot_epochs = form.snapshot_epochs.data,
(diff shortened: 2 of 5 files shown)


lukeyeager:
This PR fails a test:

    FAIL: test_model_download (digits.test_webapp.WebappTestCase)

It's because the snapshots are being saved by iteration instead of by epoch. I've had to jump through some hoops and fudge some math in the backend (as you've seen) to hide the idea of "iterations" from the user, but I want to continue to do that. In order to keep the iterations idea hidden, two things need to change:

1. Presenting the snapshot name as `Epoch#Iter127` is not acceptable. I vote for using `Epoch #0.1`, `Epoch #0.2`... It's ok if the numbers don't work out evenly for now - I can try to fix that with some clever rounding later. Just rounding to 3 sig-digs should be ok (e.g. `Epoch #0.333`).

2. Don't present an option for a snapshot every iteration - we don't want to talk about iterations, and I can't imagine anyone wanting to do that anyway. If 0 is allowed at all, it should mean "only save one snapshot after all of training has completed."

lukeyeager:
FYI, the tests will be run automatically as soon as #28 gets done.

AltenLi:
>Presenting the snapshot name as Epoch#Iter127 is not acceptable. I vote for using Epoch #0.1, Epoch #0.2... It's ok if the numbers don't work out evenly for now - I can try to fix that with some clever rounding later. Just rounding to 3 sig-digs should be ok (e.g. Epoch #0.333).

I've made the snapshot_epoch with "Just rounding to 3 sig-digs". And if is "1.0", I'll trunc it into "1".
And also passed all tests~ 

>Don't present an option for a snapshot every iteration

Yeah, U R right! I've deleted the mention of 0.

lukeyeager:
Whoops, I referenced the wrong PR# in [my commit](https://github.com/NVIDIA/DIGITS/commit/8abb8f679bdc2ae152dc5f2a149303dbfdc95c16). I rebased your changes against master in 76bcc01d16 and made some of my own in 8abb8f679b. Thanks for contributing!

Issue #36, opened 2015-03-27
Incorrect image sizes


drozdvadym:
Due to problem which was found in #29:
* the size of images, when "New Image Classification Dataset" are created, are not corresponds to "Mona Lisa" example (width and height are reversed).

Maybe is better to show some hint for sizes?

PR #38, opened 2015-03-28
fixed passing of incorrect sizes

drozdvadym:
fixed passing of sizes


drozdvadym:
corresponds to #36

lukeyeager:
Thanks for catching my mistake!

Issue #35, opened 2015-03-27
Parse arbitrary network outputs


lukeyeager:
Currently, DIGITS is hard-coded to look for three outputs (see [here](https://github.com/NVIDIA/DIGITS/blob/12e8b6041e1d8ddf6e256f9940d20d517fb673b8/digits/model/tasks/caffe_train.py#L391-L484)):

1. Train 'loss'
2. Test 'loss'
3. Test 'accuracy'

Instead, it should parse and save all train outputs and all test outputs, then decide how to graph them later. This is required for network architectures like GoogLeNet (see #11), which has multiple loss layers AND multiple accuracy layers.

Issue #34, opened 2015-03-27
Use a different js graphing library


lukeyeager:
See #9 and #31.

I can't download Google Charts and include it with DIGITS because of their [ToS](https://developers.google.com/chart/interactive/faq#offline). This means you have to be online to view the graphs.

Issue #11, opened 2015-03-20
Support for GoogLeNet


jmozah:
> GoogLeNet: bvlc_googlenet's train_val.prototxt says "Cannot specify more than one Accuracy Layer"

Hi

Tried to create a custom model and pasted the bvlc_googlenet's tain_val.prototxt... The initialization of the model failed saying the above error... Looks like a bug for me...

lukeyeager:
This is not a bug, it's expected behavior. But I'll admit it's a little annoying. If you click on the helptip next to "Custom Network" on the "Custom Network" tab, you'll see some more information about the current limitations in DIGITS. GoogLeNet does some really cool tricks in their network (See [this paper](http://arxiv.org/abs/1409.4842)

lukeyeager:
This is not a bug, it's expected behavior. But I'll admit it's pretty annoying. If you click on the helptip next to "Custom Network" on the "Custom Network" tab, you'll see some more information about the current limitations on creating custom networks DIGITS.

GoogLeNet does some really cool tricks in their network (See [this paper](http://arxiv.org/abs/1409.4842), page 7 for a visualization of the network structure). The problem is that the training network (train_val.prototxt) and the deploy network (deploy.prototxt) are drastically different.

Having two textareas for submitting a train_val network and a deploy network is cumbersome. In anticipation of [this change](https://github.com/BVLC/caffe/issues/1245) going into caffe, I just put one textarea in DIGITS and do some [ugly, tedious stuff](https://github.com/NVIDIA/DIGITS/blob/v1.0/digits/model/tasks/caffe_train.py#L104-L355) in the backend to split that one network into the two networks required by caffe.

Once that "All-in-one nets" feature comes out, you'll be able to use GoogLeNet and other interesting architectures, but for now you have to stick with the format used by LeNet, AlexNet and VGG (and many others).

lukeyeager:
Closing this bug as it is an intentional limitation for now.

ypyean:
@lukeyeager  Is there a simple way to import GoogLeNet net definition to DIGITS for model training now? Since we don't have idea when the "All-in-one nets" will be ready.

lukeyeager:
I'm reopening [and renaming] this issue since it's coming up a lot.

PR #41, opened 2015-03-31
GoogLeNet support

lukeyeager:
NOTE: Not ready for merge yet. See TODO's in commit messages.

This solves:

* #11 - You can run GoogLeNet, but all of the train/val layers will be included in the deploy network, so it's not very efficient.
* #35 - Needed for GoogLeNet
* #34 - Use C3.js instead of Google Charts. It's easiest to just do it now while I'm changing everything else anyway.


lukeyeager:
From above commit message:

> NOTE: GoogLeNet is still not fully supported

> 1. The deploy network is nearly the same as the train_val, so some unnecessary
computation will be done during inference with the outputted model

> 2. Inference within DIGITS doesn't even work yet, since
caffe.Classifier.predict uses the first output for inference instead of the
last one

1. That won't get fixed until BVLC/caffe#1245 happens.

2. I need to upgrade to a newer version of caffe anyway, and while I'm at it I'm going to start using the `Net` interface as suggested here: BVLC/caffe#1774. Once this is done, I'll merge in this branch and call it good enough for now.

lukeyeager:
Rebased against `upgrade-caffe` branch, which fixes this:

> Inference within DIGITS doesn't even work yet, since
caffe.Classifier.predict uses the first output for inference instead of the
last one

PR #58, opened 2015-04-09
Fix error reporting when deleting a dataset fails because of dependent models

Sravan2j:
1) Included DigitsError.py for including DIGITS custom exceptions
2) Created DeleteError custom exception
3) Changed scheduler.delete_job() function to display detail error message instead of default message.

### DIFF
--- a/digits/scheduler.py
+++ b/digits/scheduler.py
@@ -166,8 +166,8 @@ class Scheduler:
             job_id = job.id()
         else:
             raise ValueError('called delete_job with a %s' % type(job))
-        DependencyExists = False
-        ErrorDetails = ''
+        dependent_jobs = 0
+        error_details = ''
         # try to find the job
         for i, job in enumerate(self.jobs):
             if job.id() == job_id:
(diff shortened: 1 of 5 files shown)


lukeyeager:
Thanks, that 403 exception on delete_job has been pretty annoying and I've been meaning to get around to this. I do have a couple of nitpicky requests before I'll merge this.

digits/scheduler.py
```diff
@@ -166,7 +166,8 @@ def delete_job(self, job):
             job_id = job.id()
         else:
             raise ValueError('called delete_job with a %s' % type(job))
-
+        DependencyExists = False
+        ErrorDetails = ''
```
lukeyeager: Lower case variable names, please.

https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

digits/scheduler.py
```diff
@@ -175,7 +176,11 @@ def delete_job(self, job):
                     for j in self.jobs:
                         if isinstance(j, ModelJob) and j.dataset_id == job.id():
                             logger.error('Cannot delete %s (%s) because %s (%s) depends on it.' % (job.name(), job.id(), j.name(), j.id()))
-                            return False
+                            DependencyExists = True
+                            ErrorDetails += j.name() + ','
+                if DependencyExists:
+                    ErrorDetails = 'Cannot delete %s because jobs - %s depends on it' % (job.name(), ErrorDetails)
+                    raise errors.DeleteError(ErrorDetails)
```
lukeyeager: How about this message instead:

    Cannot delete "<dataset_name>" because <count> models depend on it: "<model1>", "<model2>", "<model3>"...

digits/test_scheduler.py
```diff
@@ -46,6 +47,9 @@ def test_add_remove_job(self):
         job = Job('tmp')
         assert self.s.add_job(job), 'failed to add job'
         assert len(self.s.jobs) == 1, 'scheduler has %d jobs' % len(self.s.jobs)
-        assert self.s.delete_job(job), 'failed to delete job'
+        try:
+            self.s.delete_job(job)
+        except errors.DeleteError as e:
+            raise AssertionError(e.__str__())
```
lukeyeager: No need to catch the error and raise an AssertionError - just let the DeleteError go through.

digits/utils/errors.py
```diff
@@ -0,0 +1,13 @@
...
+
+class DigitsError(Exception):
+    def __init__(self, value):
+        self.value = value
+    def __str__(self):
+        return self.value
+
+class DeleteError(DigitsError):
+    def __init__(self, value):
+        DigitsError.__init__(self, value)
```
lukeyeager: This is nice - I'll start using these in my code, too!

You don't really need any of that code in there. Just use this and treat them like normal exceptions:

```python
class DigitsError(Exception):
    pass
class DeleteError(DigitsError):
    pass
```

digits/views.py
```diff
@@ -119,10 +120,11 @@ def delete_job(job_id):
     job = scheduler.get_job(job_id)
     if not job:
         return 'Job not found!', 404
-    if scheduler.delete_job(job_id):
-        return 'Job deleted.'
-    else:
-        return 'Job could not deleted!', 403
+    try:
+        if scheduler.delete_job(job_id):
+            return 'Job deleted.'
```
lukeyeager: What happens if delete_job() returns False?

PR #83, opened 2015-04-20
Correction of tooltip for labels file

joyofdata:
If I interprete [titles.append(labels[int(key)]](https://github.com/NVIDIA/DIGITS/blob/master/digits/dataset/tasks/create_db.py#L243) correctly then the then numeric labels start at 0 and line numbers at 1.

### DIFF
--- a/digits/templates/datasets/images/classification/new.html
+++ b/digits/templates/datasets/images/classification/new.html
@@ -312,7 +312,7 @@ textfile_image_sets_disabled();
                                 <span name="textfile_labels_file_explanation"
                                     class="input-group-addon explanation-tooltip glyphicon glyphicon-question-sign"
                                     data-container="body"
-                                    title="The 'i'th line of the file should give the string label associated with the 'i'th numberic label."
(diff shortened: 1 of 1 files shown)


lukeyeager:
Thanks for contributing!

howtobeahacker:
hi @joyofdata,
I am still confused about numeric labels. Is string label for numeric label 0 any string we define or just number 0?
Thank you.

lukeyeager:
@howtobeahacker, you can use any string you like to represent a category. If your labels file is this:
```
dog
cat
lion
tiger
bear
```
Then you have 5 categories (0-4). When you test the network with an image, the raw output of the network might be:

    [0.3, 0, 0.5, 0.15, 0.5]

The resulting [sorted] output from DIGITS will be:
```
50% - bear
30% - dog
15% - tiger
5%  - lion
0%  - cat
```
You could just as easily use this for your labels file:
```
0
1
2
3
4
```
Or this:
```
1
2
3
4
5
```
In either case, the numbers will be interpreted as strings unrelated to the raw network output other than their ordering in the labels file.

howtobeahacker:
@lukeyeager Thanks. I got it.

PR #94, opened 2015-04-27
Add MacOS instructions

trivedigaurav:


lukeyeager:
Oh, I didn't realize you could just install gunicorn with pip. Duh. I'll move that into `requirements.txt` in an upcoming commit.

trivedigaurav:
:+1 Yes, that would be neater.

trivedigaurav:
I have updated the PR to include gunicorn in requirements.txt.

lukeyeager:
Thanks for contributing!

PR #142, opened 2015-06-10
Add travis support. Closed #28.

### LINKED ISSUE #28: Add travis CI integration
Contributors should be running the test suite themselves before submitting changes, but we should use Travis CI on GitHub anyway. Since Travis doesn't have GPUs on their build servers, we will just have to build caffe (and torch, theano, etc. in the future) in CPU-only mode. That won't give us full test coverage, but DIGITS should be as independent as possible from the backends anyway.

The main obstacle to overcome here is deciding how to clone and build caffe on the travis server. I'm considering adding caffe as a git submodule linking to the currently supported version. But this decision is related to the bigger issue of caffe integration mentioned in all of [these issues](https://github.com/NVIDIA/DIGITS/labels/caffe-version).

semisight:


lukeyeager:
Great work, thanks Alex!

I fixed the autodoc-related error in b139d77 and 10b580d.

When you fix the graphviz error, rebase against master to grab that change.

lukeyeager:
Closed in favor of #143.

PR #143, opened 2015-06-10
Add travis support (take 2). Closes #28.

### LINKED ISSUE #28: Add travis CI integration

semisight:
Take 2 on this. Please delete the old pull request.

Fixes graphviz build issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,45 @@
+language: python
+python:
+    2.7
+compiler: gcc # caffe
+sudo: required # for caffe
+env:
+    global:
+        - NUM_THREADS=4 # make
+        - WITH_CUDA=false # caffe
+        - WITH_CMAKE=true # caffe
+        - CAFFE_TAG="caffe-0.11"
+        - CAFFE_URL="https://github.com/NVIDIA/caffe.git"
+before_install:
+    - export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib
+    - export PATH=/home/travis/miniconda/bin:$PATH # caffe
+install:
+      # caffe
+    - pushd .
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,45 @@
...
+    - for req in $(cat caffe/python/requirements.txt); do pip install $req; done
+      # DIGITS
+      # use miniconda for some deps
+    - pushd .
+    - cd caffe
+    - conda install --yes gevent greenlet
+    - popd
+    - sudo apt-get install graphviz
+    - for req in $(cat requirements.txt); do pip install $req; done
+    - for req in $(cat requirements_test.txt); do pip install $req; done
```
lukeyeager: Why are you using `pip` for some dependencies and `conda` for others?

semisight: Ah! It speeds up build times somewhat with pre-built packages (by about a minute, but anything helps). See [here](http://sburns.org/2014/03/28/faster-travis-builds.html).

I probably should have split that into a new pull request however.

lukeyeager: No that's fine. I was just curious.

PR #172, opened 2015-07-19
Show ground truth label in 'Classify Images' - closes #110

### LINKED ISSUE #110: Classify Many Images doesn't show actual class label
Classify images doesn't show the "correct" label even if supplied.

Even though I can sue the valid.txt file which contains the correct labels:
image
as far as I can tell, the results dialog does not show the actual label allow with the forecast label.

gheinrich:


lukeyeager:
Looks great, thanks!

PR #183, opened 2015-07-27
Enable use of server files in dataset creation - closes #106

### LINKED ISSUE #106: Upload Text File has no Way of Specifying Path on Local System
The text file with all of the images is stored on the machine running DIGITS as opposed to the machine my browser is running on.
image

gheinrich:
Initial revision, need to run a few more tests before commit


lukeyeager:
1. Ooh, that path completion is awesome!

2. I was hoping for some sort of elegant solution like creating a new WTForms "LocalOrRemoteFile" field that generates the two HTML fields and a checkbox for you. Or at least some sort of widget/macro that lived in [helper.html](https://github.com/NVIDIA/DIGITS/blob/master/digits/templates/helper.html). Yours is kind of a brute force solution - creating two form fields and two HTML elements for each file. But I think it's OK for now. I wouldn't suggest actually creating a new WTForms field. Just rambling.

3. Do you think it would be better to allow the user to choose `upload` or `local` for each file rather than for all files all at once? I can't decide.

4. Needs tests!

5. You might want to try this out with the REST API. I've had trouble with `BooleanField` before. See [here](https://github.com/NVIDIA/DIGITS/blob/v2.0.0-rc3/digits/model/images/forms.py#L20-L23) and [here](https://github.com/NVIDIA/DIGITS/blob/v2.0.0-rc3/digits/dataset/images/classification/forms.py#L179-L182).

gheinrich:
Hi Luke, I have added a few tests in test_webapp.py to exercise the new route (for path completion) and the use of local files when creating a dataset. Those passed locally however I had an obscure failure on the Travis machine (when building Caffe?!?).

My feeling is it's better to keep things simple and use just one method for text files (i.e. either "all local" or "all remote" and not a combination of the two).

I don't understand your point about the use of `BooleanField`. In my case I have the field default to `False`, perhaps this is why I am not seeing a problem? I tried exercising the REST API and the `BooleanField` seems to be correctly interpreted (the server is looking for local files only if `textfile_use_local_files==True`:
```
gheinrich@android-devel-wks-7:/fast-scratch/gheinrich/ws/torch$ curl http://localhost:5000/datasets/images/classification.json -XPOS -F method=textfile -F textfile_use_local_files=True
{
  "errors": {
    "dataset_name": [
      "This field is required."
    ], 
    "textfile_local_labels_file": [
      "This field is required."
    ], 
    "textfile_local_train_images": [
      "This field is required."
    ]
  }
}
gheinrich@android-devel-wks-7:/fast-scratch/gheinrich/ws/torch$ curl http://localhost:5000/datasets/images/classification.json -XPOS -F method=textfile 
{
  "errors": {
    "dataset_name": [
      "This field is required."
    ], 
    "textfile_labels_file": [
      "This field is required."
    ], 
    "textfile_train_images": [
      "This field is required."
    ]
  }
}
```

gheinrich:
Tests passed on Travis machine on second attempt...

lukeyeager:
> I have added a few tests 

Awesome, thanks!

> My feeling is it's better to keep things simple and use just one method for text files (i.e. either "all local" or "all remote" and not a combination of the two).

Yeah ok.

> I don't understand your point about the use of BooleanField. In my case I have the field default to False, perhaps this is why I am not seeing a problem?

Oh right, I was only having problems because I wanted the field "on" by default.

PR #191, opened 2015-08-03
Add option to select min/max samples per class - closes #161

### LINKED ISSUE #161: Add options to choose min and max images per class
Add options to choose min images and max images per class when adding new dataset. This would ignore folders (classes) that have less than or more than given number of images. These fields could be added to dataset creation form, below validation set creation options. This would in many cases allow using of original dataset folder without additional preprocessing.

gheinrich:
I would like to add tests in test_webapp.py for this. However I understand we can't inspect the contents of a "ParseFolder" task using the REST API so it's slightly difficult to check whether the min/max values are taken into account.

### DIFF
--- a/digits/dataset/images/classification/forms.py
+++ b/digits/dataset/images/classification/forms.py
@@ -8,7 +8,7 @@ from wtforms import validators
 
 from ..forms import ImageDatasetForm
 from digits import utils
-from digits.utils.forms import validate_required_iff
+from digits.utils.forms import validate_required_iff, validate_greater_than
 
 class ImageClassificationDatasetForm(ImageDatasetForm):
     """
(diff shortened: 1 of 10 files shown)


lukeyeager:
Cool, thanks! Two requests:

1. Since the [selection](https://github.com/NVIDIA/DIGITS/blob/4400bef2ceb2176ac0984475f75994daa2ad831b/tools/parse_folder.py#L385-L398) is done before the [splitting](https://github.com/NVIDIA/DIGITS/blob/4400bef2ceb2176ac0984475f75994daa2ad831b/tools/parse_folder.py#L419) when parsing a folder, can we put the `min_per_category` and `max_per_category` HTML elements before the `pct_val` and `pct_test` split elements?

2. Can we add `min` and `max` options to the [`val_folder`](https://github.com/NVIDIA/DIGITS/blob/4400bef2ceb2176ac0984475f75994daa2ad831b/digits/dataset/images/classification/views.py#L49) and [`test_folder`](https://github.com/NVIDIA/DIGITS/blob/4400bef2ceb2176ac0984475f75994daa2ad831b/digits/dataset/images/classification/views.py#L60) parsing as well?

lukeyeager:
> I would like to add tests in test_webapp.py for this. However I understand we can't inspect the contents of a "ParseFolder" task using the REST API so it's slightly difficult to check whether the min/max values are taken into account.

Can we add some more detail to the `DatasetJob` JSON representation like we did for `ModelJob` [here](https://github.com/NVIDIA/DIGITS/blob/v2.0.0-rc3/digits/model/job.py#L37-L45)?

gheinrich:
Thanks Luke. I have implemented the changes you suggested and added a number of tests (code coverage is up slightly). I have rebased my changes on your   https://github.com/NVIDIA/DIGITS/pull/192.

lukeyeager:
Nice work, thanks for the tests!

I added some more comments inline.

lukeyeager:
LGTM.

PR #189, opened 2015-07-31
Add initial support for generic inference

lukeyeager:
Adds a new type of task to DIGITS - `Generic Inference`.

Solves #97, #117, #177 

DIGITS previously only supported "Image Classification," and made assumptions about the types of networks being used, the format of input data, and the way the output of the model should be interpreted.

The new task is more generalized, so you can do other things like object detection or per-pixel segmentation. The network can have one or more n-dimensional blobs, which DIGITS does not try to interpret in any way. I wrestled with a bunch of different names for this - `Regression`, `General-Purpose Networks`, `Multi-blob Output Models`, `Dense Prediction`, `Other Networks`, etc. None was quite accurate and simple enough, so I'm going with `Generic Inference`.

### Remaining limitations

This is much more generic, but DIGITS still puts some restrictions on what you can do with your network:

1. Only image data is supported
2. The only supported database format is still LMDB
  * So the only supported data layer type is still `Data`
3. Datasets consist of 1 or 2 LMDBs per phase (no more)

### TODO before merging

- [x] Add tests

### TODO after merging

* Create a standard input data format for creating datasets
  * **Currently, you have to create the LMDBs yourself [!]**

PR #192, opened 2015-08-04
Refactor test_webapp into several files

lukeyeager:
This gets rid of the "image_type" notion introduced in 4e48d71 that
required test_all functions with a bunch of yields.

Instead, I'm taking advantage of the fact that nose will run tests only
in classes that start with "Test". This lets me add a whole new set of
tests simply by adding a new class that defines a different value for
IMAGE_CHANNELS or CROP_SIZE, etc.

This will be helpful when creating tests for #189.


gheinrich:
That looks good to me!

Issue #71, opened 2015-04-15
The n best performing epoch should be display some place during train session


michael-george-hart:
After a long run of several hundred epochs it is often difficult eyeball the epoch performance graph to determine when and where the best epoch occurred.

I believer some place on the graph there should be at least the epoch for the best accuracy....
Additionally, it would be nice to have in the list box of epochs the associated performances..

michael-george-hart:
Here is a great reason why we all need to have access to accuracy of all the epochs somewhere on the TESTING page.... 
Today during my training I was able to eyeball an epoch, 210, which had a accuracy of 99.1%. At the end of the training, of 400 epoch, I wanted to see (eyeballing) the graphs for epoch with hopefully better accuracy; I discovered to my shock that epoch 210 was no longer a part of that performance plot being displayed; the epoch being displayed was 204, 208,212, 216 .... 

Granted, at the end of the training there were many epoch with accuracies of 98.98 being displayed.... However we all know that training is not deterministic no matter how sophisticated and robust a gradient decent algorithm being use.... 

Most people are looking for that outlier path found that are often lost during training. Then a later time we can go back fine tune our training parameters to focus on paths where the outlier was found in hopes of finding better solutions.

It is entirely by luck I happen to check the performance of the system at the time I did; allowing me to  see epoch 210.... Which make me wonder if there were any better performing epoch between 210 and 400 that is not being displayed by the performance graph..... after doing a grep (see below) on the of the log files I could not discover any based on my current understanding of the log files.

Just for the record in went in the .digits job directory for the particular job and did 

     > grep accuracy caffe_output.log | grep 99

and found 

      I0416 08:02:26.351547 14825 solver.cpp:315]     Test net output #0: accuracy = 0.990986

However, I have no idea how to associate this to the any of the epoch I see listed...

lukeyeager:
> I discovered to my shock that epoch 210 was no longer a part of that performance plot being displayed

Good point. I'm intentionally [clipping the graph data](https://github.com/NVIDIA/DIGITS/blob/7918631c7946a9dc637d27af1aa32ddda281a6af/digits/model/tasks/train.py#L435-L448) for performance reasons (see the "stride" variable). 

[...]

PR #195, opened 2015-08-05
Add results browser

semisight:
### Idea

This pull request adds a central page to view results from different ModelJobs. From there, jobs can be compared and filtered in a paginated table. I'm using DataTable, a jquery plugin that seems fairly widely used. The goal is to make finding the best network easier.

### Feedback

I'm looking for feedback on how to make this feature more useful. What columns should be added? How do users prefer to compare? Should there be a side-by-side comparison page?

Also, how should this be best integrated into digits? Currently, it is a green button on the front page that links to this. Is this ideal? Where/how should we link to this? Through the navbar?


lukeyeager:
I can't see the results yet (see the email I sent you), but from looking at the source this looks promising.

1 quick request: upload only minimized `.js` and `.css` files.

lukeyeager:
### Suggestions

* Move the sort icon next to the `th`
* Sort as numbers, not strings
* Show val_outputs
* If no value, sort to bottom of page (asc or desc)
* Sort by status
* Sort by time started
* Sort by runtime

Sorry that's a lot! Some should be easy, some may be hard.

lukeyeager:
Thanks for the fixes! Here are some more suggestions:

* Can we display the times as a readable string instead of a number of seconds?
* Can we add an explanation somewhere that tells whether you're showing the maximum or minimum or last value for each column? It's not clear to me.
* Can we show an arbitrarily large number of entries rather than choosing from 10,25,50,100?

And it looks like you haven't addressed these suggestions. Do you have reasons or just haven't gotten around to them yet?

* If no value, sort to bottom of page (either asc or desc)
* Sort by time started

semisight:
Point by point:

* What's your preferred time format? I've been working with seconds but I can make it a "H:M:S" format if that is preferable?
* I think the intent of the little glyph on the side is that if it is getting "smaller" as it goes down then it is descending, or if it gets "larger" going down it is ascending. I agree it could be clearer. Maybe make it bigger?
* I think that should be possible. I can turn off pagination, it's just on by default.

And as for the other two: I'm not quite finished yet. I'm still undecided on how best to override the default sorting per column. I believe sorting by ID should automatically sort by time started, but again, that's not exactly obvious. Maybe make another column for "start time?"

lukeyeager:
**Preferred time format**
I like this: https://github.com/NVIDIA/DIGITS/blob/v2.0.0-rc3/digits/utils/time_filters.py#L5-L16

**Maximum/Minimum/Last value**
Actually what I meant was that when I see "Accuracy" as a column, I don't know if the value in each row is the maximum accuracy for that job or just the last value.

semisight:
Preferred time format--okay. The runtime is a timedelta so I'd probably use the `print_time_diff` for that one.

Max/min/last--I can add a tooltip or comment above. All the values presented are from the last run. The assumption I'm making is that the final numbers are from the net that is saved by Caffe (unless it was aborted). I feel like it would be disingenuous to present an "all time minimum" because the current state of the net is not represented by that number.

lukeyeager:
I think many people will be interested in knowing which model had the best accuracy. See discussion at #71.

semisight:
Interesting. How about both?

lukeyeager:
Is there any way to view max/min/last for each output without creating 3x the number of columns? Some sort of radio button / select field for each column?

semisight:
I think for a first go, having a form to narrow down results would get us 90% of the way there with much less work. If this is a pain point we can always revisit it later.

lukeyeager:
@semisight, you said there's a way to view max/min/last now? I pulled the latest from `semisight/master` and I see the multi-select, but I don't see a way to sort by max accuracy.

Also, we should still display something by default when the page loads. Maybe the default columns and "loss" and "accuracy"?

semisight:
Yes, there is. It's pretty clunky right now. The multi select presents each column title 3 times: once with a "max" suffix, once with a "min" suffix, and once with no suffix. The max corresponds to the maximum value of that attribute, the min to the minimum value, and the "no suffix" to the latest value (last epoch).

I can add a default table in. Are loss and accuracy standard names? Is it okay to "bless" them, considering models aren't required to emit data under those names?

lukeyeager:
Since 9ff246eed47ec04461956b133495260855168e2e, all the standard networks output `loss` and `accuracy`. I think it's fine to show those by default to encourage standard naming on people's custom networks.

lukeyeager:
Can we add a simple page load test in these places:

https://github.com/NVIDIA/DIGITS/blob/v2.1.0/digits/test_views.py#L124
https://github.com/NVIDIA/DIGITS/blob/v2.1.0/digits/model/images/classification/test_views.py#L302
https://github.com/NVIDIA/DIGITS/blob/v2.1.0/digits/model/images/generic/test_views.py#L304

This looks pretty cool. Let's squash down to one commit and merge this thing, unless you have any other last-minute features you're trying to get done.

semisight:
Squashed everything. Here are some release notes, especially for features that aren't documented (yet).

### Show/hide columns:

Columns are grouped by 'type'. You can view the metadata like the name,
or view numeric values by latest value, min value, or max value.

### Column filters:

* Regex search on non-numeric columns
* Range search on numeric columns (use ':' to separate min and max
  values; omission of min or max substitutes -inf or +inf)

### Toggle filters:

Press 'Show/Hide Filters' to hide filters.

### Sorting and Multi-sorting:

All columns are sortable. Hold shift to sort by multiple columns.

lukeyeager:
Looks good and thanks for the release notes!

Once Travis passes, I'll merge ...

PR #200, opened 2015-08-07
Move autocomplete JS and CSS to top-level layout.html

gheinrich:


digits/templates/layout.html
```diff
@@ -46,6 +47,21 @@
...
 
+<style type="text/css">
+.autocomplete-suggestions { border: 1px solid #999; background: #FFF; overflow: auto; }
+.autocomplete-suggestion { padding: 2px 5px; white-space: nowrap; overflow: hidden; }
+.autocomplete-selected { background: #F0F0F0; }
+.autocomplete-suggestions strong { font-weight: normal; color: #3399FF; }
+.autocomplete-group { padding: 2px 5px; }
+.autocomplete-group strong { display: block; border-bottom: 1px solid #000; }
+</style>
+
```
lukeyeager: Can we move these into `digits/static/css/style.css`?

lukeyeager:
Should we make the class something like `.autocomplete_path` so we could add autocomplete for other things later (like `.autocomplete_job_id` or something)?

lukeyeager:
Can you also add this class to `custom_network_snapshot` on "classificationModel#new" and "genericModel#new" and to all the fields at "genericDataset#new"?

I was going to create a pull request to do this, but I would be stepping on your work here too much. I'll just let you do it.

gheinrich:
Hi Luke, thanks for the comments, I will propose another patch.

lukeyeager:
Looks great, thanks!

PR #202, opened 2015-08-07
Fixed broken image serving in development server

crohkohl:
Separation of changes in separate pull request: #199 

On my machine the image serving, e.g. of the mean.jpg does not work.
The browser (tested IE and Chrome) cannot interpret the image probably due to the missing content type. The send_file function takes care of that all.

### DIFF
--- a/digits/views.py
+++ b/digits/views.py
@@ -233,21 +233,7 @@ def serve_file(path):
     and this path will never be used
     """
     jobs_dir = config_value('jobs_dir')
-    path = os.path.normpath(os.path.join(jobs_dir, path))
-
-    # Don't allow path manipulation
-    if not os.path.commonprefix([path, jobs_dir]).startswith(jobs_dir):
-        raise werkzeug.exceptions.Forbidden('Path manipulation not allowed')
-
-    if not os.path.exists(path):
-        raise werkzeug.exceptions.NotFound('File not found')
-    if os.path.isdir(path):
-        raise werkzeug.exceptions.Forbidden('Folder cannot be served')
-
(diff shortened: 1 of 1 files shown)


lukeyeager:
Thanks for the fix! Looks like I was working too hard.

Can we go ahead and use `send_from_directory` while we're at it? Then we wouldn't need any of the checks for path manipulation.
http://flask.pocoo.org/docs/0.10/api/#flask.send_from_directory

Also, watch your extra whitespace on line 246.

crohkohl:
`send_from_directory` sounds even better! Will you change that or should I do it?
I am still getting used to that white-space thin in python. All me editors use tabs ... I will need to adjust them.

lukeyeager:
I can't push commits to your `fix_file_serving` branch - you'll have to do that. FWIW, it seems to be this simple:
```python
jobs_dir = config_value('jobs_dir')
return flask.send_from_directory(jobs_dir, path)
```

crohkohl:
okay, I put in your suggested change and squashed the two commits

lukeyeager:
I fixed the docs merged this into `master` with 63a8456f18a167b325fe87214969527e6462be90.

Thanks for the contribution!

crohkohl:
great!

crohkohl:
One question regarding this topic:

```python
 self.db = lmdb.open(self.output_path,
                    map_size=1000000000000, # ~1TB
                    map_async=True,
                    max_dbs=0)
```

This 1 TB is quite a lot which is allocated for each and every dataset. Should this be made configurable?

PR #203, opened 2015-08-07
Remove all references to LevelDB

lukeyeager:
@crohkohl partially addressed this in #199 for the sake of platforms where LevelDB isn't supported. This takes his partial solution to its logical extension. You can't actually use LevelDB with DIGITS anyway, so I think we should just go ahead and remove all references to it.

### DIFF
--- a/digits/dataset/tasks/create_db.py
+++ b/digits/dataset/tasks/create_db.py
@@ -30,7 +30,6 @@ class CreateDbTask(Task):
         resize_mode -- used in utils.image.resize_image()
         encoding -- 'none', 'png' or 'jpg'
         mean_file -- save mean file to this location
-        backend -- type of database to use
         labels_file -- used to print category distribution
         """
         # Take keyword arguments out of kwargs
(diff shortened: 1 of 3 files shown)


crohkohl:
One question regarding this topic:

```python
 self.db = lmdb.open(self.output_path,
                    map_size=1000000000000, # ~1TB
                    map_async=True,
                    max_dbs=0)
```

This 1 TB is quite a lot which is allocated for each and every dataset. Should this be made configurable?

lukeyeager:
> 1 TB is quite a lot

I agree it seems a bit ridiculous. Has it caused a problem for you?

It does cause problems on 32-bit systems:
https://github.com/BVLC/caffe/issues/1298
https://github.com/BVLC/caffe/issues/2293

And it may cause problems on non-linux platforms:
https://groups.google.com/forum/#!topic/caffe-users/0RKsTTYRGpQ

But it seems to be the consensus choice for 64-bit systems:
https://lmdb.readthedocs.org/en/release/#environment-class
https://github.com/BVLC/caffe/blob/eeb428e380/src/caffe/util/db_lmdb.cpp#L9

> Should this be made configurable?

That certainly seems reasonable.

crohkohl:
Well I tried to create the  mnist dataset on my machine and failed. I "only" had 1.5 tb of free space. But alone for 1 validation and 1 train set 2 TB of space are required.
I think it is unreasonable to assume that everybody can spare 2 TB for every imported dataset.
MNIST actually requires < 1 GB of memory but still would consume 2 TB.

danst18:
I have also been trying to port DIGITS to windows and I can confirm that the 1TB map size fails.

lukeyeager:
Branched the LMDB discussion out into #206 since it's not really relevant to this pull request.

PR #208, opened 2015-08-10
Move the tooltips and explanations from the html to the python.

jmancewicz:


digits/model/forms.py
```diff
@@ -175,14 +185,39 @@ def validate_lr_multistep_values(form, field):
...
+                        '    The deploy network will be created for you automatically. Once caffe implements their "all-in-one" nets feature, you will be able to specify which layers to include in the deploy file. Until then, these things are done automatically:\n'
+                        '    <ul>\n'
+                        '        <li>\n'
+                        '        The last <b>SoftmaxWithLoss</b> layer is replaced with a <b>Softmax</b> layer.\n'
+                        '        </li>\n'
+                        '        <li>\n'
+                        '        All other loss layers and accuracy layers are removed.\n'
+                        '        </li>\n'
+                        '    </ul>\n'
+                        '    </p>\n'
```
lukeyeager: Can we put this in a template and render it to a string like this?
https://github.com/NVIDIA/DIGITS/blob/v2.0.0-rc3/digits/task.py#L99-L104

digits/utils/dforms.py
```diff
@@ -0,0 +1,142 @@
+# Copyright (c) 2015, NVIDIA CORPORATION.  All rights reserved.
```
lukeyeager: I like using custom fields like this. Any reason we can't put these classes in `utils/forms`? What does `dforms` mean?

digits/model/forms.py
jmancewicz: Sure. I didn't like having html in the python.

digits/utils/dforms.py
jmancewicz: I can move it to utils/forms. dforms are digits specific subclasses of wtforms.

digits/templates/custom_network_explanation.html
```diff
@@ -0,0 +1,25 @@
...
+The deploy network will be created for you automatically. Once caffe implements their "all-in-one" nets feature, you will be able to specify which layers to include in the deploy file. Until then, these things are done automatically:
+<ul>
+    <li>
+        The last <b>SoftmaxWithLoss</b> layer is replaced with a <b>Softmax</b> layer.
+    </li>
+    <li>
+        All other loss layers and accuracy layers are removed.
+    </li>
+</ul>
+</p>
```
lukeyeager: Oh hey, this is actually wrong for GenericModelJobs. I should have created new text, but I lazily copied the same information from `classification/` to `generic/` in 93ad467735b70a5c175e9262e5d1da5c0ece88c1.

It should read something like:

### Specifying a custom network
You should enter a train_val network here. Some cleanup is done for you automatically:
* Data layers for the LMDBs in your dataset are created for you automatically
* If you provide a Data layer here and leave the `data_param.source` blank, it will be filled in for you with the appropriate path

The deploy network will be created for you automatically. Once caffe implements their "all-in-one" nets feature, you will be able to specify which layers to include in the deploy file. Until then, these things are done automatically:
* All loss layers and accuracy layers are removed from the deploy file
* Layers with names starting with `train_` are not included in the deploy file
* Layers with names starting with `deploy_

lukeyeager:
Looks great! One more nitpick:

Can we move `digits/templates/classification_custom_network_explanation.html` to `digits/templates/models/images/classification/custom_network_explanation.html` (alongside `new.html`)? Same for `generic_custom_network_explanation.html`.

Fix that and squash these into a single commit and I think this is good to go. Thanks for the great PR!

PR #209, opened 2015-08-11
Double LMDB map_size on MapFullError

### LINKED ISSUE #206: Set map_size for LMDB
*@crohkohl, @danst18, I'm breaking the discussion in #203 out into a new issue.*

Here's the situation as I understand it. Please correct me if any of this is wrong.

`map_size` | Linux | OSX & Windows
lower than size of dataset | LMDB runs out of memory | ?
higher than system memory | No problem | LMDB can't allocate enough memory

On Linux, you can just set it as high as you like and never see a problem. But that strategy blows up on other platforms.

> Should [map_size] be made configurable?
https://github.com/NVIDIA/DIGITS/pull/203#issuecomment-128859465

This is a sufficient but lazy solution. I would like to understand whether this can be avoided programmatically somehow before making a decision. My googling skills are failing me.

lukeyeager:
*Closes #206.*

### Changed
* Don't set the initial map_size to 1TB like before
  * It seems to start with 10MB by default
* When LMDB throws a MapFullError, double the map size
  * Big thanks to @dw for https://github.com/dw/py-lmdb/commit/442096e1551e204894154436877080a5b2571813

### TODO before merge
* [x] Add this change to `digits/dataset/images/generic/test_lmdb_creator.py` as well
* [x] Write some tests
* [x] Update requirements.txt to require newest version of LMDB

### DIFF
--- a/tools/create_db.py
+++ b/tools/create_db.py
@@ -35,19 +35,26 @@ except ImportError:
 
 logger = logging.getLogger('digits.tools.create_db')
 
-class DbCreator:
+class DbCreator(object):
     """
     Creates a database for a neural network imageset
     """
 
-    def __init__(self, db_path):
+    def __init__(self, db_path, lmdb_map_size=None):
         """
         Arguments:
         db_path -- where should the database be created
+
+        Keyword arguments:
+        lmdb_map_size -- the initial LMDB map size
         """
         # Can have trailing slash or not
         self.output_path = os.path.dirname(os.path.join(db_path, ''))
         self.name = os.path.basename(self.output_path)
+
+        if lmdb_map_size:
+            # convert from MB to B
+            lmdb_map_size <<= 20
 
         if os.path.exists(self.output_path):
             # caffe throws an error instead
@@ -55,7 +62,7 @@ class DbCreator:
             rmtree(self.output_path, ignore_errors=True)
 
         self.db = lmdb.open(self.output_path,
-                map_size=1000000000000, # ~1TB
+                map_size=lmdb_map_size,
                 map_async=True,
                 max_dbs=0)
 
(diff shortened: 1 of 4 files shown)


tools/create_db.py
```diff
@@ -448,7 +454,21 @@ def write_batch(self, batch):
...
+                key = '%08d_%d' % (keys[i], datum.label)
+                lmdb_txn.put(key, datum.SerializeToString())
+            except lmdb.MapFullError:
+                lmdb_txn.abort()
+
+                # double the map_size
+                curr_limit = self.db.info()['map_size']
+                new_limit = curr_limit*2
+                logger.debug('Doubling LMDB map size to %sMB ...' % (new_limit>>20,))
+                self.db.set_mapsize(curr_limit*2) # double it
```
gheinrich: since this is a new API in py-lmdb, is it worth adding a check that version is >0.86? I notice the version has not been updated yet in https://github.com/dw/py-lmdb/blob/master/lmdb/__init__.py but I guess this will come soon.

dw: Hi @gheinrich, I see the problem now. Somehow I committed the version bump to the release branch, not master. I've merged the single missing commit back to master. Thanks for catching!

dw: Just to be clear, the release branch and all released wheels, tarballs, git tags and GitHub tarballs will include the correct ``__init__.py``, the problem only affected master branch.

lukeyeager:
Ok, I've checked off all the TODOs. @crohkohl, could you run the test suite on Windows when you get a chance?

gheinrich:
This looks good to me.

crohkohl:
@lukeyeager Windows tests finally pass, see https://github.com/NVIDIA/DIGITS/pull/199#issuecomment-130214042

lukeyeager:
Squashed to a single commit and rebased to include TravisCI fix in #214.

lukeyeager:
@gheinrich and @crohkohl, thanks for the review help.

PR #217, opened 2015-08-13
Add dataset_id to ModelJob's JSON representation

lukeyeager:
Ahmed pointed out that the REST API doesn't expose any information about which `DatasetJob` a `ModelJob` depends on.


gheinrich:
That looks good to me. Is it worth adding a test (now or later)?

lukeyeager:
> Is it worth adding a test (now or later)?

Good idea. Added.

PR #218, opened 2015-08-14
Make alert box into a macro and try to prety up the field name

jmancewicz:
I tried to find a way to use the field.label in the error messages, but couldn't make that work.  I thought the next best thing would be pretty up the field name. So that crop_size becomes Crop Size.  Looks a little cleaner. Many of the field names are very close to the field.label.  Some are way off.

### DIFF
--- a/digits/templates/datasets/images/classification/new.html
+++ b/digits/templates/datasets/images/classification/new.html
@@ -1,6 +1,7 @@
 {# Copyright (c) 2014-2015, NVIDIA CORPORATION.  All rights reserved. #}
 
 {% from "helper.html" import print_flashes %}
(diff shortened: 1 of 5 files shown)


lukeyeager:
Here's how to get to `field.label.text`:

```jinja
{% for field_name, field in form._fields.iteritems() %}
{% if field.errors %}
<h1>{{ field.label.text }}</h1>
{{ field.errors }}
{% endif %}
{% endfor %}
```

semisight:
Really like this. I've been copying and pasting that chunk into new forms that I make.

jmancewicz:
Made the change to use the label rather than the capitalized field name.

PR #222, opened 2015-08-15
Show Create DB log in dataset view

gheinrich:


digits/dataset/tasks/create_db.py
```diff
@@ -53,9 +53,13 @@ def __init__(self, input_file, db_name, image_dims, **kwargs):
 
         self.entries_count = None
         self.distribution = None
+        self.create_db_log_file = "createdb_%s_output.txt" % db_name
```
lukeyeager: That's a little wordy. Would `create_%s.log` be too short?

digits/dataset/tasks/create_db.py
```diff
@@ -81,6 +85,10 @@ def __setstate__(self, state):
                 delattr(self, 'encode')
             else:
                 self.encoding = 'none'
+        if self.pickver_task_createdb <= 3:
+            print 'Upgrading CreateDbTask to version 4'
+            # nothing to do, version 4 adds field create_db_log_file
+            # but this field is ignored if None
```
lukeyeager: Do we really need a version bump here? Seems to me like we can make `create_db_log_file` an optional attribute for version 3.

digits/dataset/job.py
```diff
@@ -31,6 +31,9 @@ def json_dict(self, verbose=False):
                                       "val_count":   t.val_count,
                                       "test_count":  t.test_count,
                                       } for t in self.parse_folder_tasks()],
+                'CreateDbTasks':    [{"db_name":            t.db_name,
+                                      "create_db_log_file": t.path(t.create_db_log_file),
+                                     } for t in self.create_db_tasks()],
```
lukeyeager: This opens an interesting can of worms. Do you envision exposing all available files via the REST API eventually?

lukeyeager:
Great PR! This should definitely help with some of the confusion about DIGITS silently ignoring errors when creating datasets.

digits/dataset/images/classification/test_views.py
```diff
@@ -260,6 +260,21 @@ def check_textfiles(self, absolute_path=True, local_path=True):
...
+        info = self.dataset_info(job_id)
+        assert len(info['CreateDbTasks']) != 0
+        for task in info['CreateDbTasks']:
+            db_name = task['db_name']
+            assert db_name  and db_name != ''
+            path = task['create_db_log_file']
+            assert path and path != ''
+            assert os.path.isfile(path), "log file %s does not exist" % path
+        assert self.delete_dataset(job_id) == 200, 'delete failed'
+        assert not self.dataset_exists(job_id), 'dataset exists after delete'
```
lukeyeager: No need to explicitly delete the dataset. It will be deleted in `tearDownClass()`:
https://github.com/NVIDIA/DIGITS/blob/f3c4078d335fd8c2b36b829d0cf0258fc314e561/digits/dataset/images/classification/test_views.py#L82-L84

gheinrich:
Thanks for the feedback Luke! I have made the changes you suggested. I am not sure why code coverage went down. The tool reports a miss on lines https://github.com/gheinrich/DIGITS/blob/dev/createDbLog/digits/dataset/images/classification/test_views.py#L267-L274 even though they are part of a test that passed.

> This opens an interesting can of worms. Do you envision exposing all available files via the REST API eventually?

I leave this decision to you :-)

digits/dataset/job.py
lukeyeager: Can we just remove this section entirely for now? I'm thinking we might add a "verbose" option later which gives more info (and paths to all relevant files), but I don't want to add this and have to remove it later and break backwards-compatibility.

lukeyeager:
> I am not sure why code coverage went down

I've noticed that there are some lines in the scheduler which run sometimes and not others. It's not really a problem, but it makes our test coverage vary somewhat from run to run.

> The tool reports a miss on lines ... even though they are part of a test that passed

Yeah that's weird.

digits/dataset/tasks/create_db.py
```diff
@@ -53,9 +53,13 @@ def __init__(self, input_file, db_name, image_dims, **kwargs):
 
         self.entries_count = None
         self.distribution = None
+        self.create_db_log_file = "create_%s.txt" % db_name
```
lukeyeager: Can we use `.log` instead of `.txt`? That's what we have for the Caffe output.

digits/dataset/job.py
gheinrich: No problem. I'll need to remove the test that checks the file was created? OK?

digits/dataset/tasks/create_db.py
gheinrich: Yes, sorry I read the first comment too quickly. I used the .txt extension because in Firefox it shows the file automatically without asking the user to download/choose an external tool to open it. But you're right, it's better to keep consistency in the project.

digits/dataset/job.py
lukeyeager: Yep, lets do that for now.

Issue #197, opened 2015-08-06
Choosing data format[s] for creating generic inference datasets


lukeyeager:
[Branching discussion in #189 into separate thread]

>In order to create the dataset, is it possible to have the user specify [train|val|test].txt files in the form of:
/path/to/file [y1,...,yn]
It would be nice if DIGITS could create the image and label databases from these files (in theory that would allow the user to use non-image files too).
https://github.com/NVIDIA/DIGITS/pull/189#issuecomment-128347165

I'd like to enable at least three methods for people to create "Generic Inference" datasets.

1. By uploading prebuilt LMDBs (see discussion [here](https://github.com/NVIDIA/DIGITS/issues/97#issuecomment-120488919) about how many LMDBs to allow)

2. With textfiles and a list of floats as @gheinrich mentioned above

3. By parsing a folder
  * We could expected a `.npy` containing an n-dimensional label file for each image like so:
```
 images/
    ├── image1.png
    ├── image1.npy
    ├── image2.jpg
    └── image2.npy
```

lukeyeager:
More thoughts from @Pastafarianist at https://github.com/NVIDIA/DIGITS/issues/97#issue-72533530:

> Each line in such text file is currently matched against (.+)\s+(\d+)\s*$ (path/to/image 123). This could be replaced with (.+)((?:\s+\d+(?:\.\d*))+)\s*$ to check for a list of ints or floats (path/to/image 123 4. 5.67).

That's pretty much the same as option (2) above.

crohkohl:
Why not load caffe `datum` as defined in the protocol buffers and give it the file extension `*.cdatum`?

lukeyeager:
@crohkohl, thanks for the input! Are you suggesting replacing option (3), or creating another option?

I'd like to move away from Caffe-specific solutions where possible. You do bring up a good point, though, which is that `Datum`'s can only be 3-dimensional, whereas `numpy.ndarray`'s can be N-dimensional. So I guess we'd have to throw an error if you tried to supply a 4d label. That, or start supporting HDF5 in DIGITS.

crohkohl:
> Are you suggesting replacing option (3), or creating another option?

Well, I personally would choose a format that is independet from python for data import. I think a lot of people have tool chains other than python (like me). One could provide a custom protobuf format for example that could be exported from all supported languages.

[...]

PR #226, opened 2015-08-20
Add support for HDF5 datasets

### LINKED ISSUE #224: Support HDF5
I'm breaking the discussion about HDF5 out from #197 ...

https://github.com/BVLC/caffe/issues/2840 is a helpful summary. This is a more detailed comparison.

Are the benefits worth the effort?

## HDF5 vs. LMDB backends

 | HDF5 | LMDB
Compression | Whole dataset | Each datum
Compression options | [GZIP, LZF](http://docs.h5py.org/en/latest/high/dataset.html#lossless-compression-filters) | [Most image formats](http://docs.opencv.org/modules/highgui/doc/reading_and_writing_images_and_video.html#Mat imread(const string& filename, int flags))

## HDF5Data vs. Data layers

 | HDF5Data | Data
Data types | [N-D tensors](http://docs.h5py.org/en/latest/quick.html#core-concepts) | ["N<4"-D tensors](https://github.com/NVIDIA/caffe/blob/v0.12.2/src/caffe/proto/caffe.proto#L29-L31)
Transformations | [shuffle](https://github.com/NVIDIA/caffe/blob/v0.12.2/src/caffe/proto/caffe.proto#L531-L536) | [scale, mirror, crop, mean](https://github.com/NVIDIA/caffe/blob/v0.12.2/src/caffe/proto/caffe.proto#L345-L360)

## Other HDF5 pros/cons

* Pros
  * Not tied to any particular Caffe-specific format, so it would be easier to integrate with other frameworks
  * Can have many labels/blobs all in one file
  * Can use multiple files in one HDF5Data layer
* Cons
  * Doesn't have the memory-mapping optimizations that LMDB does

lukeyeager:
*Closes #224*

## TODO before merge

* [x] Create models from HDF5 datasets using `HDF5Data` layers
* [x] Expose backend and compression information in REST API
* [x] Shard HDF5 files into acceptable dataset sizes - https://github.com/BVLC/caffe/issues/2953#issuecomment-137274066

## TODO after merge

* Allow non-image data (see #197)
* Analyze prebuilt HDF5 datasets in "generic" path

### DIFF
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -140,8 +140,8 @@ class CaffeTrainTask(TrainTask):
         loss_layers = []
         accuracy_layers = []
         for layer in self.network.layer:
-            assert layer.type not in ['MemoryData', 'HDF5Data', 'ImageData'], 'unsupported data layer type'
-            if layer.type == 'Data':
+            assert layer.type not in ['DummyData', 'ImageData', 'MemoryData', 'WindowData'], 'unsupported data layer type'
+            if layer.type in ['Data', 'HDF5Data']:
                 for rule in layer.include:
                     if rule.phase == caffe_pb2.TRAIN:
                         assert train_data_layer is None, 'cannot specify two train data layers'
@@ -192,13 +192,22 @@ class CaffeTrainTask(TrainTask):
 
         train_val_network = caffe_pb2.NetParameter()
 
+        dataset_backend = self.dataset.train_db_task().backend
+
         # data layers
         if train_data_layer is not None:
-            if train_data_layer.HasField('data_param'):
+            if dataset_backend == 'lmdb':
+                assert train_data_layer.type == 'Data', 'expecting a Data layer'
+            elif dataset_backend == 'hdf5':
+                assert train_data_layer.type == 'HDF5Data', 'expecting an HDF5Data layer'
+            if dataset_backend == 'lmdb' and train_data_layer.HasField('data_param'):
                 assert not train_data_layer.data_param.HasField('source'), "don't set the data_param.source"
                 assert not train_data_layer.data_param.HasField('backend'), "don't set the data_param.backend"
+            if dataset_backend == 'hdf5' and train_data_layer.HasField('hdf5_data_param'):
+                assert not train_data_layer.hdf5_data_param.HasField('source'), "don't set the hdf5_data_param.source"
             max_crop_size = min(self.dataset.image_dims[0], self.dataset.image_dims[1])
             if self.crop_size:
+                assert dataset_backend != 'hdf5', 'HDF5Data layer does not support cropping'
                 assert self.crop_size <= max_crop_size, 'crop_size is larger than the image size'
                 train_data_layer.transform_param.crop_size = self.crop_size
             elif train_data_layer.transform_param.HasField('crop_size'):
@@ -211,36 +220,58 @@ class CaffeTrainTask(TrainTask):
             train_val_network.layer.add().CopyFrom(train_data_layer)
             train_data_layer = train_val_network.layer[-1]
             if val_data_layer is not None and has_val_set:
-                if val_data_layer.HasField('data_param'):
+                if dataset_backend == 'lmdb':
+                    assert val_data_layer.type == 'Data', 'expecting a Data layer'
+                elif dataset_backend == 'hdf5':
+                    assert val_data_layer.type == 'HDF5Data', 'expecting an HDF5Data layer'
+                if dataset_backend == 'lmdb' and val_data_layer.HasField('data_param'):
                     assert not val_data_layer.data_param.HasField('source'), "don't set the data_param.source"
                     assert not val_data_layer.data_param.HasField('backend'), "don't set the data_param.backend"
+                if dataset_backend == 'hdf5' and val_data_layer.HasField('hdf5_data_param'):
+                    assert not val_data_layer.hdf5_data_param.HasField('source'), "don't set the hdf5_data_param.source"
                 if self.crop_size:
                     # use our error checking from the train layer
                     val_data_layer.transform_param.crop_size = self.crop_size
                 train_val_network.layer.add().CopyFrom(val_data_layer)
                 val_data_layer = train_val_network.layer[-1]
         else:
-            train_data_layer = train_val_network.layer.add(type = 'Data', name = 'data')
+            layer_type = 'Data'
+            if dataset_backend == 'hdf5':
+                layer_type = 'HDF5Data'
+            train_data_layer = train_val_network.layer.add(type = layer_type, name = 'data')
             train_data_layer.top.append('data')
             train_data_layer.top.append('label')
             train_data_layer.include.add(phase = caffe_pb2.TRAIN)
-            train_data_layer.data_param.batch_size = constants.DEFAULT_BATCH_SIZE
+            if dataset_backend == 'lmdb':
+                train_data_layer.data_param.batch_size = constants.DEFAULT_BATCH_SIZE
+            elif dataset_backend == 'hdf5':
+                train_data_layer.hdf5_data_param.batch_size = constants.DEFAULT_BATCH_SIZE
             if self.crop_size:
+                assert dataset_backend != 'hdf5', 'HDF5Data layer does not support cropping'
                 train_data_layer.transform_param.crop_size = self.crop_size
             if has_val_set:
-                val_data_layer = train_val_network.layer.add(type = 'Data', name = 'data')
+                val_data_layer = train_val_network.layer.add(type = layer_type, name = 'data')
                 val_data_layer.top.append('data')
                 val_data_layer.top.append('label')
                 val_data_layer.include.add(phase = caffe_pb2.TEST)
-                val_data_layer.data_param.batch_size = constants.DEFAULT_BATCH_SIZE
+                if dataset_backend == 'lmdb':
+                    val_data_layer.data_param.batch_size = constants.DEFAULT_BATCH_SIZE
+                elif dataset_backend == 'hdf5':
+                    val_data_layer.hdf5_data_param.batch_size = constants.DEFAULT_BATCH_SIZE
                 if self.crop_size:
                     val_data_layer.transform_param.crop_size = self.crop_size
-        train_data_layer.data_param.source = self.dataset.path(self.dataset.train_db_task().db_name)
-        train_data_layer.data_param.backend = caffe_pb2.DataParameter.LMDB
-        if val_data_layer is not None and has_val_set:
-            val_data_layer.data_param.source = self.dataset.path(self.dataset.val_db_task().db_name)
-            val_data_layer.data_param.backend = caffe_pb2.DataParameter.LMDB
+        if dataset_backend == 'lmdb':
+            train_data_layer.data_param.source = self.dataset.path(self.dataset.train_db_task().db_name)
+            train_data_layer.data_param.backend = caffe_pb2.DataParameter.LMDB
+            if val_data_layer is not None and has_val_set:
+                val_data_layer.data_param.source = self.dataset.path(self.dataset.val_db_task().db_name)
+                val_data_layer.data_param.backend = caffe_pb2.DataParameter.LMDB
+        elif dataset_backend == 'hdf5':
+            train_data_layer.hdf5_data_param.source = self.dataset.path(self.dataset.train_db_task().textfile)
+            if val_data_layer is not None and has_val_set:
+                val_data_layer.hdf5_data_param.source = self.dataset.path(self.dataset.val_db_task().textfile)
         if self.use_mean:
+            assert dataset_backend != 'hdf5', 'HDF5Data layer does not support mean subtraction'
             mean_pixel = None
             with open(self.dataset.path(self.dataset.train_db_task().mean_file),'rb') as f:
                 blob = caffe_pb2.BlobProto()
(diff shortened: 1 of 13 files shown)


lukeyeager:
I've now integrated HDF5 support into the dataset-creating side of DIGITS so as to be able to run some speed and size tests (see original tests at #224).

* MNIST test imageset
  * 10,000 images
  * 28x28 grayscale
* ImageNet-like imageset
  * 3,676 images
  * 256x256 color

Backend | Image Encoding | Database Compression | MNIST size | MNIST time | ImageNet size | ImageNet time
------------ | ------------- | ------------- | ------------- | ------------- | ------------- | -------------
LMDB | | | 59 | 60 | 704 | 20
LMDB | PNG | | 22 | **56** | 323 | 73
LMDB | JPEG (lossy) | | 57 | 57 | **66** | 21
HDF5 | | | 46 | 59 | 697 | **18**
HDF5 | | LZF | 14 | 59 | 618 | 24
HDF5 | | GZIP | **11** | 60 | 551 | 45

*Sizes are in MB and times are in seconds.*

gheinrich:
Luke, do I understand correctly that those are the times to create a database? Do you think it would be interesting to also measure how long it takes to load a batch from the database (when the batch is cold in cache and when it is hot in cache)? Just to verify that loading the samples from the database is comfortably faster than the typical processing that follows.

lukeyeager:
Yes, I'll post that information when I can. I have to knock out that last TODO before I can integrate HDF5 datasets into networks, and it's taking more work than I had hoped initially.

lukeyeager:
Update: Caffe doesn't actually support integer data in HDF5 files (see BVLC/caffe#2978).

Rather than make a change to Caffe and require DIGITS users to update to the newest version, I'll simply force float data for now and take the hit on filesize.

lukeyeager:
Update: Caffe doesn't support the LZF compression filter - only raw and GZIP are supported.

```
HDF5-DIAG: Error detected in HDF5 (1.8.11) thread 140259111430720:
#000: ../../../src/H5Dio.c line 182 in H5Dread(): can't read data
major: Dataset
minor: Read failed
#001: ../../../src/H5Dio.c line 550 in H5D__read(): can't read data
major: Dataset
minor: Read failed
#002: ../../../src/H5Dchunk.c line 1837 in H5D__chunk_read(): unable to read raw data chunk
major: Low-level I/O
minor: Read failed
#003: ../../../src/H5Dchunk.c line 2868 in H5D__chunk_lock(): data pipeline read failed
major: Data filters
minor: Filter operation failed
#004: ../../../src/H5Z.c line 1150 in H5Z_pipeline(): required filter 'lzf' is not registered
major: Data filters
minor: Read failed
#005: ../../../src/H5PL.c line 293 in H5PL_load(): search in paths failed
major: Plugin for dynamically loaded library
minor: Can't get value
#006: ../../../src/H5PL.c line 397 in H5PL__find(): can't open directory
major: Plugin for dynamically loaded library
minor: Can't open directory or file
F0826 16:08:40.263033   752 io.cpp:268] Check failed: status >= 0 (-1 vs. 0) Failed to read float dataset data
```

tools/create_db.py
```diff
@@ -24,6 +26,7 @@
 import numpy as np
 import PIL.Image
 import lmdb
+import h5py
```
gheinrich: does this need to be added to requirements.txt?

digits/model/tasks/caffe_train.py
```diff
@@ -203,36 +212,58 @@ def save_files_classification(self):
...
+            train_data_layer.data_param.backend = caffe_pb2.DataParameter.LMDB
+            if val_data_layer is not None and has_val_set:
+                val_data_layer.data_param.source = self.dataset.path(self.dataset.val_db_task().db_name)
+                val_data_layer.data_param.backend = caffe_pb2.DataParameter.LMDB
+        elif dataset_backend == 'hdf5':
+            train_data_layer.hdf5_data_param.source = self.dataset.path(self.dataset.train_db_task().textfile)
+            if val_data_layer is not None and has_val_set:
+                val_data_layer.hdf5_data_param.source = self.dataset.path(self.dataset.val_db_task().textfile)
         if self.use_mean:
+            assert dataset_backend != 'hdf5', 'HDF5Data layer does not support mean subtraction'
```
gheinrich: Is it a temporary limitation?

lukeyeager: Somebody needs to either add this functionality to the HDF5 layer or create a new "transformation" layer that still runs in the "prefetch" phase. See https://github.com/BVLC/caffe/issues/2840.

My current solution is to just throw an error from the DIGITS side which is more legible than a Caffe error.

tools/create_db.py
lukeyeager: Yep, good point.

lukeyeager:
### Experiment 1

* MNIST dataset
  * 60,000 images in 10 classes
  * 28x28 grayscale
* LeNet network
  * 3 epochs

Backend | Image encoding | Database compression | Dataset filesize (MB) | Dataset creation time (sec) | Model training time (sec)
------------ | ------------- | ------------- | ------------- | ------------- | -------------
LMDB |  |  | 59 | 60 | 54
LMDB | PNG |  | **22** | 56 | 53
LMDB | JPEG (lossy) |  | 57 | 57 | 54
HDF5 |  |  | 181 | 62 | **19**
HDF5 |  | GZIP | **17** | 62 | **18**

### Experiment 2

* ImageNet-like dataset
  * 3,676 images in 2 classes
  * 256x256 color
* AlexNet network
  * 3 epochs

Backend | Image encoding | Database compression | Dataset filesize (MB) | Dataset creation time (sec) | Model training time (sec)
------------ | ------------- | ------------- | ------------- | ------------- | -------------
LMDB |  |  | 704 | **20** | 60
LMDB | PNG |  | 323 | 73 | 60
LMDB | JPEG (lossy) |  | **66** | **21** | 60
HDF5 |  |  | 2800 | **23** | **31**
HDF5 |  | GZIP | 772 | 88 | **33**

## Results

1. HDF5 has a lot of filesize bloat (see https://github.com/BVLC/caffe/pull/2978)

2. `HDF5Data` seems to be about 2x faster than `Data` for data reads (!)
  * I'm going to run some more experiments to figure out what's going on here

lukeyeager:
New problem: I have to cap the HDF5 files at no more than INT_MAX numbers per dataset.

See https://github.com/BVLC/caffe/issues/2953#issuecomment-137267945

lukeyeager:
I suspect the reason HDF5Data is so much faster than Data is [ironically] because it loads the whole dataset at once (BVLC/caffe#2892) instead of prefetching.

lukeyeager:
I think I've knocked out all the gotchas. And I put in a warning box explaining what is not currently supported:
hdf5-warnings

gheinrich:
It would be nice if HDF5 support could be exposed as an optional functionality of the underlying framework. Perhaps something similar to the API to tell whether the framework supports shuffling training data?

tools/create_db.py
lukeyeager: *(added in #247)*

lukeyeager:
You're right - we should definitely do that. Can we merge this (unless you have any other feedback) and do it in a later pull request? I know it's a bit of a step backwards in terms of framework-independence, but I'd like to get this off my plate.

gheinrich:
On e631db34 I had an error when trying to create an HDF5 database (MNIST with compression=GZIP)
```
2015-09-11 15:51:07 [20150911-155103-f98f] [WARNING] Create DB (train) unrecognized output: sid = h5s.create_simple(shape, maxshape)
2015-09-11 15:51:07 [20150911-155103-f98f] [WARNING] Create DB (train) unrecognized output: File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (/tmp/pip-build-JuupVN/h5py/h5py/_objects.c:2467)
2015-09-11 15:51:07 [20150911-155103-f98f] [WARNING] Create DB (train) unrecognized output: File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (/tmp/pip-build-JuupVN/h5py/h5py/_objects.c:2424)
2015-09-11 15:51:07 [20150911-155103-f98f] [WARNING] Create DB (train) unrecognized output: File "h5py/h5s.pyx", line 99, in h5py.h5s.create_simple (/tmp/pip-build-JuupVN/h5py/h5py/h5s.c:1416)
2015-09-11 15:51:07 [20150911-155103-f98f] [WARNING] Create DB (train) unrecognized output: ValueError: Zero sized dimension for non-unlimited dimension (Zero sized dimension for non-unlimited dimension)
```

I am fine with the idea of merging HDF5 support first even if this introduces a slightly digression from the "unified framework" model. Maybe a general comment is that it is not scalable to multiple DB backends (with some logic embedded in the HTML and caffe_train.py)... not to say that I could do better!

tools/create_db.py
```diff
@@ -280,7 +286,17 @@ def _create_hdf5(image_count, write_queue, batch_size, output_file,
...
 
             if len(batch) == batch_size:
-                _write_batch_hdf5(batch, data_dset, label_dset)
+                try:
+                    _write_batch_hdf5(db, batch)
+                except Hdf5DatasetExtendError:
+                    logger.info('Reached HDF5 dataset size limit')
+                    db.close()
+                    db = _create_hdf5_db(output_dir, images_written,
+                            hdf5_dset_limit, compression, image_channels,
```
gheinrich: don't you need to increase hdf5_dset_limit? How is this different from what you're doing on line 261? I am probably missing something...

lukeyeager: I'm closing the first `.h5` file and opening a new one. Caffe can open multiple `.h5` files, but none of them can have a dataset "bigger" than `INT_MAX`.

lukeyeager:
> On e631db3 I had an error when trying to create an HDF5 database (MNIST with compression=GZIP)

That's weird. What version of hdf5 do you have?
```
$ python -c "import h5py; print h5py.__doc__"

    This is the h5py package, a Python interface to the HDF5
    scientific data format.

    Version 2.2.1

    HDF5 1.8.11
```

gheinrich:
I have a slightly newer version of h5py and a slightly older version of HDF5.
```
(venv)gheinrich@android-devel-wks-7:/fast-scratch/gheinrich/ws/digits$ python -c "import h5py; print h5py.__doc__"

    This is the h5py package, a Python interface to the HDF5
    scientific data format.

    Version 2.5.0

    HDF5 1.8.4
```
I am not seeing the problem on 10f3cd21925. Is e631db3 good to test?

lukeyeager:
> ## Release 1.8.7 of May 2011 versus Release 1.8.6
HDF5 now allows the size of any dataspace dimension to be 0 (zero). This was previously allowed only if the maximum size of the dimension was unlimited.
https://www.hdfgroup.org/HDF5/doc/ADGuide/Changes_1_8_x.html

lukeyeager:
Ubuntu 14.04 was released 04/2014 and HDF5 1.8.4 was released 09/2009. It seems weird that you would have such an old version. How did you install HDF5 - do you remember?

gheinrich:
I have Ubuntu 12.04. Apparently the default HDF5 package (installed via `apt-get libhdf5-serial-dev`) is 1.8.4.
```
$ apt-cache show libhdf5-serial-dev
Package: libhdf5-serial-dev
Priority: optional
Section: universe/libdevel
Installed-Size: 16347
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Debian GIS Project <pkg-grass-devel@lists.alioth.debian.org>
Architecture: amd64
Source: hdf5
Version: 1.8.4-patch1-3ubuntu2
```
I did:
```
apt-get update
apt-get install --only-upgrade libhdf5-serial-dev
...
libhdf5-serial-dev is already the newest version.
```
So apparently a newer version of hdf5 package should be installed manually? Sounds like a minor hassle, right?

lukeyeager:
> I have Ubuntu 12.04

Ah, that explains it. [We don't technically support 12.04](https://github.com/NVIDIA/DIGITS/blob/v2.1.0/README.md#installation), BTW.

> So apparently a newer version of hdf5 package should be installed manually? Sounds like a minor hassle, right?

That is a hassle. Let's not make this any harder than it needs to be. Let me put together a workaround ...

gheinrich:
I don't think there was any doubt about it but just in case: it is working on Ubuntu 14.04.
By the way do you think it would be useful to add a little badge in front of each dataset on the main page to show which database backend was used? (similar to the model badge showing the DL framework id)

lukeyeager:
@gheinrich can you verify that the last patch makes this work for 12.04?

gheinrich:
Yes, now this is working on Ubuntu 12.04 with HDF5 1.8.4. Thanks!

lukeyeager:
Note that you need https://github.com/NVIDIA/caffe/pull/26 (Included in `v0.13.2`) to avoid "CNMEM_NOT_INITIALIZED" errors.

aralph:
Now that this PR is merged, is there somebody working on further support of HDF5 for non-image data?

lukeyeager:
@aralph not actively right now. But we are re-evaluating our data format as we evaluate adding new DL frameworks. What's your use-case - why do you want better HDF5 support?

aralph:
@lukeyeager We work with multi-channel data in Caffe. We have faced some limitations with LMDB. Development, manipulation and storing/loading HDF5 turned out to be more flexible in HDF5.
HDF5 also lets you store metadata in the database which does not seem to work in LMDB.

spotofleopard:
Being able to import prebuilt hdf5 dataset would be very helpful. One use case would be vector label, like in SVHN format 1

PR #229, opened 2015-08-20
replace generic model epoch field

jmancewicz:
*Fix bug in #208*


lukeyeager:
LGTM, thanks for the bugfix!

PR #231, opened 2015-08-20
Use input_shape instead of input_dim for deploy

lukeyeager:
See https://github.com/BVLC/caffe/pull/2950 - the examples were leading me astray.

This doesn't break backwards-compatibility with old models. It just starts generating `deploy.prototxt` in the recommended way.


lukeyeager:
Note: DIGITS will start generating models which require NVCaffe `>=0.11.0` (previously `>=0.10.0` was sufficient).

PR #235, opened 2015-08-24
Don't show mean image during dataset creation

gheinrich:
To avoid broken-image icon being displayed as mean image file
not yet created.

### DIFF
--- a/digits/templates/datasets/images/classification/show.html
+++ b/digits/templates/datasets/images/classification/show.html
@@ -89,7 +89,7 @@
         {% endautoescape %}
         );
     </script>
-    {% if task.mean_file %}
+    {% if task.status=='D' and task.mean_file %}
     <b>Image Mean:</b>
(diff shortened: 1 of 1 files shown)


digits/templates/datasets/images/classification/show.html
```diff
@@ -89,7 +89,7 @@
         {% endautoescape %}
         );
     </script>
-    {% if task.mean_file %}
+    {% if (not job.status.is_running()) and task.mean_file %}
```
lukeyeager: I think you want something more like `task.status == 'D'` rather than `job.status.is_running()`, right? We can display the mean before the job is done as long as the task is done.

lukeyeager:
At some point we should add a SocketIO message so that the image mean shows up on the page as soon as it is generated. But this PR is fine as a strict improvement.

gheinrich:
Thanks, your way is better indeed!

lukeyeager:
LGTM

Issue #234, opened 2015-08-23
Input must have 4 axes, corresponding to, (num, channels, height, width)


RadicoLabs:
Not sure if this is an issue or not - but i have been searching for days and cannot seem to figure it out!

Any image i try to classify individually using digits seems to run okay. However, when using the "classify many images" button, the network crashes because of the aforementioned title/bug/i dont even know what the hell it is.

Im entirely new to caffe and DIGITS, and as i said i've spent days googling this problem - and cant seem to figure it out. what is the 5th dimension on the image, and if i do actually have 5D images, how do i convert them to 4D? 

Thanks!

lukeyeager:
You shouldn't see that error in DIGITS. What version of Caffe and DIGITS are you using? Do you see this error with a standard network? If not, can you show me the data layers of your network?

> what is the 5th dimension on the image

What do you mean by a 5th dimension? Caffe expects 4 - as you pasted in the question title.

lukeyeager:
This bug should be fixed in #237. Please verify if you can.

RadicoLabs:
Bugfix implemented and working. Thankyou sir!

PR #237, opened 2015-08-24
Bugfix - wrong shape calculated for infer_many

lukeyeager:
Thanks to @RadicoLabs for reporting this bug (several times):

* https://github.com/NVIDIA/DIGITS/issues/234
* https://github.com/BVLC/caffe/issues/2965
* https://groups.google.com/d/msg/digits-users/fv1lcvvmXNE/QXPl7qCPJQAJ

I introduced a bug in f3cee358fa1105d96b5ddd9cc6fd2aa9db3a5857 that wasn't caught by the test suite. I introduced an additional axis to the shape of the data. The reason this wasn't caught is because the `InnerProduct` layer in the deploy network saw the same number of neurons in the previous layer as it did during training - `10x1x10x10` ~= `10x1x1x10x10`. Convolution layers, however, check for the number of axes and throw an error. 

This commit fixes the bug and adds a test class which uses LeNet (including Convolution layers).

### DIFF
--- a/digits/model/images/classification/test_views.py
+++ b/digits/model/images/classification/test_views.py
@@ -568,3 +568,13 @@ layer {
 }
 """
 
+class TestLeNet(TestCreated):
+    IMAGE_WIDTH = 28
+    IMAGE_HEIGHT = 28
+
+    CAFFE_NETWORK=open(
+            os.path.join(
+                os.path.dirname(digits.__file__),
+                'standard-networks', 'lenet.prototxt')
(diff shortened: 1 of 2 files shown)


RadicoLabs:
Awesome work man - thanks for the hotfix

And for the record - I only reported the top 2, not the third one ;)

lukeyeager:
> And for the record - I only reported the top 2, not the third one ;)

My bad. Edited OP.

Issue #187, opened 2015-07-30
Gdk-CRITICAL **: gdk_cursor_new_for_display: assertion 'GDK_IS_DISPLAY (display)' failed; TypeError: constructor returned NULL


Saneesh:
Hello,
I have successfully configured DIGITS in AWS with the supported system like Ubuntu 14.04, CUDA 7.0, GPU etc.

When I trying create the dataset through the browser, the following error is showing:
```
(parse_folder.py:13268): Gdk-CRITICAL **: gdk_cursor_new_for_display: assertion 'GDK_IS_DISPLAY (display)' failed
Default value for log_file "/home/ubuntu/digits/digits/digits.log" invalid:
You do not have write permissions
Traceback (most recent call last):
File "/home/ubuntu/digits/tools/parse_folder.py", line 21, in <module>
digits.config.load_config()
File "/home/ubuntu/digits/digits/config/load.py", line 147, in load_config
option.apply()
File "/home/ubuntu/digits/digits/config/caffe_option.py", line 232, in apply
import caffe
File "/home/ubuntu/caffe/python/caffe/__init__.py", line 1, in <module>
from .pycaffe import Net, SGDSolver
File "/home/ubuntu/caffe/python/caffe/pycaffe.py", line 14, in <module>
import caffe.io
File "/home/ubuntu/caffe/python/caffe/io.py", line 2, in <module>
import skimage.io
File "/usr/local/lib/python2.7/dist-packages/skimage/io/__init__.py", line 15, in <module>
reset_plugins()
File "/usr/local/lib/python2.7/dist-packages/skimage/io/manage_plugins.py", line 89, in reset_plugins
_load_preferred_plugins()
File "/usr/local/lib/python2.7/dist-packages/skimage/io/manage_plugins.py", line 69, in _load_preferred_plugins
_set_plugin(p_type, preferred_plugins['all'])
File "/usr/local/lib/python2.7/dist-packages/skimage/io/manage_plugins.py", line 81, in _set_plugin
use_plugin(plugin, kind=plugin_type)
File "/usr/local/lib/python2.7/dist-packages/skimage/io/manage_plugins.py", line 251, in use_plugin
_load(name)
File "/usr/local/lib/python2.7/dist-packages/skimage/io/manage_plugins.py", line 295, in _load
fromlist=[modname])
File "/usr/local/lib/python2.7/dist-packages/skimage/io/_plugins/matplotlib_plugin.py", line 4, in <module>
[...]

Issue #78, opened 2015-04-16
libdc1394 error: Failed to initialize libdc1394


joyofdata:
Hi!

(Working on Ubuntu 14.04 AMI on EC2 instance g2.2xlarge)

I installed CUDA 7 and c/p-ed the cuDNN files there and also of course installed caffe as described - everything seems to be fine.

So now after running ./digits-devserver for the first time I got an error that the process wasn't permitted to write to /home/ubuntu/.digits/ - lazy me I "solved" this issue by simply setting the permissions for the folder and its content recursively to 777 (```chmod -R 777 ~/.digits```). 

Now after working my way through the initial questions of the config module I get another error:

```
libdc1394 error: Failed to initialize libdc1394

(process:1596): Gtk-WARNING **: Locale not supported by C library.
	Using the fallback 'C' locale.
```

For completeness sake I am adding the full session - I accepted all the suggested defaults:

```
ubuntu@ip-172-31-15-133:~/digits$ ./digits-devserver 
  ___ ___ ___ ___ _____ ___
 |   \_ _/ __|_ _|_   _/ __|
 | |) | | (_ || |  | | \__ \
 |___/___\___|___| |_| |___/

Welcome to the DIGITS config module.

Where is caffe installed?
	(enter "SYS" if installed system-wide)
	[default is /home/ubuntu/caffe]
(q to quit) >>> 
Accepting default value of "/home/ubuntu/caffe"

Attached devices:
Device #0:
	Name                 GRID K520
	Compute capability   3.0
	Memory               4.0 GB
	Multiprocessors      8


Input the IDs of the devices you would like to use, separated by commas, in order of preference.
	(enter "NONE" if you want to run in CPU-only mode)
	[default is 0]
(q to quit) >>> 
Accepting default value of "0"

Where would you like to store jobs?
	[default is /home/ubuntu/.digits/jobs]
(q to quit) >>> 
Accepting default value of "/home/ubuntu/.digits/jobs"

What is the minimum log level that you want to save to your logfile? [error/warning/info/debug]
	[default is info]
(q to quit) >>> 
Accepting default value of "info"

New config:
            gpu_list - 0
          secret_key - 79bb59415609390009ba204b
           log_level - info
            jobs_dir - /home/ubuntu/.digits/jobs
          caffe_root - /home/ubuntu/caffe

libdc1394 error: Failed to initialize libdc1394

(process:1596): Gtk-WARNING **: Locale not supported by C library.
[...]

PR #239, opened 2015-08-27
Fix backend gtk3 error NVIDIA/DIGITS#187 NVIDIA/DIGITS#78

nszceta:


lukeyeager:
@gheinrich, it looks like you first suggested this fix at https://github.com/NVIDIA/DIGITS/issues/187#issuecomment-126611403. Can you explain where you got this from and why it works? I'd rather not add another library dependency if this can be solved in another way.

gheinrich:
My vague understanding is that matplotlib can operate with a number of rendering backends. There are 'interactive' backends which allow the user to interact with the backend (I suppose for zooming, panning, etc.). There are also non-interactive backends that e.g. only allow the rendering to be saved to files. By default matplotlib uses Gtk3, which is an interactive backend and needs a working display. Matplotlib has an API to switch the rendering backend. Doing `matplotlib.use('Agg')` selects a non-interactive backend, as suggested on BVLC/caffe#861.

We are using skimage which relies on matplotlib and I think the 'fix' is to make sure we switch the matplotlib backend before using skimage. In my understanding we already have a dependency on matplotlib through skimage so I think the change does not add a new dependency.

lukeyeager:
Thanks for the explanation! Should we make that an explicit requirement in `requirements.txt` or is this a caffe-specific thing?

nszceta:
It is a dependency of skimage. Nothing needed.

--
Adam Gradzki

lukeyeager:
Right, but we don't explicitly require either `scikit-image` or `matplotlib` in [requirements.txt](https://github.com/NVIDIA/DIGITS/blob/v2.1.0/requirements.txt) - those are coming from [Caffe's list](https://github.com/NVIDIA/caffe/blob/v0.13.1/python/requirements.txt). All libraries used in DIGITS ought to be listed in the DIGITS requirements, even if they're also listed in the pycaffe requirements.

@nszceta, can you add this line to `requirements.txt` in this pull request please?

    matplotlib>=1.3.1

Also, I'll need you to send us a signed CLA before I can accept your contribution. Sorry for the inconvenience.

nszceta:
I can not sign the CLA at this time. Please close the pull request.

--
Adam Gradzki

PR #240, opened 2015-08-27
Job management

jmancewicz:
Initial Job Management page.

There is no link to it

http://localhost:5000/job_management


lukeyeager:
What's the status on this?

jmancewicz:
The only comment from Andrews was to have mAP rather than Loss being graphed.  But it sounds like the plan is to show loss unless mAP is available.  That sounds like an enhancement.  It could show both if mAP is output.

This could be added to the head of the main page, with the dataset and model jobs just beneath.

lukeyeager:
I think any version of this will be an improvement on the current home page. Can you rebase to fix the merge conflicts and let me know when this is ready for review?

jmancewicz:
Rebased.  I'll move it to the home page next.

jmancewicz:
@lukeyeager, can you have a look.  I might move the job_management.html into home.html rather than include it.

lukeyeager:
Looks pretty good!

1) The "GPUs" available widget doesn't update when a job starts/finishes. That's misleading.

job-mgmt-gpus-available

2) If you refresh the home page, all the loss graphs get reset to blank, even when there is some data to display.

job-mgmt-refresh-loss

3) The job page tells you how much time is remaining, while the home page tells you how long the job has been running.

Job page:
job-mgmt-eta

Home page:
job-mgmt-ago

Let's be consistent. How about doing what Google Maps does - tell you (1) the time remaining and (2) the estimated time that will be.
http://2.bp.blogspot.com/-dJ7ajrgiRL8/U2E509S7EsI/AAAAAAAAC9I/ScmVKo03F98/s1600/Navigation+with+Lane+Guidance.png

digits/job.py
```diff
@@ -202,3 +211,44 @@ def disk_size_fmt(self):
         size = fs.get_tree_size(self._dir)
         return sizeof_fmt(size)
 
+    def get_tasks_recursively(self):
+        tasks = []
+        for task in self.tasks:
+            if task:
+                tasks.append(task)
+                tasks += task.get_tasks_recursively()
+        return tasks
```
lukeyeager: I'm confused by this function. If you have tasks A and B, and A is B's parent, wouldn't this function return `[A,B,A]`? Is that what you want? Don't you just want the list at `job.tasks`?

digits/static/js/jquery.timer.js
```diff
@@ -0,0 +1,111 @@
...
+	 	};
+
+	 	if(this.init) {
+	 		return new $.timer(func, time, autostart);
+	 	} else {
+			this.set(func, time, autostart);
+	 		return this;
+	 	}
+	};
+})(jQuery);
```
lukeyeager: This isn't a huge library, but I'd still like to minimize it for the sake of consistency. Make sure to add the license back to the top of the file after minimizing! (see related discussion at #267)

jmancewicz:
@lukeyeager 
I'll update the available GPUs, and keep the graphs after refresh.

The trick about time estimation is that Jobs page shows the estimate time remaining for the task not the job.  Projecting the time for the Job is not going to be very reliable.  Could just the last task, or display the times for each task.

digits/job.py
jmancewicz: Wait, so if A is B's parent, then it is also in the job's list of tasks? That sounds like the node is in the graph twice.

lukeyeager: Yes, all tasks are in the `job.tasks` list. The `task.parents` field is optional, and may describe dependencies between tasks. The tasks are not necessarily a fully-connected graph. We can change this behavior if you have a good reason for it.

lukeyeager:
> The trick about time estimation is that Jobs page shows the estimate time remaining for the task not the job. Projecting the time for the Job is not going to be very reliable. Could just the last task, or display the times for each task.

Oh right, I forgot about this problem. How ridiculous would it be for us to assume that all tasks take the same amount of time? Could we display the overall, naively-averaged progress on top, and the progress of each task below? That might look ugly, I'm just spitballing here.

I don't think we need to solve this before merging.

digits/job.py
jmancewicz: No need to change it at this point.  I was just working with certain assumptions about the graph.  I removed that method in job.py and task.py.  So, I'm happy with that.

jmancewicz:
@lukeyeager, I emit gpu availability after resources are allocated or deallocated.  That should be pretty solid, I think.

The sparkline drawn when the page is refreshed.

I've removed the get_tasks_recursively method, because job.tasks was the exhaustive task list.

I'd too would rather get this out and deal with the eta issue in the future.

lukeyeager:
Sounds good, thanks.

You've still got some tests failing on the Travis build. Let's get those sorted out and merge.

jmancewicz:
Quick, it's passing travis!

digits/scheduler.py
```diff
@@ -175,6 +177,30 @@ def add_job(self, job):
             return False
         else:
             self.jobs.append(job)
+
+            if flask._app_ctx_stack.top is not None:
```
lukeyeager: What does this check do?

jmancewicz: As I recall, I was getting a failure on tests that I wasn't getting in practice, and the error was in flask that flask._app_ctx_stack.top was None.  It was probably late, and I I couldn't ping you, so I did that.  Let me see if it's still needed.  If it still errors, I'll ping you.

jmancewicz: ======================================================================
ERROR: digits.test_scheduler.TestSchedulerFlow.test_add_remove_job
----------------------------------------------------------------------
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/jmancewicz/dev/digits/DIGITS/digits/test_scheduler.py", line 44, in test_add_remove_job
    assert self.s.add_job(job), 'failed to add job'
  File "/home/jmancewicz/dev/digits/DIGITS/digits/scheduler.py", line 183, in add_job
    html = flask.render_template('job_row.html', job = job)
  File "/usr/local/lib/python2.7/dist-packages/flask/templating.py", line 126, in render_template
    ctx.app.update_template_context(context)
AttributeError: 'NoneType' object has no attribute 'app'

I wasn't sure what was missing and it looks like I just committed the stopgap measure.  What's the correct way to avoid that error? @lukeyeager

lukeyeager: Does this work?
```python
with app.app_context():
    html = flask.render_template('job_row.html', job = job)
```
Examples:
https://github.com/NVIDIA/DIGITS/blob/v2.2.1/digits/job.py#L161-L162
https://github.com/NVIDIA/DIGITS/blob/v2.2.1/digits/task.py#L100-L105

jmancewicz: Not quite. New error. Some new caffe errors, so rebuilt caffe and pycaffe.  Incidentally make -j10 doesn't work in caffe until the first few targets build.  Looking into the new error.

======================================================================
ERROR: Failure: AttributeError ('module' object has no attribute 'scheduler')
----------------------------------------------------------------------
  File "/usr/local/lib/python2.7/dist-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/usr/local/lib/python2.7/dist-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/usr/local/lib/python2.7/dist-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/jmancewicz/dev/digits/DIGITS/digits/test_scheduler.py", line 7, in <module>
    from . import scheduler as _
  File "/home/jmancewicz/dev/digits/DIGITS/digits/scheduler.py", line 23, in <module>
    from digits.webapp import app
  File "/home/jmancewicz/dev/digits/DIGITS/digits/webapp.py", line 20, in <module>
    scheduler = digits.scheduler.Scheduler(config_value('gpu_list'))
AttributeError: 'module' object has no attribute 'scheduler'

lukeyeager: That looks like it could be some sort of circular import issue. Are you messing with any import statements in scheduler or webapp? The import order is very touchy and delicate.

jmancewicz: Ok @lukeyeager,  I'm at an error that has to do with the fact that the test_add_remove_job creates base class Jobs. When they are created or deleted, a message is sent to rebuild the home page and display the recently created or deleted job. That process queries the method Job.job_type(), which raises an exception in the base class.  So, it's the fact that the test creates an instance that is not meant to be instantiated.

The simplest fix seems to be to subclass a TestJob(Job) that implements job_type().

Does that seem reasonable?

lukeyeager: You're talking about this test?
https://github.com/NVIDIA/DIGITS/blob/v2.2.1/digits/test_scheduler.py#L42-L47

That's a pretty sloppy test - my bad.

You could also give Jobs a default type or make `job_type()` return `None` instead of throwing an error. Both seem easier than subclassing Job to me.

jmancewicz: File "/home/jmancewicz/dev/digits/DIGITS/digits/templates/job_row.html", line 5, in top-level template code
    <td><h4 class="list-group-item-heading"><a href="{{ url_for(show_func, job_id=job.id()) }}">{{ job.name() }}  </a></h4></td>
  File "/usr/local/lib/python2.7/dist-packages/flask/helpers.py", line 287, in url_for
    raise RuntimeError('Application was not able to create a URL '

@lukeyeager, I think this is close to the error that caused me to add the line you questioned. url_for doesn't work in this test.  From what I can see it's that the server is not running or the SERVER_NAME is not set.  It feels like we talked about this, but I don't recall if there was a resolution.

jmancewicz: yes, that's the test.  I did the subclass, and ran into the above error.

jmancewicz: @lukeyeager, as far as imports go, the only import change from origin/master in webapp or scheduler is in scheduler.py

```from digits.utils import subclass, override```

Comment (jmancewicz) [inline digits/scheduler.py]: So I seem to be past whatever the import issue may have been, but url_for is not working, which is why I had bailed out if there was not an app context (if that is what that was).

**Comment by lukeyeager** [comment]
The training loss spark line isn't showing up for me. Is it for you?

job-mgmg-no-loss

**Comment by jmancewicz** [comment]
Removing the 'ago' text which shows how long ago the job started.  There are potential time computation issues between client and server that need to be resolved.  This will most likely will return in the future.

File: digits/views.py [hunk 4]
```diff
@@ -166,6 +166,21 @@ def job_status(job_id):
...
+
+    """
+
+    running_datasets    = get_job_list(dataset.DatasetJob, True)
+    running_models      = get_job_list(model.ModelJob, True)
+
+    return flask.render_template('job_management.html',
+                                 running_job = running_datasets + running_models,
+                             )
+
```
lukeyeager: @jmancewicz did you mean to leave this route in? It throws an error on my machine when I try to access this url:
```
UndefinedError
'running_datasets' is undefined

    rv = self.dispatch_request()
  File "/usr/share/digits/digits/views.py", line 181, in job_management
    running_job = running_datasets + running_models,
  File "/usr/lib/python2.7/dist-packages/flask/templating.py", line 128, in render_template
    context, ctx.app)
  File "/usr/lib/python2.7/dist-packages/flask/templating.py", line 110, in _render
    rv = template.render(context)
  File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 969, in render
    return self.environment.handle_exception(exc_info, True)
  File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 742, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/share/digits/digits/templates/job_management.html", line 204, in top-level template code
    {% block content %}
  File "/usr/share/digits/digits/templates/job_management.html", line 206, in block "content"
    {% set running_jobs = running_datasets + running_models %}
UndefinedError: 'running_datasets' is undefined
```

jmancewicz:
Ah. Nope. That's left over from the first version.

PR #241, opened 2015-08-28
Force task progress to 1.0 when done

jmancewicz:
Force task progress to 1.0 (100%) when the status is set to Done.

### DIFF
--- a/digits/status.py
+++ b/digits/status.py
@@ -85,6 +85,7 @@ class StatusCls(object):
     """
 
     def __init__(self):
+        self.progress = 0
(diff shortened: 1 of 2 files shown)


lukeyeager:
It seems more logical to put this fix where we set the status to `DONE`:

https://github.com/NVIDIA/DIGITS/blob/v2.1.0/digits/task.py#L248

Would that work too?

lukeyeager:
LGTM, thanks!

jmancewicz:
It made sense to have the Status class take care of this.

PR #243, opened 2015-08-28
Add login facility to DIGITS

semisight:
DO NOT merge yet!

This is a work in progress for "light login" capabilities. This does NOT attempt to recreate full authentication or access control. It lays the groundwork for doing things like tagging different jobs by the user that submitted them, or for warning the user if they're about to delete another person's job.

The design: add a decorator to all view functions except for the login page. The decorator checks for a signed cookie with the username in it. If there is no cookie, it redirects to the login page where a user can identify themselves.

NOTE: this pull request breaks nearly every test in the DIGITS test suite. Fake authentication will need to be added to each test.

### DIFF
--- a/digits/dataset/images/classification/views.py
+++ b/digits/dataset/images/classification/views.py
@@ -4,7 +4,7 @@ import os
 
 import flask
 
-from digits import utils
+from digits.utils.session import session_required
 from digits.utils.routing import request_wants_json, job_from_request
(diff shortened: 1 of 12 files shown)


lukeyeager:
Thanks!

Closing for now as you won't be working on it anymore, but we'll use this a reference for when we do get around to implementing this feature.

Issue #188, opened 2015-07-31
Add javascript to zoom in on weights and activations


lukeyeager:
See discussion here: https://groups.google.com/d/msg/digits-users/iGNobWrf1Ls/qRqTWNZ3EQAJ

It would be nice to be able to zoom in on weights and activations displayed after inference, especially when the generated image is much bigger than the width of the page.

I looked into this previously, but abandoned it for other work. I was working with `magnificent.js` and `jquery.elevatezoom.js`. I don't remember which was better, but I do remember having a hard time getting the javascript to render correctly while operating within a bootbox modal div.

lukeyeager:
Closed in #267.

PR #244, opened 2015-08-28
Open new tab when testing image[s]

lukeyeager:
Open a new browser tab when you click on `Classify One`, `Classify Many`, `Test One` or `Test Many`.

### DIFF
--- a/digits/templates/models/images/classification/classify_many.html
+++ b/digits/templates/models/images/classification/classify_many.html
@@ -1,4 +1,22 @@
 {# Copyright (c) 2015, NVIDIA CORPORATION.  All rights reserved. #}
+{% extends "layout.html" %}
+
+{% block nav %}
+<li><a href="{{url_for('models_show', job_id=job.id())}}">{{job.name()}}</a></li>
+<li class="active"><a>Classify Many</a></li>
+{% endblock %}
+
+{% block content %}
+
+<div class="page-header">
+    <h1>
+        {{job.name()}}
+        <small>
+            {{job.job_type()}}
+        </small>
+    </h1>
+</div>
+
 
 <table class="table">
     <tr>
@@ -23,3 +41,6 @@
     </tr>
     {% endfor %}
 </table>
+
+{% endblock %}
+
--- a/digits/templates/models/images/classification/show.html
+++ b/digits/templates/models/images/classification/show.html
@@ -76,6 +76,7 @@
 
             {% set task = job.train_task() %}
             <hr>
+            <h2>Trained Models</h2>
             <form id="test-model-form"
                 enctype="multipart/form-data"
                 method="post"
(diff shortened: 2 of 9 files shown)


lukeyeager:
Waiting for #249 (more important) to be merged, and then I'll rebase this on top of that.

lukeyeager:
Rebased. @gheinrich, you can merge this now if you like to make way for #188, unless you have any suggestions/critiques.

digits/templates/models/images/generic/show.html
```diff
@@ -146,20 +154,30 @@ <h4 class='text-center'>Dataset</h4>
...
                             <small>Accepts a list of filenames or urls (you can use your val.txt file)</small>
                         </div>
-                        <button name="infer-many-btn" onClick="return inferMany();" class="btn btn-primary">Infer Many Images</button>
+                        <button name="infer-many-btn"
+                            formaction="{{url_for('generic_image_model_infer_many', job_id=job.id())}}"
+                            formmethod="post"
+                            formenctype="multipart/form-data"
+                            formtarget="_blank"
+                            class="btn btn-primary">
+                            Test One
```
gheinrich: I think this should be "Test Many"

digits/templates/models/images/classification/classify_many.html
```diff
@@ -1,4 +1,17 @@
 {# Copyright (c) 2015, NVIDIA CORPORATION.  All rights reserved. #}
+{% extends "layout.html" %}
+
+{% block nav %}
+<li><a href="{{url_for('models_show', job_id=job.id())}}">{{job.job_type()}}</a></li>
```
gheinrich: is it worth showing the job name too?

digits/templates/models/images/classification/show.html
```diff
@@ -76,6 +76,7 @@ <h4 class='text-center'>Dataset</h4>
 
             {% set task = job.train_task() %}
             <hr>
+            <h2>Test Model</h2>
```
gheinrich: I am unsure about this but perhaps the 'Test Model' title should be moved down to after the 'Download Model' button?

digits/templates/models/images/classification/show.html
```diff
@@ -117,18 +122,21 @@ <h4 class='text-center'>Dataset</h4>
...
                             formaction="{{url_for('models_download', job_id=job.id())}}"
                             formmethod="post"
                             formenctype="multipart/form-data"
                             class="btn btn-info">
-                            Download
+                            Download Model
                         </button>
                     </div>
                 </div>
                 {% if task.get_framework_id() in framework_ids %}
```
gheinrich: the 'Test Model' title could possibly be moved after this line. I have added this line to verify that the framework used for training is available in the current config - so that for example we don't allow the user to try inference on a Torch model if Torch is not available. This may happen if you move from the Torch branch to the master branch.

gheinrich: Alternatively, perhaps we need to move this line up (before the model selection button)?

digits/templates/models/images/classification/show.html
lukeyeager: I see the model selection as mainly used for sake of testing the model, and the `Download Model` button as an extra feature on the side (literally, in this case :D).

digits/templates/models/images/classification/show.html
lukeyeager: I just renamed it to `Trained Models` instead of `Test Model`. Does that work?

lukeyeager:
Thanks for the feedback - commit updated.

PR #248, opened 2015-08-29
Update README.md

ZhijieWang:
Confirm the error in the following discussion thread.
https://groups.google.com/forum/#!topic/digits-users/58zEGIbwQw0

Environment:

Ubuntu 14.04 fresh install. Everything up to date with apt-get update and upgrade.
Cuda 7.0 Fresh install
Gigabyte Geforce 770 4GB OC
Caffe version 0.13 Nvidia port, did not test with 0.12 branch
Followed instruction step by step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ You can run DIGITS in two ways:
 
 Starts a development server that listens on port 5000 (but you can change the port if you like - try running it with the --help flag).
 
+If you run into errors starting "Did you forget to "make pycaffe"? ... ImportError: libcudart.so.7.0: cannot open shared object file: No such file or directory " Then try executing following command `sudo ldconfig /usr/local/cuda/lib64`
+
 Then, you can view your server at `http://localhost:5000/`.
 
 ### Production mode


lukeyeager:
Thanks for contributing!

Did this comment not do the trick?

> * Don't forget to set your path. For example:
  * `CUDA_HOME=/usr/local/cuda`
  * `LD_LIBRARY_PATH=$CUDA_HOME/lib64:$LD_LIBRARY_PATH`

>https://github.com/NVIDIA/DIGITS/blob/v2.1.0/README.md#cuda

lukeyeager:
@ZhijieWang, I just got confirmation from one of our engineers that `ldconfig` is not a good solution. I'll paste his full response:

>`sudo ldconfig /foo/bar/path` is only temporary until `ldconfig` is run again.

>The following will happen to users using this method:
1. User installs CUDA
2. User runs `sudo ldconfig /usr/local/cuda/lib64`
3. User is happy because ldconfig sees cuda libraries
4. Months down the road user installs a library package. Library package runs `ldconfig` as part of it's required post installation actions. This removes /usr/local/cuda/lib64 from ldconfig's database.
5. User is grumpy because ldconfig no longer sees cuda libraries

This will be fixed for CUDA 8.0.

> CUDA 8.0 rpm/deb packages will start installing ld.so.conf.d files and will `ldconfig` for you so you don’t need to edit LD_LIBRARY_PATH in CUDA 8.0 for rpm/deb.

PR #249, opened 2015-08-31
Unified framework interface

gheinrich:
Move (most of) the framework-specific code into unified interfaces: Framework and Train classes.

Squashed commit of the following:

commit 6e7922867f15f701f87a3cae77ef2421de464d96
Author: Greg Heinrich <gheinrich@nvidia.com>
Date:   Sun Aug 30 17:24:53 2015 +0200

    Framework-agnostic model files

commit fead6f44c87fdfcaba6906af14aa27203428b2cc
Author: Greg Heinrich <gheinrich@nvidia.com>
Date:   Sun Aug 30 10:59:02 2015 +0200

    Fix missing shuffle parameter

commit 1c0edd08cd26bd670e1c8d66a20cd71dce38f464
Author: Greg Heinrich <gheinrich@nvidia.com>
Date:   Sun Aug 30 10:45:19 2015 +0200

    Move framework-specific constants out of utils/constants.py

commit d710a2436e7464ed5b2048392daa3d79fc863acc
Author: Greg Heinrich <gheinrich@nvidia.com>
Date:   Sun Aug 30 10:15:27 2015 +0200

    Remove duplicate h5py requirement

commit 4e0c86b5656a85cd065a3de88a142d7001af1d9c
Author: Greg Heinrich <gheinrich@nvidia.com>
Date:   Sat Aug 29 22:42:38 2015 +0200

    Show framework ID on result page

commit d5b68300eaa228ee68866f0f1eb5c2fdc75bab15
Merge: d102d34 8a3d791
Author: Greg Heinrich <gheinrich@nvidia.com>
Date:   Sat Aug 29 22:16:37 2015 +0200

    Merge remote-tracking branch 'origin/master' into HEAD

    Conflicts:
    	docs/API.md
    	docs/FlaskRoutes.md

commit d102d341ac76f2d20be4d9d627b7208b90368103
Author: Greg Heinrich <gheinrich@nvidia.com>
Date:   Fri Aug 28 16:34:50 2015 +0200

    Change comment style

commit 0ba079df9d3cf680959c1912c71c16f68014edfd
Author: Greg Heinrich <gheinrich@nvidia.com>
Date:   Fri Aug 28 16:15:36 2015 +0200

    Report training loss on first iteration - closes #121

commit 11e0c24c4f52006c09c59aee340bd08fedf46263
Author: Greg Heinrich <gheinrich@nvidia.com>
Date:   Fri Aug 28 12:12:22 2015 +0200

    Make Torch optional

[...]

PR #253, opened 2015-08-31
Results browser tweaks

lukeyeager:
*See #195*

Added some tests, cleaned up some HTML and moved the template file location.

### DIFF
--- a/digits/model/images/classification/test_views.py
+++ b/digits/model/images/classification/test_views.py
@@ -326,6 +326,11 @@ class BaseTestCreated(BaseViewsTestWithModel):
                 break
         assert found, 'model not found in list'
 
+    def test_models_page(self):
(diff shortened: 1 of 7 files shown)


lukeyeager:
@jmancewicz, can you take a look at this?

lukeyeager:
Rebased to `master` after #249 changed the autodoc files. See #257 for how to avoid this type of thing in the future.

lukeyeager:
@jmancewicz, updated to include your suggestions from https://github.com/lukeyeager/DIGITS/commit/c7aa5d1418033d484272595d747ed7f04ed64fb3#commitcomment-13007232. Does this look better?

Next time, add comments to the PR diff instead of to the comment itself so that the comment stays attached to this PR. See https://github.com/NVIDIA/DIGITS/pull/249#discussion-diff-38360470 as an example.

PR #257, opened 2015-09-01
Remove line number from autodoc files

lukeyeager:
With this change, you'd still have to run `scripts/generate_docs.py` when (1) adding or removing a route or (2) changing a filename, but not when simply adding a line to a function in a view. It was a cool feature, but not worth the headache.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -20,7 +20,7 @@ Methods: **GET**
 
 Arguments: `job_id`
 
-Location: [`digits/dataset/views.py@15`](../digits/dataset/views.py#L15)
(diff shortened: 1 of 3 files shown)


lukeyeager:
@gheinrich, do you agree that this would be helpful? I don't think many people are using the links anyway.

FYI, I'm talking with @acoomans about posting a new version of Flask-Autodoc to PyPI so we don't require [a custom version](https://github.com/NVIDIA/DIGITS/blob/v2.1.0/requirements_test.txt#L4).

gheinrich:
Oh yes, that would have saved a lot of failed tests!

PR #259, opened 2015-09-01
Add status test

jmancewicz:


digits/test_status.py
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) 2014-2015, NVIDIA CORPORATION.  All rights reserved.
```
lukeyeager: Should be `2015`

lukeyeager:
This is excellent, thanks!

digits/test_status.py
jmancewicz: Changed. Thanks.

lukeyeager:
It looks like `utils/time_filters.py` is causing the same kind of test coverage oscillation:

https://coveralls.io/builds/3463835

jmancewicz:
I'll have a look.

Issue #255, opened 2015-09-01
"Infer Many Images" function does not work properly for Generic Inference


andredubbel:
After playing around a bit with the new generic inference functionality I found what seems to be a bug in with the "Infer Many Images"-function; all outputs are the same:

![screenshot from 2015-09-01 15 51 55](https://cloud.githubusercontent.com/assets/14016877/9606119/cf03b5a6-50c1-11e5-8110-665b964cdda7.png)

The displayed filenames are correct, and when testing them one by using the "Infer One Image" button they output reasonable results. I'm running the latest version of master 8a3d791fdd3488d30ffa66c1eb75820908ec6d2e

gheinrich:
Hi @andredubbel can you try this patch: https://github.com/gheinrich/DIGITS/commit/d26703a206a1bfb9c2191ad0a0fb6ee3d5aea05a
Thanks!

andredubbel:
It got better but there's still something weird going on: 

![screenshot from 2015-09-01 17 06 51](https://cloud.githubusercontent.com/assets/14016877/9608068/cdaf110e-50cc-11e5-923c-f57201b37e6e.png)


For some reason the first and third row are the exactly the same, and the same goes for the second and forth. After that each row seems to be unique.

If I apply the 4 first images one by one the second and fourth gets the values shown in the image while the first and third get other values.

gheinrich:
I don't see this on my test sample. Any chance you can double check? Thanks!

descampsa:
I had a similar issue, the results of the first chunk of images is invalid because they are overwriten by the second chunk results.
Try this patch: https://github.com/descampsa/DIGITS/commit/f4a636d4bf8241abe26754cac3c993954ebc5fd1

lukeyeager:
@gheinrich, your patch seems to solve the problem for me. Can you open a pull request with the fix?

@andredubbel, thanks for the bug report!

lukeyeager:
Closed in #261.

PR #260, opened 2015-09-02
Fix error in score reporting of "Infer Many Images" function

descampsa:
Fix https://github.com/NVIDIA/DIGITS/issues/255

### DIFF
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -6,6 +6,7 @@ import time
 import math
 import subprocess
 import operator
+import copy
 
 import numpy as np
 from google.protobuf import text_format
@@ -1208,7 +1209,7 @@ class CaffeTrainTask(TrainTask):
                         'data', image)
             output = net.forward()[net.outputs[-1]]
             if scores is None:
-                scores = output
+                scores = copy.deepcopy(output)
             else:
                 scores = np.vstack((scores, output))
             print 'Processed %s/%s images' % (len(scores), len(caffe_images))


lukeyeager:
@descampsa, #255 reports a problem with `infer_many()` (for `GenericModelJobs`), but this is a fix for `classify_many()` (for `ClassificationModelJobs`). There's no problem with `classify_many()` as far as I know.

Please explain.

descampsa:
Humm, sorry about the poorly described patch and the confusion.
I noticed a bug in `classify_many`, and the symptoms described in https://github.com/NVIDIA/DIGITS/issues/255 were so close to mine that i thought it was the same issue.

Here is the bug i noticed : the scores for the first chunk of images (images 0 to 15 in my case) returned by classify_many are incorrect, and they are the same as the next chunk (image 16 to 31). 
This is due to the way the score vector is accumulated : the first iteration does a swallow copy of the result, and second iteration write on the same vector and overwrite first results. 
Next iteration are fine because they use vstack.
The patch fix the issue for me by using `copy.deepcopy` at first iteration.

As for https://github.com/NVIDIA/DIGITS/issues/255, i am not sure anymore if it is related, but it seems that a similar logic is used in `infer_many`, so maybe a similar patch (at line [1258](https://github.com/descampsa/DIGITS/blob/f4a636d4bf8241abe26754cac3c993954ebc5fd1/digits/model/tasks/caffe_train.py#L1258) ) is needed.

lukeyeager:
Thanks for the explanation. I can reproduce the bug on my machine exactly as you described - thanks for the bugfix! As you guessed, this is broken for `infer_many` as well, can you add the same fix at line 1258?

Also, I'll need you to send us a signed CLA before I can accept your contribution.

descampsa:
If i have to sign an agreement, i prefer that you consider this as an issue report, and fix it yourself, especially for such a straightforward fix.
Thanks.

PR #266, opened 2015-09-03
test time_filters.py

jmancewicz:
Try to cover everything in utils/time_filters.py.


lukeyeager:
Wow [100% coverage](https://coveralls.io/builds/3471896/source?filename=digits%2Futils%2Ftime_filters.py), sweet!

PR #272, opened 2015-09-04
Keep contents of auto-complete fields when hitting back button

### LINKED ISSUE #269: Autocomplete fields empty after navigating back to previous page
How to reproduce (on Chrome, at least):

1. Input values to create a generic inference dataset (lots of autocomplete fields) and submit the form
2. Go back in your browser
3. The autocomplete fields are empty (but the "Dataset name" field is not)

gheinrich:
closes #269

See http://bugs.jqueryui.com/ticket/7790

### DIFF
--- a/digits/templates/layout.html
+++ b/digits/templates/layout.html
@@ -61,6 +61,7 @@ window.onload = function () {
             return baseName(suggestion.value);
         }
     });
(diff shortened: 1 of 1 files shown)


lukeyeager:
That does the trick, thanks!

PR #283, opened 2015-09-09
Notes functionality added for every Job

DESHRAJ:
Now, notes can be added for every job. Also, the user can update the notes. Let me know if any changes are required.

### DIFF
--- a/digits/job.py
+++ b/digits/job.py
@@ -55,6 +55,7 @@ class Job(StatusCls):
         self.pickver_job = PICKLE_VERSION
         self.tasks = []
         self.exception = None
+        self._notes = None
 
         os.mkdir(self._dir)
 
@@ -131,6 +132,12 @@ class Job(StatusCls):
 
     def name(self):
         return self._name
+
+    def notes(self):
+        if hasattr(self, '_notes'):
+            return self._notes
+        else:
+            return None
 
     def job_type(self):
         """
--- a/digits/templates/job.html
+++ b/digits/templates/job.html
@@ -219,6 +219,42 @@ $('#delete-job').on('click', function(event) {
         </div>
         </div>
         <div class="gpu-utilization-info" style="display:none;"></div>
+
+        <div id="notes-information" class="panel panel-default">
+            <div class="panel-heading">
+                <h4>Notes</h4>
+            </div>
+            <div class="panel-body">
+                <div id="show-job-notes" class="col-sm-12">
+                    <p style="display:inline;">{{ job.notes() }}</p>
+                    <a href=# onClick="$('#show-job-notes').hide(); $('#edit-job-notes').show(); return false;"><span class="glyphicon glyphicon-edit"></span></a>
+                </div>
+
+                <div id="edit-job-notes" class="col-sm-12" style="display:none;">
+                    <script type="text/javascript">
+                        function updateJobNotes() {
+                        var newNotes = $("textarea[name=job-notes]").val();
+                        $.ajax("{{url_for('edit_job_notes', job_id=job.id())}}",
+                                {
+                                    type: "PUT",
+                                    data: {"job_notes": newNotes}
+                                    }
+                              )
+                        .done(function() {
+                                $("#show-job-notes p").html(newNotes);
+                                $('#edit-job-notes').hide(); $('#show-job-notes').show();
+                                })
+                        .fail(function(data) { errorAlert(data); });
+                        return false;
+                    }
+                    </script>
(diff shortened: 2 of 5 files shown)


DESHRAJ:
The build has failed because /home/travis/build/NVIDIA/DIGITS/docs/FlaskRoutes.md needs to be regenerated. @lukeyeager , please let me know if I need to change something to make it run successfully or it needs to be configured from the server side.

digits/job.py
```diff
@@ -132,6 +133,9 @@ def path_is_local(self, path):
     def name(self):
         return self._name
 
+    def notes(self):
+        return self._notes
```
lukeyeager: Let's add a check to make sure that `_notes` exists and return `None` if not. For older jobs I'm getting this error:

    AttributeError: 'ImageClassificationModelJob' object has no attribute '_notes'

DESHRAJ: Ok, I have added the condition now. Please check the condition.

lukeyeager:
@DESHRAJ this is great work, thanks!

**TravisCI**
You need to run `scripts/generate_docs.py`. Make sure you run `pip install -r requirements_test.txt` first to get the latest version of `Flask-Autodoc`.

**Tests**
Please add a test for this. Props if you add one for editing job names too (I forgot to do that).

**CLA**
I'll need you to send us a signed CLA before I can accept your contribution.

lukeyeager:
@jmancewicz, how do you feel about the placement for the notes? I don't have a problem with it - it's no worse than the rest of the layout.

digits/job.py
```diff
@@ -132,6 +133,12 @@ def path_is_local(self, path):
     def name(self):
         return self._name
 
+    def notes(self):
+        if self._notes:
```
lukeyeager: This actually doesn't help. You want this:
```python
if hasattr(self, '_notes'):
```

DESHRAJ:
Ok sure, I will add the test for editing the name also. :)

DESHRAJ:
@lukeyeager , I have successfully sent the CLA. Also, I have updated the docs. Now, its all working.

lukeyeager:
I see that, thanks!

1. You said you were going to add some tests - do you need some help?

2. Before merging, I'll need you to squash this all into a single commit.
```
git remote update
git rebase -i origin/master
# squash all but the first commit
```

DESHRAJ:
@lukeyeager , Is it right to rebase the commits that are already pushed ? Its not recommended to rebase the commits that are pushed. Please let me know if I am wrong. :)

lukeyeager:
**Squashing**
It's a question of style, really. I don't think that a change this small requires 8 commits. Squashing commits in a feature branch before merging them is perfectly fine.
You're probably thinking of the rule that public branches should never have their commit history rewritten. But a pull request branch doesn't count as a "public branch."

**Tests**
You didn't actually add any tests yet - only some helper stubs for tests. You can check to see if the total number of tests increased.
If you can't figure it out, don't worry about it. I'll write some tests for it after we merge this.

DESHRAJ:
Yeah, I was thinking about the public branches rule but now it seems fine. Ok I will do this as soon as possible. 

For the tests part: I think I need some help for that. Can you provide me some link for demo tests and the file where I need to write the tests ?

lukeyeager:
You probably want to add a `test_edit_notes()` function to these four classes:

https://github.com/NVIDIA/DIGITS/blob/3791003670/digits/dataset/images/classification/test_views.py#L382
https://github.com/NVIDIA/DIGITS/blob/3791003670/digits/dataset/images/generic/test_views.py#L194
https://github.com/NVIDIA/DIGITS/blob/3791003670/digits/model/images/classification/test_views.py#L325
https://github.com/NVIDIA/DIGITS/blob/3791003670/digits/model/images/generic/test_views.py#L328


**Reference**
Here's a PR with some really basic tests: https://github.com/NVIDIA/DIGITS/pull/253
And here's a PR with a slightly more complicated test: https://github.com/NVIDIA/DIGITS/pull/284

DESHRAJ:
Hey @lukeyeager , can you merge the pull request and write the test cases yourself. I am finding it somewhat difficult and time taking.

lukeyeager:
> Hey @lukeyeager , can you merge the pull request and write the test cases yourself. I am finding it somewhat difficult and time taking.

Sure, no problem. Thanks for squashing, and for the great work!

jmancewicz:
I agree, the location is a bit out of the way, but not worse than what else is going on.  It may be moved around in the not too distant future.

PR #285, opened 2015-09-09
Fix previous network snapshot lists

lukeyeager:
Fix bug introduced in 66d8404713b0fa94f84a33a2cf4031fa20e6aa3c for `ClassificationJobs` - the snapshot list for the first job on the second page was really the list for the first job on the first page.

Add pagination for `GenericJobs`. It wasn't added with 66d8404713b0fa94f84a33a2cf4031fa20e6aa3c.


lukeyeager:
@jmancewicz, please verify that this fixes the issue.

jmancewicz:
Looks good.

PR #286, opened 2015-09-10
Add more server info to /index.json

lukeyeager:
`jobs_dir` and `server_name` (if set)


lukeyeager:
Added `version`, too. @gheinrich, any concerns with this?

gheinrich:
That looks good to me (how do you set the server name by the way? it's unset on my config)

lukeyeager:
> how do you set the server name by the way? it's unset on my config

You can edit it with `python -m digits.config.edit --verbose`

PR #296, opened 2015-09-11
Sort subfolders in parse_folder.py

### LINKED ISSUE #289: Incorrect data labeling
Hi, all!
I've created a dataset of English alphabet characters. It was automatically split into train, valid (15%) and test (10%) sets. I trained a CNN on it and got 97% accuracy on validation set.
I want to test the net on test images. For that purposes I want to use test.txt created by DIGITS.

But labeling in that file is a bit strange. Here's a portion of a file:

Files from folder '63' should be labeled '63' but not '0'.
When I try to classify some of those images via python, they are classified as '0' (which is correct according to the txt). When I classify them in DIGITS GUI they are correctly classified as '63'.
What am I doing wrong?

lukeyeager:
*Close #289*

### DIFF
--- a/tools/parse_folder.py
+++ b/tools/parse_folder.py
@@ -331,6 +331,8 @@ def parse_folder(folder, labels_file,
             logger.error('folder does not exist')
             return False
 
+    subdirs.sort()
+
     if len(subdirs) < 2:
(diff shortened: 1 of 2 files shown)


gheinrich:
This looks good to me!

PR #302, opened 2015-09-13
Fix CNMEM flags in Caffe build instructions

gheinrich:


lukeyeager:
The libname is definitely wrong - thanks for the fix!

If the `USE_CNMEM` flag is required, wouldn't the `USE_CUDNN` flag be required too? Now that I'm looking more carefully, I use both:
```
cmake \
    -DUSE_CUDNN=True -DCUDNN_INCLUDE=$CUDNN_HOME/include -DCUDNN_LIBRARY=$CUDNN_HOME/lib64/libcudnn.so \
    -DUSE_CNMEM=True -DCNMEM_INCLUDE=$CNMEM_HOME/include -DCNMEM_LIBRARY=$CNMEM_HOME/build/libcnmem.so \
    ..
```
See [relevant CMake lines](https://github.com/NVIDIA/caffe/blob/v0.13.2/cmake/Cuda.cmake#L235-L252).

gheinrich:
When I clone NVIDIA/caffe-0.13 I notice CUDNN is ON by default however CNMEM is OFF. In my vague understanding of CMake rules this comes from:
https://github.com/NVIDIA/caffe/blob/v0.13.2/CMakeLists.txt#L17-L18

lukeyeager:
Oh yeah, good call. LGTM.

PR #309, opened 2015-09-15
Show dataset size

gheinrich:


lukeyeager:
Cool stuff, thanks! Any reason we can't add this for ModelJobs, too? Those snapshots can really rack up in disk space.

And it would probably be good to at least write a basic sanity test that checks the size of an empty directory or something.

gheinrich:
Sure, I'll do that!

digits/utils/filesystem.py
```diff
@@ -0,0 +1,30 @@
...
+    return human-readable size
+    """
+    for unit in ['','K','M','G','T','P','E','Z']:
+        if abs(num) < 1024.0:
+            if unit != '':
+                return "%3.1f%s%s" % (num, unit, suffix)
+            else:
+                return "%d%s" % (num, suffix)
+        num /= 1024.0
+    return "%.1f%s%s" % (num, 'Y', suffix)
```
lukeyeager: Is this just a refactoring of `digits.utils.sizeof_fmt` ([here](https://github.com/NVIDIA/DIGITS/blob/v2.1.0/digits/utils/__init__.py#L100-L106))?

gheinrich:
Sorry I hadn't noticed there was already a function for this. I'll remove the one I added.

lukeyeager:
Awesome thanks! Go ahead and merge if this is ready - it looks good to me.

Issue #300, opened 2015-09-13
Digits Freezing on Model Creation


RadicoLabs:
So the image set creation runs fine and completes every task in the job, except for "Create DB (train)" 

This gets caught up with a couple of seconds to go... Any ideas?

When i cancel the test server from the command line it dumps this into the "CreateDB" Panel:

    Exception in thread Thread-7:
    Traceback (most recent call last):
    File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
    File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
    File "/home/dev/DIGITS/tools/create_db.py", line 416, in _load_thread
    image = utils.image.load_image(path)
    File "/home/dev/DIGITS/digits/utils/image.py", line 56, in load_image
    image.load()
    File "/usr/lib/python2.7/dist-packages/PIL/ImageFile.py", line 206, in load
    raise IndexError(ie)
    IndexError: string index out of range
    Traceback (most recent call last):
    File "/home/dev/DIGITS/tools/create_db.py", line 637, in <module>
    lmdb_map_size   = args['lmdb_map_size']
    File "/home/dev/DIGITS/tools/create_db.py", line 143, in create_db
    mean_files, **kwargs)
    File "/home/dev/DIGITS/tools/create_db.py", line 213, in _create_lmdb
    time.sleep(0.2)
    KeyboardInterrupt

gheinrich:
Do you have commit 2a8fc13? This might make the error more legible. Also, are you creating the image dataset out of a folder or a list of image files?

RadicoLabs:
Ha yeah i have that commit - That was my error a few days ago... Now im stuck here.. ;)

And im trying to create from folder

RadicoLabs:
Not sure if Harmless warning or contributing to the error - but the console is throwing LoadImageError: IOError: image file is truncated a lot whilst iterating through training

RadicoLabs:
Update: I ran DIGITS on an old load of images i had and it created an imageSet perfectly fine - Maybe there is something wrong with my data - Who knows... will keep posting updates

RadicoLabs:
Update - Now it's getting stuck on Create DB (val) - My Data should be more or less perfect -Eradicated all images that werent PNG or JPG and changed the file permissions to mirror that of my old Image set.... 

[...]

PR #312, opened 2015-09-16
Update image.py

RadicoLabs:
Load Truncated images instead of ignoring them.

Ignoring too many seems to cause a crash with DIGITS

### DIFF
--- a/digits/utils/image.py
+++ b/digits/utils/image.py
@@ -9,6 +9,9 @@ import numpy as np
 import scipy.misc
 import math
 
+from PIL import ImageFile
+ImageFile.LOAD_TRUNCATED_IMAGES = True
+
 from . import is_url, HTTP_TIMEOUT, errors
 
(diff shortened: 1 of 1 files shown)


lukeyeager:
@RadicoLabs, thanks for addressing #300!

Do you feel confident enough in python to write a test for this in `digits/utils/test_image.py`? If not, that's ok.

Also, I'll need you to send us a signed CLA before I can accept your contribution.

RadicoLabs:
Thanks Luke,

Will send over that agreement ASAP - Also, Having never written a test case for Nvidia before, im not sure what needs to be done.... If you could write this one? then i can take a crack at future ones? 

Thanks mate

lukeyeager:
> If you could write this one? then i can take a crack at future ones?

Sure, no problem. We can merge this first and I'll figure out how to write a test in another PR later.

PR #315, opened 2015-09-16
Fix create_db_task.backend == None scenario

lukeyeager:
If you switch around between different versions of the code, the backend
field can get set to None and that breaks things. This is a more
explicit check.


gheinrich:
Ah thanks I came across this issue.
This looks good to me (also I tested on my datasets and it is working fine).

PR #323, opened 2015-09-21
Show DB backend on home and model creation pages

gheinrich:
note: doesn't show a badge for generic datasets

### DIFF
--- a/digits/dataset/images/classification/job.py
+++ b/digits/dataset/images/classification/job.py
@@ -3,6 +3,7 @@
 from digits.utils import subclass, override
 from digits.status import Status
 from ..job import ImageDatasetJob
+from digits.dataset import tasks
 
 # NOTE: Increment this everytime the pickled object changes
 PICKLE_VERSION = 2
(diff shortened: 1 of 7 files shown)


lukeyeager:
Looks fine to me. Things are starting to get a little cluttered on the home page, but @jmancewicz should be revamping the whole thing with #240 soon anyway.

> note: doesn't show a badge for generic datasets

Let's just put an LMDB badge on all of them since that's the only backend supported at the moment anyway. It would probably be a good idea to actually add a `backend` field to generic datasets too, since we'll be supporting more options in the future.

gheinrich:
Actually, in my understanding image dataset jobs (neither generic nor classification ones) don't have a backend field. For image classification datasets, the information is coming from the training DB task (there is an assumption that an image classification dataset uses a consistent DB backend). The thing is generic datasets don't have a training DB task. Where do you suggest I add the backend field?

lukeyeager:
> For image classification datasets, the information is coming from the training DB task (there is an assumption that an image classification dataset uses a consistent DB backend).

Good point, that is a little strange. 

> The thing is generic datasets don't have a training DB task. Where do you suggest I add the backend field?

In the `AnalyzeDbTask`'s.

PR #325, opened 2015-09-22
Change file inputs to look the same in chrome and firefox and have bo…

jmancewicz:
This is an attempt to unuglify the input file widgets in FireFox, and apply the bootstrap look. Now they look the same in FireFox and Chrome.

screenshot from 2015-09-22 13 43 51

### DIFF
--- a/digits/dataset/images/classification/forms.py
+++ b/digits/dataset/images/classification/forms.py
@@ -188,7 +188,7 @@ class ImageClassificationDatasetForm(ImageDatasetForm):
     textfile_use_local_files = wtforms.BooleanField(u'Use local files',
         default=False)
 
-    textfile_train_images = wtforms.FileField(u'Training images',
+    textfile_train_images = utils.forms.FileField(u'Training images',
             validators=[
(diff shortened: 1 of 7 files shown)


jmancewicz:
The buttons are big, but consistent with bootstrap look.  In the future, I wouldn't mind modifying bootstrap to be a little less of a real estate hog.

lukeyeager:
This isn't working for me with Chrome on Linux.

file-selector

Looks great with Chrome on Windows, though, and with Firefox on either.

jmancewicz:
Ok, I've tested on Firefox and Chrome on Linux and Firefox, Chrome, and Safari on OS X.  I had removed two lines from the css which cause it to break on one of the browsers.

I've also set the column widths to be 20%, 40%, 40% for the the textfile table to stop the columns from changing width when 'Use local paths on server' is toggled.

lukeyeager:
> Ok, I've tested on Firefox and Chrome on Linux and Firefox, Chrome, and Safari on OS X.

Verified. Thanks for the UI fix!

PR #268, opened 2015-09-03
Improvements for analyze_db tool

lukeyeager:
*See #264, #265*

- [x] Open LMDBs as read-only
- [x] Allow quick read for large databases
- [x] Write tests
- [x] Expose log like #222

Issue #21, opened 2015-03-23
Display images in dataset


lukeyeager:
In the web UI, load and display some of the images in the dataset to help users remember which image set this is.

PR #330, opened 2015-09-25
Support for train data augmentation

groar:
Added data augmentation to classification dataset creation. Only the train set is augmented at db creation with new images obtained by a combination of the following transformations : 
- Rotation 
- Translation
- Contrast modulation
- Hue modulation

There are a few parameters for each transformation, including the probability for each image of being augmented. 

Dataset augmentation

For now, only added tests to tools/create_db, but will add a few more regarding task creation and utils/image. Edit: done. 

Translating is quite slow, so I should check if the way I do it can be improved. 

### TODO before merge

- [ ] Fix hue modulation and translation
- [ ] Add the possibility of setting the number of images generated out of each training sample
- [ ] Fix the UI

### DIFF
--- a/digits/dataset/images/classification/forms.py
+++ b/digits/dataset/images/classification/forms.py
@@ -38,6 +38,158 @@ class ImageClassificationDatasetForm(ImageDatasetForm):
             tooltip='Compressing the dataset may significantly decrease the size of your database files, but it may increase read and write times.',
             )
 
+    ### Data augmentation - rotation 
+    has_augmentation_rotation = utils.forms.BooleanField('Perform rotations on images',
+            default = False,
+            tooltip = "You can augment your dataset by performing rotation on images.",
+            validators=[ 
+                ]
+            )
+
+    augmentation_rotation_angle_min = utils.forms.IntegerField(u'angle min',
+            default=-90,
+            validators=[ 
+                validators.NumberRange(min=-360, max=360)
+                ],
+            tooltip = "The minimum rotation angle that can be performed during augmentation."
+            )
+
+    augmentation_rotation_angle_max = utils.forms.IntegerField(u'angle max',
+            default=90,
+            validators=[ 
+                validators.NumberRange(min=-360, max=360)
+                ],
+            tooltip = "The maximum rotation angle that can be performed during augmentation."
+            )
+
+    augmentation_rotation_probability = utils.forms.FloatField(u'probability',
+            default=0.75,
+            validators=[ 
+                validators.NumberRange(min=0.0, max=1.0)
+                ],
+            tooltip = "The probability for an image to be rotated during augmentation."
+            )
+
+    ### Data augmentation - hue 
+    has_augmentation_hue = utils.forms.BooleanField('Perform hue modulation on images',
+            tooltip="You can augment your dataset by performing hue modulation on images.",
+            default=False,
+            validators=[ 
+                ]
+            )
+ 
+    augmentation_hue_angle = utils.forms.FloatField(u'angle',
+            default=0.5,
+            validators=[ 
+                validators.NumberRange(min=0.0, max=1.0)
+                ],
+            tooltip = "The maximum angle of the hue modulation."
+            )
+
+    augmentation_hue_angle_min = utils.forms.IntegerField(u'angle min',
+            default=66,
+            validators=[ 
+                validators.NumberRange(min=0, max=360)
+                ],
+            tooltip = "The minimum angle of the hue modulation."
+            )
+    augmentation_hue_angle_max = utils.forms.IntegerField(u'angle max',
+            default=122,
+            validators=[ 
+                validators.NumberRange(min=0, max=360)
+                ],
+            tooltip = "The maximum angle of the hue modulation."
+            )
+
+    augmentation_hue_probability = utils.forms.FloatField(u'probability',
+            default=0.75,
+            validators=[ 
+                validators.NumberRange(min=0.0, max=1.0)
+                ],
+            tooltip = "The probability for an image to be hue-modulated during augmentation."
+            )
+
+    ### Data augmentation - contrast
+    has_augmentation_contrast = utils.forms.BooleanField('Perform contrast modulation on images',
+            tooltip="You can augment your dataset by performing contrast modulation on images.",
+            default=False,
+            validators=[ 
+                ]
+            )
+ 
+    augmentation_contrast_strength_min = utils.forms.FloatField(u'strength min',
+            default=0.75,
+            validators=[ 
+                validators.NumberRange(min=0.0, max=2.0)
+                ],
+            tooltip = "The minimum strength of the contrast modulation."
+            )
+
+    augmentation_contrast_strength_max = utils.forms.FloatField(u'strength max',
+            default=1.25,
+            validators=[ 
+                validators.NumberRange(min=0.0, max=2.0)
+                ],
+            tooltip = "The maximum strength of the contrast modulation."
+            )
+
+    augmentation_contrast_probability = utils.forms.FloatField(u'probability',
+            default=0.75,
+            validators=[ 
+                validators.NumberRange(min=0.0, max=1.0)
+                ],
+            tooltip = "The probability for an image to be contrast-modulated during augmentation."
+            )
+
+    ### Data augmentation - translation
+    has_augmentation_translation = utils.forms.BooleanField('Perform translation on images',
+            tooltip="You can augment your dataset by performing translation on images.",
+            default=False,
+            validators=[ 
+                ]
+            )
+ 
+    augmentation_translation_dx_min = utils.forms.FloatField(u'dx min',
+            default=-0.5,
+            validators=[ 
+                validators.NumberRange(min=-1.0, max=1.0)
+                ],
+            tooltip = "The minimum dx of the translation."
+            )
+
+    augmentation_translation_dx_max = utils.forms.FloatField(u'dx max',
+            default=0.5,
+            validators=[ 
+                validators.NumberRange(min=-1.0, max=1.0)
+                ],
+            tooltip = "The maximum dx of the translation."
+            )
+
+    augmentation_translation_dy_min = utils.forms.FloatField(u'dy min',
+            default=-0.5,
+            validators=[ 
+                validators.NumberRange(min=-1.0, max=1.0)
+                ],
+            tooltip = "The minimum dy of the translation."
+            )
+
+    augmentation_translation_dy_max = utils.forms.FloatField(u'dy max',
+            default=0.5,
+            validators=[ 
+                validators.NumberRange(min=-1.0, max=1.0)
+                ],
+            tooltip = "The maximum dy of the translation."
+            )
+
+    augmentation_translation_probability = utils.forms.FloatField(u'probability',
+            default=0.75,
+            validators=[ 
+                validators.NumberRange(min=0.0, max=1.0)
+                ],
+            tooltip = "The probability for an image to be translation-modulated during augmentation."
+            )
+
+
     # Use a SelectField instead of a HiddenField so that the default value
     # is used when nothing is provided (through the REST API)
     method = wtforms.SelectField(u'Dataset type',
--- a/digits/dataset/images/classification/test_views.py
+++ b/digits/dataset/images/classification/test_views.py
@@ -66,7 +66,7 @@ class BaseViewsTestWithImageset(BaseViewsTest):
     IMAGE_WIDTH     = 10
     IMAGE_CHANNELS  = 3
     BACKEND         = 'lmdb'
-    COMPRESSION     = 'none' 
+    COMPRESSION     = 'none'
 
     UNBALANCED_CATEGORY = False
 
--- a/digits/dataset/images/classification/test_views.py
+++ b/digits/dataset/images/classification/test_views.py
@@ -66,7 +66,7 @@ class BaseViewsTestWithImageset(BaseViewsTest):
     IMAGE_WIDTH     = 10
     IMAGE_CHANNELS  = 3
     BACKEND         = 'lmdb'
-    COMPRESSION     = 'none'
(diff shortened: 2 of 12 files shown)


lukeyeager:
Wow this is awesome, great work! My little smoketests all look good so far, I'll try to dig into the code some more today and see if I have any feedback for you.

Now is when I really wish we had #21 so I could see this stuff in action ...

digits/dataset/images/classification/views.py
```diff
@@ -88,35 +88,70 @@ def from_folders(job, form):
...
+                    parents      = val_parents,
+                    input_file   = utils.constants.VAL_FILE,
+                    db_name      = utils.constants.VAL_DB,
+                    backend      = backend,
+                    image_dims   = job.image_dims,
+                    resize_mode  = job.resize_mode,
+                    encoding     = encoding,
+                    compression  = compression,
+                    labels_file  = job.labels_file,
+                    augmentation = {},
```
lukeyeager: You can just do here what you did below at line 171. I think it's better to keep the number of changed lines small than preserve indentation. I sort of regret hand-aligning indentation in general.

digits/dataset/tasks/create_db.py
```diff
@@ -11,7 +11,7 @@
 from digits.task import Task
 
 # NOTE: Increment this everytime the pickled version changes
-PICKLE_VERSION = 3
+PICKLE_VERSION = 4
```
lukeyeager: I don't think we need to increment the version here. We can just add this snippet to the end of `__set_state__()`:
```python
if not hasattr(self, 'augmentation'):
    self.augmentation = None
```

digits/templates/datasets/images/classification/show.html
```diff
@@ -101,6 +101,17 @@
     <b>Image Mean:</b>
     <img src="{{url_for('serve_file', path=task.path('mean.jpg', relative=True))}}" />
     {% endif %}
+    <br/>
+    <b>Augmentations:</b>
```
lukeyeager: This should only be shown if some augmentations exist. Otherwise you end up with this:

augmentations-none-display-issue

digits/utils/image.py
```diff
@@ -378,3 +504,75 @@ def get_color_map(name):
         greenmap    = [0,0,0.5,1,1,1,0.5,0,0]
         bluemap     = [0.5,1,1,1,0.5,0,0,0,0]
     return 255.0 * np.array(redmap), 255.0 * np.array(greenmap), 255.0 * np.array(bluemap)
+
+def rgb_to_hsv(rgb):
+    # Translated from source of colorsys.rgb_to_hsv
+    # r,g,b should be a numpy arrays with values between 0 and 255
+    # rgb_to_hsv returns an array of floats between 0.0 and 1.0.
```
lukeyeager: It looks like you're getting this code from [this SO post](http://stackoverflow.com/a/7274986/2404152). SO licenses user contributions with a viral license.

> user contributions licensed under cc by-sa 3.0 with attribution required

http://creativecommons.org/licenses/by-sa/3.0/

Unfortunately, that means you're going to need to re-implement this yourself or get permission from the original author to use this code :/

lukeyeager:
Really great stuff. I've mentioned a few nitpicks above (and the licensing thing - hopefully you can find a good solution for that). Otherwise, this LGTM!

@jmancewicz, any thoughts?

groar:
Wow, the SO license is a good point I should have checked ! I replaced those functions with matplotlib ones. It seems they are already implemented in Pillow (see https://github.com/python-pillow/Pillow/pull/816) but I couldn't figure out for sure. 

I also ignored hue modulation in the grayscale case since it does not make sense (we should do something better, like disabling the checkbox in the "new dataset" view). 

I think I have corrected all the things you've mentioned. Tell me if you see something else.

lukeyeager:
@groar, Awesome, thanks for the changes! Can you squash this down and remove the extra commits?

@gheinrich, @jmancewicz - I'd feel better if one of y'all took a look at this too before merging. I don't have time to test this exhaustively, but it looks pretty good to me.

digits/dataset/images/classification/test_views.py
```diff
@@ -66,7 +66,7 @@ class BaseViewsTestWithImageset(BaseViewsTest):
     IMAGE_WIDTH     = 10
     IMAGE_CHANNELS  = 3
     BACKEND         = 'lmdb'
-    COMPRESSION     = 'none'
+    COMPRESSION     = 'none'
```
gheinrich: I think trailing spaces should be avoided if possible (especially here since there is no functional change to this line)

digits/dataset/tasks/create_db.py
```diff
@@ -91,13 +93,15 @@ def __setstate__(self, state):
...
-                self.encoding = 'none'
+                self.encoding = 'none' 
         self.pickver_task_createdb = PICKLE_VERSION
 
         if not hasattr(self, 'backend') or self.backend is None:
             self.backend = 'lmdb'
         if not hasattr(self, 'compression') or self.compression is None:
             self.compression = 'none'
+        if not hasattr(self, 'augmentation'):
+            self.augmentation = None
```
gheinrich: maybe you need to do:
`self.augmentation = {}`
I had an error when loading an older dataset:
```
  File "/fast-scratch/gheinrich/ws/digits/digits/templates/datasets/images/classification/show.html", line 105, in template
    {% if task.augmentation|length != 0 %}
TypeError: object of type 'NoneType' has no len()
```
I think this is because self.augmentation is set to None here.

digits/dataset/images/classification/test_views.py
lukeyeager: Yeah, there's a lot of this. I use [this plugin](https://github.com/bronson/vim-trailing-whitespace) (shameless plug), which makes the whitespace issues super obvious.

gheinrich:
I have a question about the terminology here: in my understanding 'data augmentation' is a method to artificially create *many* unseen training examples by performing transformations on existing samples from the dataset. Am I understanding correctly that this PR does not actually increase the number of training samples? In this case is it technically correct to call this 'data augmentation'? I am asking candidly as I am not sure my understanding of the term is correct.

Either way, would it be useful to add an option to let the user specify a measure of how many images may be generated out of each training sample in the original dataset?

digits/dataset/images/classification/test_views.py
gheinrich: My text editor (visual slick) has an option to automatically remove trailing spaces.

digits/dataset/images/classification/test_views.py
```diff
@@ -442,6 +459,9 @@ class TestCreatedWide(TestCreated):
 class TestCreatedTall(TestCreated):
     IMAGE_HEIGHT = 20
 
+class TestCreatedAugmentation(TestCreated):
+    IMAGE_CHANNELS = 1
+
```
gheinrich: maybe add a test with IMAGE_CHANNELS=3 so you can test hue augmentation?

digits/dataset/images/classification/views.py
```diff
@@ -167,6 +203,37 @@ def from_files(job, form):
     if not image_folder:
         image_folder = None
 
+    augmentation = {
+    }
+
+    if form.has_augmentation_contrast.data:
+        augmentation['contrast'] = {
+            'strength': form.augmentation_contrast_strength.data,
```
gheinrich: I might be missing something but I think this field does not exist? Actually when trying to create a dataset from text files I get:
```
  File "/fast-scratch/gheinrich/ws/digits/digits/dataset/images/classification/views.py", line 211, in from_files
    'strength': form.augmentation_contrast_strength.data,
AttributeError: 'ImageClassificationDatasetForm' object has no attribute 'augmentation_contrast_strength'
```
Besides this code looks similar to lines 91 to 192 but is not exactly the same (I think the correct version is in the lines 91 to 192). Can you move this to a function and remove the duplicate code?

digits/dataset/images/classification/forms.py
```diff
@@ -38,6 +38,158 @@ def validate_backend(form, field):
...
+
+    ### Data augmentation - hue 
+    has_augmentation_hue = utils.forms.BooleanField('Perform hue modulation on images',
+            tooltip="You can augment your dataset by performing hue modulation on images.",
+            default=False,
+            validators=[ 
+                ]
+            )
+ 
+    augmentation_hue_angle = utils.forms.FloatField(u'angle',
```
gheinrich: is this really needed?

digits/dataset/images/classification/views.py
```diff
@@ -167,6 +203,37 @@ def from_files(job, form):
...
+            'probability': form.augmentation_contrast_probability.data
+        }
+
+    if form.has_augmentation_hue.data and job.image_dims[2] == 3:
+        augmentation['hue'] = {
+            'angle': form.augmentation_hue_angle.data,
```
gheinrich: do you mean to use `hue_angle_min` and `hue_angle_max`? The field `hue_angle` doesn't seem to be rendered in the HTML template...

digits/dataset/tasks/create_db.py
groar: Probably yes, will fix that.

digits/dataset/images/classification/test_views.py
groar: Ok I finally installed a sublime plugin to highlight those trailing spaces! I should be able not to add any new trailing space at least.

digits/dataset/images/classification/test_views.py
groar: Hum yeah I don't know why I added those lines, they are testing nothing. The real tests are done in `TestAugmentedDataset`. I will test the different augmentations yes.

digits/dataset/images/classification/views.py
groar: This is a residual from a previous version, well spotted. I'll remove it.

digits/dataset/images/classification/forms.py
groar: What do you mean? The `has_augmentation_hue` field?

digits/dataset/images/classification/views.py
groar: Yeah this is related to the other remark you made.

groar:
> Would it be useful to add an option to let the user specify a measure of how many images may be  generated out of each training sample in the original dataset? (as opposed to just the probability that each transformation is applied)

@gheinrich Yes this is indeed important. We should have both a given probability and the possibility of getting several images out of an original training sample for a given transformation. I will do that as soon as I get some time.

digits/dataset/images/classification/forms.py
lukeyeager: I think he means the `augmentation_hue_angle` field. Why does a hue have an angle?

lukeyeager:
@groar, it sounds like you have some more things to do before you feel like this is ready to merge.

Can you add a TODO list to the original post (like #209, #226, #268) so we can see the progress?

digits/dataset/images/classification/forms.py
gheinrich: I meant that the field `augmentation_hue_angle` appears to be unused (`augmentation_hue_angle_min` and `augmentation_hue_angle_max` are used instead).

jmancewicz:
First off, this is awesome... but I have a few initial thoughts

A little thing, just about the code.  Can you try to not leave trailing whitespaces.  Future changes, to whitespace can cause merge and ownership issues. It's a good practice to strip them before commit.

I think it makes sense to move the transformation matrix operations together.  I suggest moving 'Perform translation on images' beneath 'Perform rotations on images'.  Also, and this is being picky, but translation is singular and rotations is plural.  Actually, I'd skip changing that and make the phrases more succinct.  Perhaps...

augment

or

augment2

That might be too vague, but I doubt it. The tooltips make everything clear.

Generally, field labels are capitalized [Image size] (some are title style [Resize Transformation], and maybe should be changed), but with the exception of the 'dx' and 'dy', I recommend capitalization of these labels. 

I'm conflicted about the unit of dx and dy.  I'm fine with it being in normalized image units, but wonder if % would make it clearer what's going to happen to the images.

I'll dig into the code now.

jmancewicz:
Would it be better to not use clamp to edge wrap mode?  Is it better to use a solid color?

translate

jmancewicz:
Also, the hue modulation does not appear to be modulation.  It looks like it's setting the hue to a constant for each new image.  That might be fine, but I expected it to rotate the hue.

by changing the code from

    hsv[:,:, 0] = angle / 360.0
to
    hsv[:,:, 0] += angle / 360.0
    hsv[:,:, 0] %= 1
or if you wanted to extend the range from [0,360] to [-360,360], then 
    hsv[:,:, 0] += angle / 360.0 + 1.0
    hsv[:,:, 0] %= 1
should do.  That's all assuming that you meant to shift the hue.  If the intent was to set the hue, then maybe we shouldn't call it modulation.  I'm not familiar with how hue affects training.  It almost seems like a rgb mult would do better for simulating varying lighting conditions.

jmancewicz:
Yes, the translation is very slow.  I wonder if it is from the clamp to edge wrap mode.  I expect this could be made much faster using OpenGL.

groar:
@jmancewicz 

- you're right about the hue, it should be a shift, not just setting it. It's not completely clear though what transformations work best in general as it heavily depends on the task (I don't know of any comprehensive study on data jittering, even on specific tasks. If you know any let me know I'm interested !). I personally had significant improvements on classification and regression tasks by modulating the hue, as well as by tinting images (by simply setting the hue). 

In fact, I think it would be even better to perform general HSV modulation (so it can be used to produce several useful transformations, see for example: http://papers.nips.cc/paper/5548-discriminative-unsupervised-feature-learning-with-convolutional-neural-networks.pdf). 

- Concerning the translation, you're right, it should be random noise or solid color.
- I also agree with changing the labels. I'm not sure about the unit of dx and dy though.

@lukeyeager yes that's a good idea, I'll do that.

lukeyeager:
Closing as abandoned.

PR #336, opened 2015-09-30
Allow DIGITS to parse NVcaffe version strings

lukeyeager:
Previously, it totally failed to parse "0.14.0-alpha", which is valid semantic versioning.

This PR uses `pkg_resources` from `setuptools` to do the comparison.

\<rant>

Technically, `setuptools` adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/) and not [Semantic Versioning 2.0.0](http://semver.org/spec/v2.0.0.html), but I can't find any package/tool which (1) actually follows Semantic Versioning correctly when the version is correct, and (2) still does something useful when it's a little bit off.

Things I tried which don't work quite right:
* raw string comparisons
* git built-in (`git tag`)
* GitHub
* `distutils.version.LooseVersion`
* `semantic_version.Version`
* `pkg_resources.parse_version` (close enough)

https://gist.github.com/lukeyeager/9922a2229f7a03d1463c

\</rant>

### DIFF
--- a/digits/config/caffe_option.py
+++ b/digits/config/caffe_option.py
@@ -6,6 +6,7 @@ import sys
 import imp
 import platform
 import subprocess
+import pkg_resources
 
 from digits import device_query
 import config_option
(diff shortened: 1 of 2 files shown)


gheinrich:
This looks good to me!

PR #338, opened 2015-10-02
Fix warning with setuptools >= 8.0

### LINKED ISSUE #337: No response after starting the server
I got the following RuntimeWarning after starting the server by `./digits-devserver`, and then there is no response. The thread is in uninterruptible sleep state.

```
/home/ubuntu/anaconda/lib/python2.7/site-packages/setuptools-18.3.2-py2.7.egg/pkg_resources/__init__.py:203: RuntimeWarning: You have iterated over the result of pkg_resources.parse_version. This is a legacy behavior which is inconsistent with the new version class introduced in setuptools 8.0. In most cases, conversion to a tuple is unnecessary. For comparison of versions, sort the Version instances directly. If you have another use case requiring the tuple, please file a bug with the setuptools project describing that need.

Saved config to /home/ubuntu/DIGITS/digits/digits.cfg
```

lukeyeager:
*Fix #337*

Add lots of tests, too.

### DIFF
--- a/digits/config/caffe_option.py
+++ b/digits/config/caffe_option.py
@@ -6,9 +6,9 @@ import sys
 import imp
 import platform
 import subprocess
-import pkg_resources
 
 from digits import device_query
+from digits.utils import parse_version
 import config_option
 import prompt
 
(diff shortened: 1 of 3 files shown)


lukeyeager:
> Do you need setuptools>=8.0? The requirements.txt file still points to setuptools>=3.3

@gheinrich This block works with 3.3 - 7.8, and also for 8.0+:
```python
    try:
        return pkg_resources.SetuptoolsVersion(v)
    except AttributeError:
        return pkg_resources.parse_version(v)
```

PR #341, opened 2015-10-02
Use new multi-GPU flag for NVcaffe v0.14

lukeyeager:
Upstream caffe decided to use the "--gpu" flag instead of the original "--gpus" flag which was merged into NVcaffe.

### DIFF
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -703,7 +703,11 @@ class CaffeTrainTask(TrainTask):
             if len(identifiers) == 1:
                 args.append('--gpu=%s' % identifiers[0])
             elif len(identifiers) > 1:
-                args.append('--gpus=%s' % ','.join(identifiers))
+                if config_value('caffe_root')['version'] < utils.parse_version('0.14.0-alpha'):
+                    # Prior to version 0.14, NVcaffe used the --gpus switch
+                    args.append('--gpus=%s' % ','.join(identifiers))
+                else:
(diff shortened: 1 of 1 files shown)


digits/model/tasks/caffe_train.py
```diff
@@ -703,7 +703,11 @@ def task_arguments(self, resources, env):
             if len(identifiers) == 1:
                 args.append('--gpu=%s' % identifiers[0])
             elif len(identifiers) > 1:
-                args.append('--gpus=%s' % ','.join(identifiers))
+                if config_value('caffe_root')['version'] < utils.parse_version('0.13.99'):
```
gheinrich: would this work too: 
if config_value('caffe_root')['version'] < utils.parse_version('0.14')
I don't suppose the odds of seeing a 0.13.991 version are high but since you're using a strict inequality, why not compare against 0.14?

lukeyeager: Because the actual version is `0.14.0-alpha`, which is sorted **earlier** than `0.14.0`. That's correct according to most versioning schemes, but it makes it a little trickier to write this kind of check. I could just go ahead and use `0.14.0-alpha` instead of `0.13.99`. Does that seem better?

gheinrich:
Thanks for the explanation. Comparing against `0.14.0-alpha` sounds slightly more explicit to me but I leave this decision to you.

lukeyeager:
Switched to using `alpha` per your suggestion. Merging.

PR #351, opened 2015-10-10
Fix checks in bash script to throw useful errors

lukeyeager:
Previously, if you didn't have nosetests installed, ./digits-test would
fail like this:

    $ ./digits-test
    Unknown option: --
    usage: /home/lyeager/digits/venv/bin/python [option] ... [-c cmd | -m
    mod | file | -] [arg] ...
    Try `python -h' for more information.

Now, it would fail like this:

    $ ./digits-test
    ERROR: "nosetests" executable not found!

### DIFF
--- a/digits-server
+++ b/digits-server
@@ -1,7 +1,26 @@
 #!/bin/bash
 # Copyright (c) 2014-2015, NVIDIA CORPORATION.  All rights reserved.
 
-$(command -v python) $(command -v gunicorn) \
+set -e
+
+function set_exe {
+    # Sets a global variable to the location of an executable
+    # Arguments:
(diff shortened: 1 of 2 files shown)


lukeyeager:
@jmancewicz can you test the `set_exe` function on a mac? I think `hash` and `command` should work the same way, but I'd like to have a little confirmation before merging this. No need to actually install Caffe/DIGITS - just see if that bash syntax still works. Thanks!

jmancewicz:
I didn't have nose tests installed and set_exe worked as expected.

lukeyeager:
> I didn't have nosetests installed and set_exe worked as expected.

Perfect, thanks!

PR #366, opened 2015-10-14
Display InferenceErrors like all other exceptions

lukeyeager:
@gheinrich in #249 you added this try/except block for `InferenceError`s. Why?

Current behavior for errors:
inference-error-before

After this change, it matches the format for other exceptions:
inference-error-after

In debug mode, you get more information:
inference-error-after-debug

### DIFF
--- a/digits/model/images/classification/views.py
+++ b/digits/model/images/classification/views.py
@@ -272,10 +272,7 @@ def image_classification_model_classify_one():
         layers = 'all'
 
     predictions, visualizations = None, None
-    try:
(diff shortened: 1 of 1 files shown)


gheinrich:
Yes, that was a bad idea. Thanks for the fix and sorry for the bug!

lukeyeager:
No problem. I'll merge this when TravisCI passes.

lukeyeager:
Also, do you have any idea what that error actually means? I'm trying to debug issues with #321.

PR #367, opened 2015-10-15
Fix some of the open file issues with Torch

lukeyeager:
When running the image classification model tests like this:

    ./digits-test digits/model/images/classification/test_views.py

the number of open files for the test suite grows to > 900. You can verify this with:

    while :; do lsof -p $PID | grep tmp | wc -l; sleep 1; done

where PID is the process ID of the test suite.

This patch brings the number of open files down from > 900 to ~90. There's more work to be done to fix this completely, but this takes a big step in the right direction.

NOTE: I can't actually run the whole testsuite or I get a bunch of these OSErrors:

    OSError: [Errno 24] Too many open files

### DIFF
--- a/digits/frameworks/torch_framework.py
+++ b/digits/frameworks/torch_framework.py
@@ -94,9 +94,9 @@ class TorchFramework(Framework):
         return visualization of network
         """
         # save network description to temporary file
-        _, temp_network_path = tempfile.mkstemp(suffix='.lua')
-        with open(temp_network_path, "w") as outfile:
-            outfile.write(desc)
+        temp_network_handle, temp_network_path = tempfile.mkstemp(suffix='.lua')
+        os.write(temp_network_handle, desc)
+        os.close(temp_network_handle)
 
         try: # do this in a try..finally clause to make sure we delete the temp file
             # build command line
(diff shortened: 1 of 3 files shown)


lukeyeager:
This is a pretty big deal. If you want to train with Torch and keep the server running for a long time, eventually the server is going to die with the above OSError message.

@gheinrich, can you verify this on your system? I'm not doing anything weird on my machine, am I?

gheinrich:
Thanks Luke, I hadn't noticed this, probably because my limit is ridiculously high on my system:
```
$ ulimit -Sn
1048576
```
I don't know where the default value comes from but I think Travis machines are also running Ubuntu 12.04 and perhaps this explains why tests are running well on Travis.
I made more changes along the same lines of your patch in other places as well, hopefully this covers all the `mkstemp` ones:
https://github.com/gheinrich/DIGITS/tree/dev/pr-367

lukeyeager:
Ah yes, that'll do it. Here's my limit:
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.3 LTS
Release:	14.04
Codename:	trusty

$ ulimit -Sn
1024
```

lukeyeager:
Awesome, your changes bring the max number of open files down from 900 -> 90 -> 11. That sounds much more reasonable. I'll merge after Travis passes.

PR #376, opened 2015-10-19
Fix parsing of infinite loss in Torch

gheinrich:
Lua uses 'inf' to represent infinite numbers. Additionally Torch may use
'nan' in its representation of infinite values in Tensors.

We must catch both cases in the parsing of training and validation
losses.

Add test to verify functionality.

### DIFF
--- a/digits/model/images/generic/test_views.py
+++ b/digits/model/images/generic/test_views.py
@@ -73,6 +73,8 @@ local net = nn.Sequential()
 net:add(nn.MulConstant(0.004))
 net:add(nn.View(-1):setNumInputDims(3))  -- 1*10*10 -> 100
 net:add(nn.Linear(100,2))
+-- remove any non determinism
+net:apply(function(layer) if layer.weight then layer.weight:fill(0) end if layer.bias then layer.bias:fill(0) end end)
 return function(params)
     return {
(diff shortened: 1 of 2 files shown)


gheinrich:
Note that I was unable to get the new test to pass for Caffe on the Travis machine (it did always pass on my machine). Maybe that has something to do with CuDNN support (which I suppose is not compiled in on the Travis machine)?

lukeyeager:
> Note that I was unable to get the new test to pass for Caffe on the Travis machine

Looks like they passed to me ... Did you restart it?

> CuDNN support (which I suppose is not compiled in on the Travis machine)?

That's true. Travis builds caffe without CUDA or cuDNN.
https://github.com/NVIDIA/DIGITS/blob/v2.2.1/scripts/travis/install-caffe.sh#L30-L31

lukeyeager:
Tests pass on my machine and they seem to pass on Travis despite your comment above. Merging.

gheinrich:
I forgot to mention I [skip](https://github.com/gheinrich/DIGITS/blob/58e07e883675d9608215daf1f019db78fe5a20fb/digits/model/images/generic/test_views.py#L359) the test in Caffe case.

Issue #346, opened 2015-10-06
Torch tests non-deterministic


lukeyeager:
It looks like the image classification tests for Torch aren't always passing.

https://travis-ci.org/NVIDIA/DIGITS/builds/83389812
```
======================================================================
FAIL: digits.model.images.classification.test_views.TestTorchCreatedHdf5.test_classify_one
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/miniconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/NVIDIA/DIGITS/digits/model/images/classification/test_views.py", line 464, in test_classify_one
    assert predictions[0][1] == category, 'image misclassified'
AssertionError: image misclassified
-------------------- >> begin captured logging << --------------------
digits.webapp: INFO: torch classify one task started.
digits.webapp: DEBUG: torch classify one task : Loading network definition from /tmp/tmpy7uo1W/20151002-224746-faf2/model
digits.webapp: DEBUG: torch classify one task : Loading /tmp/tmpy7uo1W/20151002-224746-faf2/snapshot_10_Weights.t7 file
digits.webapp: INFO: torch classify one task completed.
--------------------- >> end captured logging << ---------------------

======================================================================
FAIL: digits.model.images.classification.test_views.TestTorchCreatedHdf5.test_classify_one_json
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/miniconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/NVIDIA/DIGITS/digits/model/images/classification/test_views.py", line 483, in test_classify_one_json
    assert data['predictions'][0][0] == category, 'image misclassified'
AssertionError: image misclassified
-------------------- >> begin captured logging << --------------------
digits.webapp: INFO: torch classify one task started.
digits.webapp: DEBUG: torch classify one task : Loading network definition from /tmp/tmpy7uo1W/20151002-224746-faf2/model
digits.webapp: DEBUG: torch classify one task : Loading /tmp/tmpy7uo1W/20151002-224746-faf2/snapshot_10_Weights.t7 file
digits.webapp: INFO: torch classify one task completed.
--------------------- >> end captured logging << ---------------------
```

gheinrich:
I think there are still some rare occasions when this test fails despite #348. I will keep looking into it.

[...]

PR #378, opened 2015-10-20
Use fixed seed in model tests

gheinrich:
Setting weights and biases to zero initially. This applies to the
model used in all Torch classification tests except those that use
the standard LeNet model.

Tested with:
./digits-test -v --with-coverage --cover-package=digits -s --tests=digits.model.images.classification

Without patch:
10 failures / 128 iterations

With patch:
1 failure / 160 iterations
Note: the one failure was on a test that uses the LeNet model,
which isn't modified by this patch. That failure has to do with
the pretty aggressive learning rate policy, which is set to 'fixed'
for this test. Log shows model learning up to 100% validation
accuracy then regressing to 66% accuracy on last epoch.

see bug #346

### DIFF
--- a/digits/model/images/classification/test_views.py
+++ b/digits/model/images/classification/test_views.py
@@ -156,6 +156,7 @@ class BaseViewsTestWithDataset(BaseViewsTest,
                 'batch_size':       10,
                 'train_epochs':     cls.TRAIN_EPOCHS,
                 'framework' :       cls.FRAMEWORK,
+                'random_seed':      0xCAFEBABE,
                 'shuffle':          'true' if cls.SHUFFLE else 'false'
                 }
         if cls.CROP_SIZE is not None:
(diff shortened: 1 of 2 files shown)


gheinrich:
New patch uploaded based on @lukeyeager's idea. I ran 100 iterations (83400 tests) of the full test suite and still had some failures, but at a much lower rate than initially:

` ` `
digits.model.images.classification.test_views.TestCaffeLeNet.test_classify_one ...
digits.model.images.classification.test_views.TestCaffeLeNet.test_classify_one_json ...
digits.model.images.classification.test_views.TestTorchCreatedHdf5Shuffle.test_classify_one
digits.model.images.classification.test_views.TestTorchCreatedHdf5Shuffle.test_classify_one_json
` ` `

lukeyeager:
Using the same random seed seems like a good idea to me. I'd still like to get this down to no errors at all, obviously, but I'll go ahead and merge this since it seems to help.

PR #379, opened 2015-10-21
Add workspaces

ahmedosman:
Hi Luke,  

This pull request includes two features:

1- Workspaces: Create Data and Classification Jobs inside workspaces.
**Create/Delete a workspace, all jobs (data/classification) are created inside a specific workspace and can only be used within the workspace. We changed only the hierarchy of the jobs directory to be digits/$workspace_name/$job_id. We are using one instance of the digits server across all workspaces.
** There is always a default workspace, that can't be delete.  

2- Feature Extraction/Filter visualisation:
**Given a trained model, you can upload a single/many image. Perform prediction on the images and export layer wise filter weights, activation values. Users can specify *all* for exporting/visualizing all layers information, or specific a single layer name or specify a subset of layers as comma seperate entries (i.e conv1,conv2,fc7) 
**Visualization are displayed in browser.
**Data can be export as npy (python numpy array) or .mat file (matlab cell array).

Ahmed


digits-server
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 # Copyright (c) 2014-2015, NVIDIA CORPORATION.  All rights reserved.
 
+/usr/bin/pip --upgrade gunicorn
```
lukeyeager: We definitely don't want to force people to upgrade gunicorn. What is this here for? Do we need to increase the minimum requirement in `requirements.txt`?

digits-server
```diff
@@ -1,7 +1,8 @@
...
 $(command -v python) $(command -v gunicorn) \
-    --config gunicorn_config.py \
+   --reload --pythonpath /root/digits --config gunicorn_config.py \
```
lukeyeager: Why are you calling `--reload` and setting the PYTHONPATH here?

lukeyeager:
Hi @ahmedosman, thanks for this huge PR from @batra-mlp-lab! Nice work!

1. Try rebasing this on top of the current `master` branch. There seem to me some merge conflicts.

2. We're going to need some tests for additions of this magnitude.

3. Separate features should be submitted as separate pull requests.

lukeyeager:
Here's a starter list of things to address:

1. You've checked in several `.DS_Store` files. Those shouldn't be there.

2. There are a lot of files under `mockup_visualization/c3/htdocs/`. Do we need all of those files?

gunicorn_config.py
```diff
@@ -1,5 +1,6 @@
 # Copyright (c) 2014-2015, NVIDIA CORPORATION.  All rights reserved.
-
+import sys
+sys.path.append("/root/digits/digits")
```
lukeyeager: Definitely don't want to hardcode any paths here. That's a strange path anyway.

ahmedosman:
Hi @lukeyeager  Thanks for the feedback. Let me look in those issues and get back to you.

lukeyeager:
Closing as abandoned.

PR #380, opened 2015-10-22
remove ago from running jobs

jmancewicz:
Meant to add this to #240.


lukeyeager:
Sounds good.

> Removing the 'ago' text which shows how long ago the job started. There are potential time computation issues between client and server that need to be resolved. This will most likely will return in the future.
https://github.com/NVIDIA/DIGITS/pull/240#issuecomment-150319673

PR #394, opened 2015-11-02
Reduce Torch7 install verbosity

gheinrich:
On DIGITS ToT, Travis logs are close to 4MB in size, which is the
hard limit for log size on Travis. Longer logs would yield an error.

This change causes the Torch7 install log to only be printed on an
install error.

This reduces the size of a typical log down to 250kB and 4000 lines,
meaning that the full log can be displayed in real time since it is
less than 10,000 lines.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ install:
     - ./scripts/travis/install-caffe.sh $(pwd)/deps/caffe
 
     # Torch
-    - ./scripts/travis/install-torch.sh $(pwd)/deps/torch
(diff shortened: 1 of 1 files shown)


lukeyeager:
Excellent!

lukeyeager:
@jmancewicz is seeing build timeouts. I think we need to add a `travis_wait` before the install command.

http://docs.travis-ci.com/user/build-timeouts/#Build-times-out-because-no-output-was-received

gheinrich:
Sorry! I realize it usually takes 9 minutes and something to install Torch so if the network is just a little bit slower you hit the 10-minute limit. I'll push a patch.

PR #395, opened 2015-11-02
Don't crop image before feeding to DL framework

gheinrich:
When using 'crop in form' method, images are cropped in Python
code before 'classify one'. This is slightly inconsistent as
this operation is not performed during training or in the
context of 'classify many'. Besides, this is problematic for
Torch (at least) since the mean image is not cropped and this
makes it difficult to perform mean subtraction during 'classify
one' as the input and mean image have different sizes.

Caffe uses the 'crop_size' option in the data layer transformations.
Similarly, for Torch the crop command-line flag may be used to
crop the image within Torch. Either way, it is not necessary
to crop the image in Python code.

### DIFF
--- a/digits/model/images/classification/views.py
+++ b/digits/model/images/classification/views.py
@@ -265,9 +265,6 @@ def image_classification_model_classify_one():
     db_task = job.train_task().dataset.train_db_task()
     height = db_task.image_dims[0]
     width = db_task.image_dims[1]
-    if job.train_task().crop_size:
-        height = job.train_task().crop_size
-        width = job.train_task().crop_size
     image = utils.image.resize_image(image, height, width,
             channels = db_task.image_dims[2],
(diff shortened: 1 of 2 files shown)


lukeyeager:
Hmm, that is inconsistent. Thanks for pointing this out. If we're going to remove pre-cropping from the classification path, then we should probably do it for the generic path as well:

https://github.com/NVIDIA/DIGITS/blob/98b729869f/digits/model/images/generic/views.py#L249-L251
https://github.com/NVIDIA/DIGITS/blob/98b729869f/digits/model/images/generic/views.py#L301-L303

One thing to note is that with this change, images are resized twice. That's a bit of a bummer for performance, but not that big of a deal.

gheinrich:
Thanks! I had overlooked the generic inference part, which probably doesn't work for Torch when doing 'crop in form'. I need to add tests for that too.

PR #396, opened 2015-11-02
Reduce Torch install verbosity

gheinrich:
As suggested by @lukeyeager prepending the install command
with `travis_wait` will print a log message every minute
for 20 minutes, allowing Torch install to take a little more
than the usual <10 minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ install:
     - ./scripts/travis/install-caffe.sh $(pwd)/deps/caffe
 
     # Torch
-    - ./scripts/travis/install-torch.sh $(pwd)/deps/torch
+    - travis_wait ./scripts/travis/install-torch-wrapper.sh $(pwd)/deps/torch $(pwd)/torch-install-log.txt
 
     # DIGITS
     - sudo apt-get install graphviz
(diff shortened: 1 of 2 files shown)


gheinrich:
I was not able to verify whether this is working as on the last instance it took less than 10 minutes to install Torch.

gheinrich:
Also I couldn't simply prepend `travis_wait` to the command-line as the output redirection I was using for the Torch install script was also catching the log messages from the `travis_wait` command. So I moved the command to install Torch to a shell script.

gheinrich:
Also tested with debug patch which adds `sleep 180` to Torch install script:
https://github.com/gheinrich/DIGITS/commit/166fdb300a0c362feaabe5c5c09520e65f84468d
Corresponding Travis build:
https://travis-ci.org/gheinrich/DIGITS/builds/88911845
Torch install took 745s but there was no timeout.

lukeyeager:
I think this solution is acceptable, but let's create the file in `scripts/travis/` rather than creating it manually in the `.travis.yml` with `echo` commands.

Also, you need to rebase after https://github.com/NVIDIA/DIGITS/pull/397.

gheinrich:
Thanks Luke I have created a wrapper to the torch install script on `scripts/travis/install-torch-wrapper.sh`

PR #403, opened 2015-11-06
Import HDF5 from main thread

gheinrich:
Doing it from a worker thread is unsafe.
Reference:
https://github.com/deepmind/torch-hdf5/issues/42

### DIFF
--- a/tools/torch/data.lua
+++ b/tools/torch/data.lua
@@ -2,6 +2,7 @@
 require 'torch' -- torch
 require 'nn' -- provides a normalization operator
 require 'utils' -- various utility functions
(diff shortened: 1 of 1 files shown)


lukeyeager:
@moconnor725 can you try this out on your machine?

soumith:
added more details in deepmind/torch-hdf5#42 , tl;dr is that it's a hdf5 bug. 1.8.13 has the issue, 1.8.15 doesn't.

moconnor725:
https://github.com/NVIDIA/DIGITS/pull/403 does fix my problem

PR #405, opened 2015-11-06
Add the ability to delete multiple models at once

jmancewicz:
Added a button to the model details page.  There's a dataTable of existing model jobs which can be filtered so a smaller list. The `Delete Selected Jobs` will then delete the items in the filtered list.


gheinrich:
That's very useful indeed! I often find myself wanting to delete many test jobs at once. One comment I have is it wasn't entirely clear to me how to select jobs until I read your comment:
> The Delete Selected Jobs will then delete the items in the filtered list.

Usually when there is a button `Delete Selected Jobs` one would expect a way to select jobs through check boxes. Would it be a good idea to add a check box for each job on the list (plus one for "select all")? I feel it would be slightly more straightforward.

lukeyeager:
> Would it be a good idea to add a check box for each job on the list (plus one for "select all")?

:+1:

jmancewicz:
I totally agree.

jmancewicz:
image

Rather than adding checkboxes, I added extended selection to the table.  I think it feels more useful than the checkboxes.  Plus the checkboxes fight the datatables. The Delete Selected Jobs button will delete the intersection of selected and filtered items.

jmancewicz:
@lukeyeager and @gheinrich, can you guys have a look at this when you have time?

gheinrich:
I like the GUI. I'm looking at the code now.

lukeyeager:
I don't think we should take a new feature for v3.0 this late in the game. Let's put it off until we mark `v3.1-dev` on the `master` branch. Looks cool though!

digits/static/js/table_selection.js
```diff
@@ -0,0 +1,210 @@
...
+    this.mousedown_last_rowIndex = -1;
+
+    function on_selection_changed() {
+        var e = $.Event('selection_changed');
+        e.count = $(TableSelection.current_table + ' tr.selected').length;
+        $(window).trigger(e);
+    }
+
+    // Disablew text selection because we want to select rows not text
+    $(table_id + ' tbody').disableSelection();
```
gheinrich: this means we can't copy/paste from the table but I don't think this is an issue, is it?

digits/static/js/table_selection.js
```diff
@@ -0,0 +1,210 @@
...
+
+    // track the current table for key events
```
gheinrich: are these "mouse" events?

digits/static/js/table_selection.js
gheinrich: actually it is possible to select text when starting the selection from outside the table.

digits/static/js/table_selection.js
```diff
@@ -0,0 +1,210 @@
...
+        }
+    });
+
+    // Because the key event is global, only register once.
+    if (!TableSelection.initialized) {
+        TableSelection.initialized = true;
+        $(document).keydown(function(e) {
+            if (TableSelection.current_table == null) return;
+            // Use alt-a for select all rows
+            if ((e.ctrlKey || e.metaKey) && e.keyCode == 65) {
```
gheinrich: Ctrl-A isn't working for me if there are already some selected rows.
By the way is it common practice to use Ctrl-A (v.s. Ctrl-a)?

digits/static/js/table_selection.js
```diff
@@ -0,0 +1,210 @@
...
+                e.preventDefault();
+                deselect($(TableSelection.current_table + ' tr'));
+                that.last_selected_row = null;
+            }
+            // Use alt-i for invert selection
+            if ((e.ctrlKey || e.metaKey) && e.keyCode == 73) {
+                e.preventDefault();
+                $(TableSelection.current_table + ' tr.selected').addClass('swap');
+                $(TableSelection.current_table + ' tr').not('swap').addClass('selected');
+                $(TableSelection.current_table + ' tr.swap').removeClass('swap selected');
```
gheinrich: nice trick!

gheinrich:
Except for a few minor comments, looks good to me, well done!

digits/static/js/table_selection.js
jmancewicz: You are correct. You can not copy and paste. That is the trade off, but I see this style of selection far more useful.  If you notice, I took over Alt-A (Ctrl-A) select all text when the pointer is within the rectangle of a table.

digits/static/js/table_selection.js
jmancewicz: Which? There are mouse and key events.

digits/static/js/table_selection.js
jmancewicz: It's intended to be ctrl-a. I don't know if ANYone supports ctrl-A explicitly, but javascript events *appear* to issue the same key code despite case a => 65 and A => 65 and shftKey == true.  So, as it's written, ctrl-A and ctrl-a should both work.

When you try ctrl-a, is the pointer over the table?  I'm setting the current_table on the enter and leave events.  If the current_table is null, then it doesn't work.  The reason that I don't allow it to be a document wide hot-key, is for the case where there are multiple tables on the same document.  I worked that up in my test case, and it works.  The key events in jquery didn't work for me if I tried to limit it to the table with $('#my_table').keydown(...); it only seemed to work with document.  If you guys know better, please let me know.

digits/static/js/table_selection.js
jmancewicz: Do you mean that the variable names might be named better with mouse than mousedown? The first four are state at the time of the mousedown event.  I should look into the name of that last one `mousedown_last_rowIndex`.

gheinrich: sorry I think it's OK. I had incorrectly understood the code, now I understand the `mouseenter` event is used to track the current table to handle next key events properly.

digits/static/js/table_selection.js
gheinrich: In Chrome and in Firefox, when I select one entry in the table and then do Ctrl-a (with the pointer over the table), none of the other rows get selected. If I then to Ctrl-d (everything gets deselected as expected) and Ctrl-a (everything gets selected as expected).

jmancewicz: Funny that you mention Chrome and Firefox, because I'd been developing using Safari, and when I went to test on Chrome and Firefox, several things stopped working, because Safari still supported deprecated javascript.

This wasn't from that though. I"d tried to minimize the selection_changed events, and had a fault  in the logic.  Thanks for catching it.  I've tested the fix on Safari, Firefox, and Chrome.

gheinrich: Thanks, ctrl-a is working as expected on your latest patch.

lukeyeager:
I just used this to cleanup a bunch of junk jobs. Works great!

I don't love that it requires adding a ton of files, nor that there are no tests ... maybe there's nothing we can do about it?

lukeyeager:
@jmancewicz can you minimize `jquery-ui.js`? Then I think we can merge this.

I was worried about adding a bunch of extra files, but it's not as excessive as I thought. This PR brings the size of `digits/static/` up from 1.3MB to 1.8MB. That seems acceptable to me.

lukeyeager:
Closing in favor of https://github.com/NVIDIA/DIGITS/pull/728.

PR #406, opened 2015-11-06
Move scheduler.jobs to a dict.

jmancewicz:
Just moving the `scheduler.jobs` list from a `list` to a `dict` indexed by `job.id()`.

### DIFF
--- a/digits/model/images/classification/views.py
+++ b/digits/model/images/classification/views.py
@@ -475,7 +475,7 @@ def image_classification_model_top_n():
 
 def get_datasets():
     return [(j.id(), j.name()) for j in sorted(
-        [j for j in scheduler.jobs if isinstance(j, ImageClassificationDatasetJob) and (j.status.is_running() or j.status == Status.DONE)],
+        [j for j in scheduler.jobs.values() if isinstance(j, ImageClassificationDatasetJob) and (j.status.is_running() or j.status == Status.DONE)],
         cmp=lambda x,y: cmp(y.id(), x.id())
         )
         ]
@@ -493,14 +493,14 @@ def get_default_standard_network():
 
 def get_previous_networks():
     return [(j.id(), j.name()) for j in sorted(
-        [j for j in scheduler.jobs if isinstance(j, ImageClassificationModelJob)],
+        [j for j in scheduler.jobs.values() if isinstance(j, ImageClassificationModelJob)],
         cmp=lambda x,y: cmp(y.id(), x.id())
         )
         ]
 
 def get_previous_networks_fulldetails():
     return [(j) for j in sorted(
-        [j for j in scheduler.jobs if isinstance(j, ImageClassificationModelJob)],
+        [j for j in scheduler.jobs.values() if isinstance(j, ImageClassificationModelJob)],
         cmp=lambda x,y: cmp(y.id(), x.id())
         )
         ]
--- a/digits/scheduler.py
+++ b/digits/scheduler.py
@@ -5,6 +5,7 @@ import time
 import shutil
 import traceback
 import signal
+from collections import OrderedDict
 
 import gevent
 import gevent.event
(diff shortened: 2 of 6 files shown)


digits/scheduler.py
```diff
@@ -88,7 +88,7 @@ def __init__(self, gpu_list=None, verbose=False):
         gpu_list -- a comma-separated string which is a list of GPU id's
         verbose -- if True, print more errors
         """
-        self.jobs = []
+        self.jobs = dict()
```
lukeyeager: I think we're going to want an `OrderedDict` here. If we add them in order when loading from `jobs_dir` then we can avoid calls to `sort()` elsewhere.

digits/model/images/classification/views.py
```diff
@@ -475,7 +475,7 @@ def image_classification_model_top_n():
 
 def get_datasets():
     return [(j.id(), j.name()) for j in sorted(
-        [j for j in scheduler.jobs if isinstance(j, ImageClassificationDatasetJob) and (j.status.is_running() or j.status == Status.DONE)],
+        [j for j in scheduler.jobs.values() if isinstance(j, ImageClassificationDatasetJob) and (j.status.is_running() or j.status == Status.DONE)],
```
lukeyeager: I think you can still just use `for j in scheduler.jobs` throughout this PR. No need for `.values()`, right?

digits/scheduler.py
```diff
@@ -246,37 +242,37 @@ def delete_job(self, job):
...
+                        dependent_jobs.append(j.name())
+            if len(dependent_jobs)>0:
+                error_message = 'Cannot delete "%s" because %d model%s depend%s on it: %s' % (
+                        job.name(),
+                        len(dependent_jobs),
+                        ('s' if len(dependent_jobs) != 1 else ''),
+                        ('s' if len(dependent_jobs) == 1 else ''),
+                        ', '.join(['"%s"' % j for j in dependent_jobs]))
+                raise errors.DeleteError(error_message)
+            del self.jobs[job_id]
```
lukeyeager: This would be a bit safer:
```python
self.jobs.pop(job_id, None)
```

lukeyeager:
(toggled close/open to force a TravisCI build)

lukeyeager:
From the failed tests, it looks like you've changed some of the logic of when jobs are allowed to be deleted? I'm assuming you didn't mean to do that.

digits/scheduler.py
jmancewicz: pop would avoid a potential KeyError, which in this case should not happen, because of the `job = self.jobs.get(job_id, None)`.  However it's a better practice.  I'll make that change.

digits/model/images/classification/views.py
jmancewicz: The js in `j in scheduler.jobs` would be the keys (the job_ids).

digits/scheduler.py
jmancewicz: I was thinking that dict() was sorted associative container.  I was wrong.  It doesn't look like python has that.  The OrderedDict is insertion order, which should be sorted at red time and job add time.

digits/model/images/classification/views.py
lukeyeager: Oh right, duh.

jmancewicz:
I had changed some of delete_job, but not enough.  I had the `for j in scheduler.jobs` rather than `for j in scheduler.jobs.values()` I'd changed it in multiple places in DIGITS, but missed that one.  Didn't see it until I redid the changes from scratch and diffed against that branch.  That's why I was asking about a debugger.

lukeyeager:
It looks like you never actually said why this change is desirable. This speeds up the lookup time for `get_job()`, `abort_job()` and `delete_job()` in the scheduler from `O(n)` where n is the total number of jobs in the job list to `O(1)`.

It's passing the testsuite and my own smoketests and everything looks fine. Very concise - I like it! Squash those last 3 commits and merge this, please.

lukeyeager:
Squash those commits and let's merge now that [v3.1 development has started](https://github.com/NVIDIA/DIGITS/commit/374057ecbca244a0ff94cc9d2f543543a5eef894).

PR #361, opened 2015-10-13
Fix infer_many for Caffe

lukeyeager:
*Fix #255, Fix #349*

Previously, `scores = output` performed a shallow copy that was
overwritten when the next output was calculated. Now, we're using
`np.copy` to do a deep copy and avoid the error.

PR #409, opened 2015-11-09
Deep copy scores in classification example

yangj1e:
fix for classification example, similar to #361


lukeyeager:
Ah yes, I forgot to fix it there. Thanks!

PR #410, opened 2015-11-10
Torch - report error on job summary

gheinrich:
Introduce a wrapper script to catch and report errors
in a more straightforward way. The main error is now
shown on the job summary window, in big, red letters.


gheinrich:
The error now appears on the job summary. For example, on an attempt to train a model using incompatible input scale:
digits-error

lukeyeager:
This seems to be working for me. It caught cuDNN and size mismatch errors well.

I'd like to merge this before RC2 - too risky?

gheinrich:
Thanks Luke. I think it's not too risky.

PR #414, opened 2015-11-12
Allow opening of read-only LMDB in Torch

gheinrich:
Implement similar patch to https://github.com/BVLC/caffe/pull/3088

If we can't open the LMDB database, try again with no locking.

If this fails again, throw a human-readable error, for example:
2015-11-11 21:29:33 [INFO ] opening LMDB database: /fast-scratch/gheinrich/ws/digits/digits/jobs/20151111-210842-a4ec/train_db
/fast-scratch/gheinrich/ws/torch/install/bin/luajit: ...nrich/ws/torch/install/share/lua/5.1/threads/threads.lua:264:
[thread 4 callback] /fast-scratch/gheinrich/ws/digits/tools/torch/data.lua:117: Failed to open LMDB database: Permission denied

### DIFF
--- a/tools/torch/data.lua
+++ b/tools/torch/data.lua
@@ -112,7 +112,26 @@ function LMDBSource:new(lighningmdb, path)
     self.e = self.lightningmdb.env_create()
     local LMDB_MAP_SIZE = 1099511627776 -- 1 TB
     self.e:set_mapsize(LMDB_MAP_SIZE)
-    self.e:open(path, self.lightningmdb.MDB_RDONLY + self.lightningmdb.MDB_NOTLS,0664)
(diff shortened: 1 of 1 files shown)


gheinrich:
cc @jmancewicz and @lukeyeager, if you don't mind having a look. Thanks!

jmancewicz:
Looks good to me.  I'm just curious why you aren't opting to NOLOCK from the start?

gheinrich:
I am not using NOLOCK from the start because if the lock file is writable then taking the lock provides the additional safety net that no other process can write to the database (assuming the other process isn't using NOLOCK).

PR #415, opened 2015-11-12
Update list of requirements to force Flask-SocketIO==0.6.0

lukeyeager:
https://groups.google.com/d/msg/digits-users/_6xGrd_FmHA/YYrG4AVFAQAJ
```
$ ./digits-server
Error: class uri 'socketio.sgunicorn.GeventSocketIOWorker' invalid or not found: 

[Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/gunicorn/util.py", line 132, in load_class
    mod = __import__('.'.join(components))
ImportError: No module named socketio.sgunicorn
]
```

Somebody upgraded their pip package on PyPI and now the standard DIGITS install instructions don't work.

I suspect the culprit is Flask-SocketIO v0.6 -> v1.0.

The fix in this pull request worked for that specific user and for me, but I haven't had the time to debug this properly yet.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,9 +7,9 @@ requests>=2.2.1
 gevent>=1.0
 Flask>=0.10.1
 Flask-WTF>=0.11
+Flask-SocketIO==0.6.0
 gunicorn==17.5
 setuptools>=3.3
 lmdb>=0.87
 h5py>=2.2.1
 pydot2
-Flask-SocketIO


lukeyeager:
@miguelgrinberg can you comment on whether you'd expect upgrading Flask-SocketIO from 0.6.0 to 1.0 would cause this error?

[Here's how I'm calling gunicorn](https://github.com/NVIDIA/DIGITS/blob/v3.0.0-rc.1/digits-server#L23-L26):
```
$PYTHON_EXE $GUNICORN_EXE \
    --config gunicorn_config.py \
    digits.webapp:app \
    $@
```
[And here's my default gunicorn config](https://github.com/NVIDIA/DIGITS/blob/v3.0.0-rc.1/gunicorn_config.py#L3-L5):
```python
worker_class = 'socketio.sgunicorn.GeventSocketIOWorker'
bind = '0.0.0.0:34448' # DIGIT
loglevel = 'debug'
```

miguelgrinberg:
@lukeyeager the gunicorn configuration is different for 1.0. See the documentation, the gevent-socketio worker isn't used anymore.

miguelgrinberg:
@lukeyeager more than happy to help with the transition when the time comes, so reach out if you need advice!

SevenGong:
@lukeyeager  I got the same problem as yours. Is this problem still not solved by now?  Looking forward to your rely.

lukeyeager:
@SevenGong if you follow these instructions to install the correct version of packages you shouldn't get the error I mentioned in the OP:
https://github.com/NVIDIA/DIGITS/blob/digits-4.0/docs/BuildDigits.md#python-packages

PR #420, opened 2015-11-19
Trim whitespaces off of labels file

gheinrich:
The labels.txt file used to map class names to class IDs might
contain line endings that could confuse the parser during inference.


lukeyeager:
This is a fix for Allison's bug, right?

Doesn't cause any problems for me. Seems safe enough.

gheinrich:
Yes this fixes a problem with Windows line endings that was causing identical symptoms to those seen and reported by Allison.

PR #421, opened 2015-11-19
Fix the upgrade path for TrainTask.use_mean

lukeyeager:
*Fixes a bug from #321*

When upgrading `TrainTasks` from before #321, we need to update the `use_mean` field accordingly.

To reproduce the bug:

1. `git checkout v2.2.1` and start up the server
2. Create a new classification model job
3. Verify that classifying an image works correctly
4. Shut down the server, `git checkout v3.0.0-rc.1` and start up the server again
5. Try classifying an image with the same model


lukeyeager:
@jmancewicz can you review and merge this, please?

gheinrich:
This looks OK to me.

jmancewicz:
Good catch.

PR #422, opened 2015-11-20
Update docs for v3.0

lukeyeager:
Chipping away at #401. I've updated the list there with more TODOs.

* Move build instructions out of README
* Add instructions for installing debian packages on ubuntu (doesn't work yet since packages aren't public yet)

There are probably some typos/errors in here. I'll merge it tomorrow if I don't get any feedback, but I expect this still needs a little work (in addition to the other docs TODOs).

Issue #401, opened 2015-11-05
Update docs for v3.0


lukeyeager:
* [x] Update getting started guide
  * [x] Updated images
  * [ ] Images/GIFs of new features?
* [x] Recommend caffe v0.14 and cuDNN v4 for manual installation
* [x] Recommend debian package installation
  * [x] Note that you have to shut down the digits server before uninstalling/reinstalling a driver
  * [x] Note that images and "generic" databases need to be readable by user www-data when using the .deb packages
* [x] Torch instructions (@gheinrich)
  * [x] Remind people they need some of the .deb packages that Caffe requires (like hdf5)
  * [x] Split up build and usage instructions into different files
* [x] Correct `digits-server` port in README (#392)
* [x] HOWTO for python layers (@jmancewicz) (#329)
* [x] HOWTO for generic models - autoencoder (@gheinrich)

lukeyeager:
* HOWTO for python layers (@jmancewicz)
* Note that you have to shut down the digits server before uninstalling/reinstalling a driver

lukeyeager:
Added to the list above.

lukeyeager:
@jmancewicz we're pretty much done with this except for the lack of a Python layer HOWTO. Do you have time to do that this week? Do you need to offload that to me?

lukeyeager:
Python example added with #447.

PR #423, opened 2015-11-20
Update docs for v3.0

lukeyeager:
*Redo of #422*

Chipping away at #401. I've updated the list there with more TODOs.

* Move build instructions out of README
* Add instructions for installing debian packages on ubuntu (doesn't work yet since packages aren't public yet)

There are probably some typos/errors in here. I'll merge it tomorrow if I don't get any feedback, but I expect this still needs a little work (in addition to the other docs TODOs).


docs/BuildCaffe.md
```diff
@@ -0,0 +1,80 @@
...
+    python-dev python-pip python-numpy gfortran
+% sudo apt-get install --no-install-recommends libboost-all-dev
+```
+
+### CUDA toolkit
+
+To install the CUDA toolkit, first get access to the required repositories by following [these instructions](UbuntuInstall.md#repository-access).
+Then install the toolkit with this command:
+```sh
+% sudo apt-get install cuda-toolkit-7-0
```
gheinrich: I am curious to know why we are not recommending CUDA 7.5?

docs/GettingStarted.md
```diff
@@ -11,21 +11,23 @@ Table of Contents
...
-If you are installing from source, check out the [README](../README.md#installation).
+Follow [these instructions](BuildDigits.md) to build from source.
 
 ## Starting the server
 
 If you are using the web installer use the `runme.sh` script:
-
-    % cd $HOME/digits-2.0
-    % ./runme.sh
+```sh
```
gheinrich: should this section be deleted and replaced with something that says that under the .deb package method the server is automatically started on local port x under user www-data, with digits log going into /xxx/yyy/digits.log

docs/GettingStarted.md
```diff
@@ -11,21 +11,23 @@ Table of Contents
 
 ## Installation
 
-If you are using the web installer, check out this [installation page](WebInstall.md).
+Follow [these instructions](UbuntuInstall.md) to install via debian packages.
```
gheinrich: some people might argue that it's technically a `deb` package not a `debian` package :-)

docs/GettingStarted.md
lukeyeager: Definitely. I was going to wait to do it in another PR, but I guess I might as well do it now ...

docs/GettingStarted.md
lukeyeager: Good point. I'll check and make sure that I'm consistent with using "deb package" instead of "debian package".

docs/UbuntuInstall.md
```diff
@@ -0,0 +1,44 @@
+# Ubuntu Installation
+
+Debian packages are provided for easy installation on Ubuntu 14.04.
```
gheinrich: another one there but I'm sure you would have found it :-)

lukeyeager:
Thanks for the feedback, Greg. I've addressed those suggestions and now I'm working on new screenshots for the GSG...

PR #424, opened 2015-11-20
Autoencoder tutorial

gheinrich:
*Chipping away at #401*


lukeyeager:
Great stuff, thanks!

This doesn't have to be Torch-specific, can you add instructions for doing it with Caffe as well?

It works for me when I use this prototxt (with the selected lines removed):
https://github.com/BVLC/caffe/blob/df21b6a1c1/examples/mnist/mnist_autoencoder.prototxt#L18-L34

gheinrich:
With the Caffe model the result is slightly less impressive. With this normalized input image:
caffe-norm-image
I get this reconstructed image:
caffe-reconsructed
I am not sure it is learning much, the loss functions look strange:
caffe-loss
The Caffe model is using only 30 neurons at the output of the encoder, which is less than I have been using in Torch (50 neurons) but even with 20 neurons in Torch the reconstructed image looked more similar to the input image. Maybe it's just the unflattening of data in Caffe that makes the image look different?

lukeyeager:
Yeah, I see weird loss graphs too. The loss actually goes negative (you can see it if you hack the javascript a bit).

Ok maybe you're right - let's not suggest Caffe for this example.

examples/autoencoder/README.md
```diff
@@ -0,0 +1,115 @@
...
+![Create generic dataset form](create-generic-dataset-form.png)
+
+## Model creation
+
+Now that you have a generic dataset to train on, you may create a generic model by clicking on `New Model\Images\Other` on the main page:
+
+![Create generic model](create-model.png)
+
+Select the dataset you just created and under the `Custom network` tab, select `Torch`. There you can paste the following network definition:
+```
```
lukeyeager: You can use ```lua for syntax highlighting here and elsewhere.

https://help.github.com/articles/github-flavored-markdown/#syntax-highlighting

lukeyeager:
Great write-up! Let's add the syntax highlighting and merge.

gheinrich:
Thanks I have updated the doc with Lua syntax highlighting.

lukeyeager:
Great, thanks!

lukeyeager:
Merged into `master` and `digits-3.0`.

PR #425, opened 2015-11-20
remove job_management route

jmancewicz:


lukeyeager:
This was accidentally added with #240. Thanks for the fix, Joe!

lukeyeager:
Merged into `master` and `digits-3.0`.

PR #426, opened 2015-11-20
Autoencoder tutorial

gheinrich:


lukeyeager:
Thanks for re-filing #424 against `digits-3.0`. But I'd rather leave that one open and manually merge the merged commits over to `digits-3.0` like I did for #425.

PR #440, opened 2015-11-29
Update Torch Alexnet reference model

gheinrich:
- Move to "1-tower" architecture,
- Enable CuDNN fastest mode,
- Remove unnecessary input grad computations on first layer
- Use optimal backend for SoftLogMax layer

This model trains more quickly than the previous model:
14m3s -> 10m35s (two epochs on 256x256 upscaled CIFAR10).

### DIFF
--- a/digits/standard-networks/torch/ImageNet-Training/alexnet.lua
+++ b/digits/standard-networks/torch/ImageNet-Training/alexnet.lua
@@ -1,11 +1,9 @@
--- source: https://github.com/soumith/imagenet-multiGPU.torch/blob/master/models/alexnet_cudnn.lua
-
-require 'nn'
 if pcall(function() require('cudnn') end) then
    print('Using CuDNN backend')
    backend = cudnn
    convLayer = cudnn.SpatialConvolution
    convLayerName = 'cudnn.SpatialConvolution'
+   cudnn.fastest = true
 else
    print('Failed to load cudnn backend (is libcudnn.so in your library path?)')
    if pcall(function() require('cunn') end) then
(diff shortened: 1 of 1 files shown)


lukeyeager:
Cool! Is this network more or less "standard" than the previous one? Does the standard Caffe network follow the "1-tower" or "2-tower" architecture? I'd like to keep the network as close as possible to what others are using in standard literature, rather than providing our own custom versions which may or may not be better (faster, more compact, more stable, etc.) than the original.

gheinrich:
The 2-tower architecture (see [here](https://github.com/gheinrich/DIGITS/commit/25e680fb4f83e02cd7bb6eeb4148d25960ba7bcd#diff-8d3c6576ee7035f1a3b28f25690b64feL28)) was a useful representation of the network, introduced in the Krizhevsky paper. This makes it easy to understand how to split training over two GPUs. This representation doesn't actually change the parameters of the network (number of filters, etc.). Other than that this new model is using the configuration from the Krizhevsky paper (which is the one used in the reference model for Caffe).

As a side note, apparently the `group` parameter used [there](https://github.com/NVIDIA/DIGITS/blob/master/digits/standard-networks/caffe/alexnet.prototxt#L105) is non standard. As a result some layers of Alexnet are actually using half as many filters as you’d expect. Background info: https://github.com/BVLC/caffe/issues/778

digits/standard-networks/torch/ImageNet-Training/alexnet.lua
```diff
@@ -1,11 +1,9 @@
--- source: https://github.com/soumith/imagenet-multiGPU.torch/blob/master/models/alexnet_cudnn.lua
-
-require 'nn'
```
lukeyeager: `nn` is still used - why is it not required anymore?

digits/standard-networks/torch/ImageNet-Training/alexnet.lua
```diff
@@ -80,11 +67,10 @@ return function(params)
         assert(params.inputShape[2]==256 and params.inputShape[3]==256, 'Network expects 256x256 images')
     end
     return {
-        model = createModel(1, channels),
+        model = createModel(params.ngpus, channels),
```
lukeyeager: Multi-GPU is still not supported, right? But I guess it's blocked in DIGITS before you ever get this far so this is ok?

gheinrich:
Multi-GPU is still not supported indeed (this change is a small step in the direction of getting it to work through).

PR #329, opened 2015-09-23
Add widget to copy python file from client or server to job dir

jmancewicz:
Copies the python file to the job directory from the server of the client.

PR #444, opened 2015-12-01
Make python layers work for inference

lukeyeager:
This addresses a bug from https://github.com/NVIDIA/DIGITS/pull/329. Without this fix, you can train a model with a PythonLayer, but you can't test it.
```
SystemError
NULL result without error in PyObject_Call

    rv = self.dispatch_request()
  File "/home/lyeager/digits/digits/model/images/classification/views.py", line 291, in image_classification_model_classify_one
    predictions, visualizations = job.train_task().infer_one(image, snapshot_epoch=epoch, layers=layers)
  File "/home/lyeager/digits/digits/model/tasks/caffe_train.py", line 1017, in infer_one
    layers=layers,
  File "/home/lyeager/digits/digits/model/tasks/caffe_train.py", line 1042, in classify_one
    net = self.get_net(snapshot_epoch)
  File "/home/lyeager/digits/digits/model/tasks/caffe_train.py", line 1368, in get_net
    caffe.TEST)
SystemError: NULL result without error in PyObject_Call
```
The message in the console output is much more informative:
```
ImportError: No module named python-layer
```
This requires two changes:

1. Add `job_dir` to the path before loading a new Caffe model so that the python file can be found

2. Force a reload of any modules that were loaded by a previous model (see comments)

### DIFF
--- a/digits/model/images/classification/test_views.py
+++ b/digits/model/images/classification/test_views.py
@@ -952,7 +952,7 @@ layer {
     bottom: "output"
     top: "py_test"
     python_param {
-        module: "py_test"
+        module: "digits_python_layers"
         layer: "PythonLayer"
     }
 }
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -6,6 +6,7 @@ import time
 import math
 import subprocess
 import operator
+import sys
 
 import numpy as np
 import scipy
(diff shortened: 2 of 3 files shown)


lukeyeager:
Don't merge this - I don't want to figure out the implications of force reloading `digits` or `caffe` if the python layer file is named `digits.py` or `caffe.py`. I'm investigating another solution...

lukeyeager:
I think this is a safer solution. I'm hard-coding the filename to `digits_python_layers.py` - thus requiring that any models using python layers uploaded through DIGITS to specify their network like so:
```
layer {
    name: "MyPythonLayer"
    type: "Python"
    bottom: "blobA"
    top: "blobB"
    python_param {
        module: "digits_python_layers"
        layer: "MyPythonLayer"
    }
}
```
@jmancewicz what do you think? This is the simplest solution I can think of that gets around the issue of somebody uploading a filename with the name of a standard module.

jmancewicz:
Forcing the name to be `digits_python_layers.py` handles the reload of `caffe` or `digits` issue.  Adding the dashless job number does as well but is uglier.  Can you envision a time where multiple python layer files need to be uploaded?

lukeyeager:
> Can you envision a time where multiple python layer files need to be uploaded?

It might be convenient, but not needed. You can put the definition for multiple Python layers into a single Python file.

PR #447, opened 2015-12-01
Add a HOWTO example for Python layers

lukeyeager:
*Chipping away at #401*

NOTE: you need the fix from https://github.com/NVIDIA/DIGITS/pull/444 before you can test this.


lukeyeager:
@jmancewicz, @gheinrich - can one of you guys try this out? It'll be a good way to test #444 as well.

jmancewicz:
Worked very well.  I like the extra credit section.

gheinrich:
Worked for me too (a note to highlight the fact that Python layers must be enabled when compiling Caffe could be useful). Thanks.

lukeyeager:
> (a note to highlight the fact that Python layers must be enabled when compiling Caffe could be useful)

Done. I also addressed Joe's whitespace suggestion. Thanks for taking a look!

PR #454, opened 2015-12-04
Export number of classes to Torch models

gheinrich:
Allow networks to select desired number of outputs depending on
number of classes in (classification) dataset.

Stop automatically adjusting number of outputs in Lua wrappers.
This is less confusing and this will make it easier to enable
multi-GPU training with Torch in DIGITS.

### DIFF
--- a/digits/model/images/classification/test_views.py
+++ b/digits/model/images/classification/test_views.py
@@ -68,6 +68,8 @@ layer {
     TORCH_NETWORK = \
 """
 return function(p)
+    -- adjust to number of classes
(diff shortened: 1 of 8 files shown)


lukeyeager:
Great idea! Works for me. I wonder if we could do something similar with Caffe? It'd have to be something that gets replaced before the protobuf is parsed, since you can't have variables in protobuf.

There's a merge conflict - let's address that and then I think this is good to to.

gheinrich:
Thanks Luke, I have rebased my patch on the tip of the master branch.

PR #456, opened 2015-12-07
Set linear weights to zero in automatic tests

gheinrich:
By default Caffe zero-initialises weights and biases for linear
layers (http://caffe.berkeleyvision.org/tutorial/layers.html).
Torch on the other hand initializes linear layer parameters by
picking from a uniform distribution.
For the problem that we are trying to solve in the automatic
tests it is quicker to train from zero-initialised weights
as we only really need to learn the sign of those parameters.


lukeyeager:
Hopefully this finally knocks out #346 for good!

Issue #407, opened 2015-11-07
Digits UI with Authentication


gautamsingaraju:
I run Digits on an EC2 instance. When I set long running jobs, I would like to have some mechanism of authentication to the UI. It could be as simple as username and password being read from the disk to integration with LDAP.

This is a feature request for some basic WebUI authentication.

lukeyeager:
We're planning to do #201 first. Will that be enough for a start?

Once we add authentication to the server, there is a certain amount of security that would be expected. Currently, we're treating security as a non-issue. So this probably won't happen soon.

singarajus:
Thanks, makes sense. I have "constrained" the access to a few IP addresses using firewalls.

lukeyeager:
@singarajus I've opened a PR for basic login functionality at #463.

singarajus:
Thanks, Luke!

lucbettaieb:
Bump!

My campus' network forward everything, so if I run a DIGITS server on my computer, theoretically anybody could access it if they had the hostname.  Any way I can make the server password protected?

lucbettaieb:
:+1: This would be great

elezar:
Some form of authentication/login would be really useful.

croninstephen:
I'd love to see that feature too. My machine is on a university network and I don't want people to jsut be poking around.

PR #463, opened 2015-12-10
Add basic login functionality

### LINKED ISSUE #201: Add concept of user
As we add features like results browsing and move DIGITS towards more of a workgroup application, we need some way to separate the information in a common jobs directory in to usable groupings.  The first such logical grouping is that of "by user."  A typical user of DIGITS isn't aware of what other users are doing or how they are experimenting with neural networks.  By adding the concept of a user, we can automatically add tags that allow filtering and enable a DIGITS user to see just their own results.  By changing the filter mechanism, they would be able to see all user's results.

Initially. there is NO authentication with this mechanism.  The DIGITS user will be presented with an option to specify a user name and that will be saved as a cookie in the browser.

Tasks:
* each presentation page (home, create dataset, create model) needs to be able to query for user cookie if not available, present an alert asking for user name.  Having NO user name is not allowed but we should allow anonymous as a valid entry
* each presentation page will present on the top bar the current user name and a link to log out.
* when creating a data set or model, the entity will be tagged with the current user name
* when browsing results, default to browsing the results of the current user

lukeyeager:
*Close #201*

* Login does not require password (not solving #407)
* Adds login info to the navbar
* Requires login before creating a new Job
* Only the original creator can abort/edit/delete a Job

after-login

## TODO after merge
* Sort/filter by username in model browser at `/models`
* Show only the current user's jobs on the homepage


jmancewicz:
What are you doing with existing jobs?

lukeyeager:
> What are you doing with existing jobs?

They are marked with `job.username=None`. Any logged in user can edit/delete them.

jmancewicz:
Perfect

lukeyeager:
*Need to fix the test suite before merging...*

iwalkdaline:
How do I clone the repository to get this functionality or enable it?  I've already cloned the master expecting it would be there.

lukeyeager:
@iwalkdaline this is an unmerged pull request - it hasn't been merged into `master` yet. You should probably just wait until we merge it.

If you really want to try it out for some reason, the git commands would be something like this:
```
git remote add lukeyeager https://github.com/lukeyeager/DIGITS.git
git remote update
git checkout lukeyeager/login
```

iwalkdaline:
We really want some form of authentication on our server.  I was pursuing a
SSL approach with a self-signed certificate but that required slightly more
work than I anticipated.  As this is implemented, I would proceed with it
unless there's some corner case I should be aware of.

___________________________________________
Ivan Walker
(601)918-0435
Graduate Student
GVU BrainLab
Georgia Institute of Technology
Atlanta, GA  30032  USA

lukeyeager:
Update on this - I had to remove all the Flask-Autodoc stuff in order to add custom route decorators.

I suspect nobody was using the auto-generated documentation anyway. If so, we'll find a way to add it back in somehow.

lukeyeager:
@iwalkdaline I'm going to merge this as-is. We're not going to address authentication in DIGITS for now, but the scaffolding that I'm putting in place with this PR should make it easier for you to modify DIGITS to add authentication on your own.

Please feel free to upvote (:+1:) #407 if you like.

iwalkdaline:
I see.  Also, I'm trying to add a password field and I think this will
address my needs but not sure of the data flow here.  Seems I might need to
1) add it webapp.py (in username decorator or a new decorator)
2) add it to auth.py to verify the password
3) modify views.py

I've tried so far but get the error that password isn't defined globally.

- Ivan
​

PR #471, opened 2015-12-11
Speedup generic tests

lukeyeager:
Since the generic test suite all uses the same LMDBs, we can just create them once rather than re-creating them for each test class.

Command:
```sh
./digits-test digits/*/images/generic/ -s -v
```
Before:
```
153 tests run in 223.1 seconds.
3 skipped (150 tests passed)
```
After:
```
153 tests run in 156.3 seconds.
3 skipped (150 tests passed)
```

### DIFF
--- a/digits/dataset/images/generic/test_views.py
+++ b/digits/dataset/images/generic/test_views.py
@@ -61,9 +61,10 @@ class BaseViewsTestWithImageset(BaseViewsTest):
     @classmethod
     def setUpClass(cls):
         super(BaseViewsTestWithImageset, cls).setUpClass()
-        cls.imageset_folder = tempfile.mkdtemp()
-        # create imageset
-        cls.test_image = create_lmdbs(cls.imageset_folder)
+        if not hasattr(BaseViewsTestWithImageset, 'imageset_folder'):
+            # Create folder and LMDBs for all test classes
+            BaseViewsTestWithImageset.imageset_folder = tempfile.mkdtemp()
+            BaseViewsTestWithImageset.test_image = create_lmdbs(BaseViewsTestWithImageset.imageset_folder)
         cls.created_datasets = []
 
     @classmethod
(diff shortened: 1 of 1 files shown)


gheinrich:
Tested on my machine:
before: 6m0s,
after: 3m0s

digits/dataset/images/generic/test_lmdb_creator.py
```diff
@@ -26,9 +26,9 @@
 import caffe_pb2
 
 
-IMAGE_SIZE  = 10
+IMAGE_SIZE  = 8
 TRAIN_IMAGE_COUNT = 100
-VAL_IMAGE_COUNT = 20
+VAL_IMAGE_COUNT = 10
```
gheinrich: on my machine the two changes from this file don't seem to produce a measurable speed up. Do we need them?

digits/dataset/images/generic/test_views.py
```diff
@@ -61,18 +61,16 @@ class BaseViewsTestWithImageset(BaseViewsTest):
...
+            BaseViewsTestWithImageset.imageset_folder = tempfile.mkdtemp()
+            BaseViewsTestWithImageset.test_image = create_lmdbs(BaseViewsTestWithImageset.imageset_folder)
         cls.created_datasets = []
-
     @classmethod
     def tearDownClass(cls):
         # delete any created datasets
         for job_id in cls.created_datasets:
             cls.delete_dataset(job_id)
-        # delete imageset
```
gheinrich: This isn't new with this commit but these tests are leaking temporary files. The generic model tests alone leak 113 temporary folders. With this commit: 114.

lukeyeager: Yeah, I'm intentionally leaking a temporary directory in this PR. They get cleaned up on reboot and I can't imagine many people use the test suite on anything other than a development machine.

I'm surprised that so many temporary folders are created, though. Do you know where that's coming from? I'm going to go ahead and merge this PR, but we can continue the discussion here and perhaps address it in a separate PR if necessary.

gheinrich:
I used to know where the leak was coming from but didn't write it down and I have forgotten since then... sorry! When I get some time I will try again.

Incidentally, today I realized I had accidentally left the classification tests run in a loop for about a  week and this had eaten up all available disk space in the `/tmp` partition of my work station...

PR #473, opened 2015-12-15
Display mean and explore btn when task is done

### LINKED ISSUE #404: Need to refresh page after creating an LMDB to see Explore DB and mean image
While creating an LMDB, the page does not display the 'Explore the DB' button or mean image. As a user, it would make sense that they show up when the job successfully finishes.

lukeyeager:
*Close #404*

Shouldn't require a page refresh to see these elements.


gheinrich:
Looks good and works for me.

PR #476, opened 2015-12-16
Upgrade to Flask-SocketIO v1.0

lukeyeager:
In #415 I locked the version to v0.6.0. Now I'm upgrading to (and requiring) v1.0.

I followed the upgrade guideline in [the Flask-SocketIO documentation](http://flask-socketio.readthedocs.org/en/stable/#upgrading-to-flask-socketio-1-x-from-older-releases). Seems to be working just fine.


lukeyeager:
Don't merge this yet. I haven't decided what to do about this comment yet:

> If you wish to continue using gevent, then uninstall gevent-socketio from your virtual environment, as this package is not used anymore and may collide with its replacement, python-socketio.

I have the [python-socketio](http://packages.ubuntu.com/trusty/python-socketio) deb package installed on my machine (because it's required for the `digits` deb package), which provides provides @abourget's PyPI package called [gevent-socketio](https://pypi.python.org/pypi/gevent-socketio) instead of @miguelgrinberg's PyPI package called [python-socketio](https://pypi.python.org/pypi/python-socketio). [Yuck!]

It looks like we'd need to force the next `digits` deb package to conflict with the `python-socketio` package and we'd have to create a new package for the new socketio package (called `python-python-socketio`?). Plus, anyone who has DIGITS installed via debs and also wants to try out the `master` branch on the same machine might run into some painful dependency issues.

miguelgrinberg:
@lukeyeager so you don't use a virtual environment? A virtualenv would give you complete control of your dependencies. Even if the gevent-socketio is installed as a deb package, you can have python-socketio in your virtualenv w/o any collisions.

lukeyeager:
> so you don't use a virtual environment?

I certainly do use a virtual environment when I'm developing. But our [install process](https://github.com/NVIDIA/DIGITS/blob/digits-3.0/docs/UbuntuInstall.md#digits) for DIGITS is simply `apt-get install digits` (once you've set up access to the correct repositories). That one command installs all the necessary packages to `/usr/lib/python2.7/dist-packages`.

miguelgrinberg:
@lukeyeager Okay, I understand. Unfortunately python isn't very good at dealing with duplicate package names, I haven't specifically tested an upgrade situation in Ubuntu, it may be that the new package overrides the old, so everything would work.

lukeyeager:
> I haven't specifically tested an upgrade situation in Ubuntu, but it may be that the new package overrides the old, so everything would work.

@miguelgrinberg Sure, I could set up the new package to properly handle the transition. I'm just trying to decide if it's worth it. How much faster is v1.0 vs v0.6.0 (with socket.io.js v1.3.7 vs. v0.9.6)? Do you have any performance numbers for comparison?

miguelgrinberg:
@lukeyeager the reason for all these changes was not performance. I know eventlet is slightly faster than gevent for WebSockets. If you will continue using gevent and gevent-websocket then I don't think you'll have any significant performance improvements.

The reason all these changes were made was to unblock the clients, which were stuck with an ancient version of the Socket.IO client and protocol, known to have problems with memory usage. The Socket.IO 1.x line has a much thought out design, with the separation between Engine.IO for the low-level transport work and Socket.IO for the high-level messaging, namespaces, rooms, etc.

Another reason that you may want to consider as an upgrade factor is the support for multiple servers, which is coming soon to Flask-SocketIO. Not sure if you need this, but if you want to scale out your application to support large number of clients, with this new version you will be able to launch several instances of the server with a load balancer in front and a Redis or RabbitMQ queue in the backend. No changes to the application will be required to benefit from this, other than changing how the server object is created.

lukeyeager:
Wow, thanks for the prompt response and the detailed info @miguelgrinberg! And, if I haven't thanked you already, thanks for the excellent [repos](https://github.com/search?l=Python&q=user%3Amiguelgrinberg&type=Repositories), [documentation](http://flask-socketio.readthedocs.org/en/latest/) and [blog posts](http://blog.miguelgrinberg.com)!

I think we'll just stick with the older version for now because I don't see a real need for any of the new features yet and because I don't want to deal with the headache of creating new packages for the next release and dealing with dependency conflicts in the meantime. But I'll keep this PR around, since we will likely end up upgrading eventually.

miguelgrinberg:
Sure, that seems reasonable, since nothing is obviously broken for you.

As a side note, only now I realize that you also own the deb package that installs gevent-socketio. You could create a new version of this package that uses a different name, or actually one that installs no files. Then there would be no collision. Just an idea. :)

PR #481, opened 2015-12-18
Disable colormap for "Top N" feature

lukeyeager:
#### Before

topn-jet

---
#### After

topn-white


gheinrich:
Looks good to me and your patch is working for me (tested on color and gray scale images).

PR #484, opened 2015-12-23
Fix init of blue channel in unencoded LMDB

gheinrich:
Commit f710c5f failed to properly initialize the blue channel
when loading data from unencoded LMDB. Instead the blue channel
was initialized with the contents of the red channel. This was
ironically enough to pass the classification tests since the
'blue' class was effectiveley the only class with all zeroes.


gheinrich:
cherry-pick of 8da4aef

PR #485, opened 2015-12-23
Support table of outputs in Torch models

gheinrich:
Some model layers in Torch output tables of Tensors. This commit adds support for these layers. This also enables ND tensors in the output.


lukeyeager:
Verified by testing #486.

PR #488, opened 2015-12-31
Fixes model usage in validation code.

Liuftvafas:
Fixed model usage in validation.
Removed duplicate code in logging validation progress.
Fixed whitespaces.

### DIFF
--- a/tools/torch/data.lua
+++ b/tools/torch/data.lua
@@ -381,11 +381,14 @@ function DBSource:lmdb_getSample(shuffle, idx)
         y = image.decompress(x,msg.channels,'byte'):float()
     else
         x = x:narrow(1,1,total):view(msg.channels,msg.height,msg.width):float() -- using narrow() returning the reference to x tensor with the size exactly equal to total image byte size, so that view() works fine without issues
-        y = x -- make y see x's storage
         if self.ImageChannels == 3 then
             -- unencoded color images are stored in BGR order => we need to swap blue and red channels (BGR->RGB)
+            y = torch.FloatTensor(msg.channels,msg.height,msg.width)
             y[1] = x[3]
+            y[2] = x[2]
             y[3] = x[1]
+        else
+            y = x
         end
     end
 


gheinrich:
@Liuftvafas thank you very much for your contributions! Would you mind sending a signed copy of the Contributor License Agreement as described [there](https://github.com/NVIDIA/DIGITS/blob/master/CONTRIBUTING.md#pull-requests)?

You're right that the usage of a function parameter `nn_model` and global variable `model` to designate the same object in `Validation()` is definitely inconsistent. Did you see an actual issue there, or as you suggesting this change merely to improve readability/robustness? If anything, I'd rather stick to using `nn_model` in this function as I think this would go in the direction of a cleaner separation of namespaces so that eventually we can move the code to another module, if we wish to.

Along the same lines, perhaps `network`, `validationConfusion`, `validationBatchSize` and `valSize` could be made function parameters. If I haven't missed any other globals, that would make the function `Validation()` independent of global variables, which I think would be a good thing. Then the pull request and commit could be renamed to something like "Refrain from using globals in Validation()".

Do you think you could submit the whitespace changes as a separate commit in order to make it easier to see where the actual change lies?

Also, do you think you could submit the in-place multiplication change in `test.lua` as a separate pull request (as this isn't well captured by the title of this pull request).

Liuftvafas:
OK, I will update the code to use more function parameters both in Validation and Train functions to get rid of global variables.
I'm not sure how to remove the in-place multiplication from this pull request (I've already submitted it as separate pull request). Should I close this pull request and restart local/global variables refactoring from scratch in the new one?

gheinrich:
Hey @Liuftvafas you can create a local branch of yours and cherry-pick the patches you want to submit in this Pull Request then force push to your `Liuftvafas:master` branch. You may do something like:

```
git checkout origin/master
git checkout -b dev_branch
git cherry-pick <SHA1>
git push --force https://github.com/Liuftvafas/DIGITS.git HEAD:refs/heads/master 
```

I am re-opening this Pull Request so we keep the history of comments.

PR #491, opened 2016-01-04
Move image encoding out of main create_db thread

### LINKED ISSUE #254: Problem with CPU utilization while creating datasets
is there any difference in the CPU usage in digits v2.2 ?

Cuz while doing Data Pre Processing on v2.1 it did use all my 30 CPU cores and in v2.2 it uses 1 CPU core only.

so it reflects the time remaining.
 
Thank you very Much.
yanky

lukeyeager:
*Fixes #254 (coming from #245)*

Improves CPU utilization when creating encoded LMDB datasets.

Tools used to debug: [cProfile](https://docs.python.org/2/library/profile.html).


lukeyeager:
[Confirmed](https://github.com/NVIDIA/DIGITS/issues/254#issuecomment-169045231) by @michaelholm-ce.

PR #492, opened 2016-01-05
Update copyright headers for 2016

lukeyeager:
Happy new year, everyone!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-# Copyright (c) 2015, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2015-2016, NVIDIA CORPORATION.  All rights reserved.
 
 sudo: required
 dist: precise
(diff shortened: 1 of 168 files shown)


lukeyeager:
For my sake for next year, here is how I generated these changes:
```python
#!/usr/bin/env python

import re
import sys
import time

CURRENT_YEAR = str(time.localtime().tm_year)

def process_file(filename):
    changed = False
    result = ''
    with open(filename) as infile:
        for line in infile:
            if 'NVIDIA CORPORATION' in line:
                match = re.match('(.*Copyright \(c\) )(\d{4})(-\d{4})?(, NVIDIA CORPORATION.*$)', line)
                if match:
                    before = match.group(1)
                    year1 = match.group(2)
                    year2 = match.group(3) or ''
                    after = match.group(4)
                    line = ''.join([before, year1, '-', str(CURRENT_YEAR), after, '\n'])
                    changed = True
            result += line

    if changed:
        with open(filename, 'w') as outfile:
            outfile.write(result)
    else:
        print 'File not changed -', filename


for filename in sys.argv[1:]:
    process_file(filename)
```
```sh
git grep 'NVIDIA CORPORATION' | cut -d: -f1 | xargs update_copyright_headers.py
```

PR #495, opened 2016-01-06
Refrain from using globals in Validation().

Liuftvafas:


Liuftvafas:
I don't see why continuous integration could fail on my changes here. It seems that fail occurs on installing torch dependencies but not my code.

lukeyeager:
> I don't see why continuous integration could fail on my changes here. It seems that fail occurs on installing torch dependencies but not my code.

You're right - it doesn't look like a problem with your code. I've restarted the TravisCI build, let's see how it goes ...

lukeyeager:
Sure enough, it passed. See https://github.com/torch/ezinstall/issues/80.

tools/torch/main.lua
```diff
@@ -485,11 +485,13 @@ if opt.optimState ~= '' then
...
 
 local function updateConfusion(y,yt)
-    if confusion ~= nil then
-        confusion:batchAdd(y,yt)
+    if trainConfusion ~= nil then
+        trainConfusion:batchAdd(y,yt)
     end
 end
 
+local label_function = network.labelHook or function (input, dblabel) return dblabel end
```
gheinrich: sorry for being picky but the naming convention is slightly inconsistent with the other changes (where camel case is being used for globals)

tools/torch/main.lua
```diff
@@ -181,8 +181,8 @@ if opt.subtractMean ~= 'none' then
 end
 
 local classes
-local confusion
-local validation_confusion
+local trainConfusion
+local validConfusion
```
gheinrich: question: don't you think the use of the contraction `valid` could be slightly ambiguous (given that `valid` is both an adjective and the contraction of `validation`?). `validationConfusion` sounds a bit wordy however I think this is unambiguous.

gheinrich: Or perhaps something like `valConfusion` would be better?

tools/torch/main.lua
Liuftvafas: I did not get  which is the consistent naming convention in this file because it was mixed inside the same Validation() function and in global variables. For example, the following names are used in the code: **network_func**, **epoch_round**, **Weights**, **Gradients**, **trainBatchSize**, etc.

I could make them consistent if only I knew which notation to use.

tools/torch/main.lua
Liuftvafas: I liked `valid` only because it aligns well in the code with `train` (same length). But we could use `val` and in filenames **val.txt** and **val_db**.

gheinrich: Yes I think `val` is good. It's short and it's already in other parts of the code to denote the validation phase. Thanks!

tools/torch/main.lua
gheinrich: Indeed the trouble is that the naming conventions are inconsistent in the first place... I was just noting that you seemed to favor camel case notation for globals [there](https://github.com/NVIDIA/DIGITS/pull/495/files#diff-1a7e0f05f8c1adb8faaf9f4f5cda363eR185) and underscore notation for locals [there](https://github.com/NVIDIA/DIGITS/pull/495/files#diff-1a7e0f05f8c1adb8faaf9f4f5cda363eR592). I think that's fine. However [this change](https://github.com/NVIDIA/DIGITS/pull/495/files#diff-1a7e0f05f8c1adb8faaf9f4f5cda363eR493) introduced a new global with underscore notation so I was thinking this wasn't entirely consistent... I am not suggesting that we fix all of these in this pull request but maybe if we could stick to one policy for new code that would be nice.

tools/torch/main.lua
```diff
@@ -582,70 +584,76 @@ end
...
-                inputs=inputs:float()
+                inputs = inputs:float()
                 targets = targets:float()
             end
 
             local y = model:forward(inputs)
-            local labels = network.labelHook and network.labelHook(inputs, targets) or targets
-            local err = loss:forward(y,labels)
+            local labels = label_function(inputs, targets)
+            local err = loss:forward(y, labels)
```
gheinrich: `loss` is another global that should be converted to a function parameter

gheinrich:
Thanks @Liuftvafas! Can you squash your commits and then I think the change will be good to merge.

For the record, I tried moving Validation() out of `main.lua` into a separate `.lua` file and this worked fine.

Liuftvafas:
Moving Validation() to separate file was really what I had to do for testing (but I didn't this time). Thanks.

Just to clarify, do you suggest to squash all 4 commits from this pull request into a single one?

gheinrich:
Yes please, we usually squash commits into a single one before merging a pull request, unless several authors contributed. This allows keeping the commit history more concise.

gheinrich:
The change looks good to me, thanks!

Issue #49, opened 2015-04-07
load a caffe pre-trained model?


mnmoustafa:
It is not an issue per se.
Is it possible to load a pre-trained model from caffe (still nvidia branch) command line? For instance, if I want to start from the imagenet model (and weights) and modify/train the last layer, etc... or load something from the zoo.
Thanks,

lukeyeager:
It is possible, but it's not well documented. On the "New Image Classification Model" page, under the "Custom Network" tab, you have two form fields:

1. **Custom network** - this is where you paste in the text of the train_val prototxt that you want to use.
2. **Pretrained model** - if you know what you're doing, you can enter the path to a pretrained model here that corresponds to the network that you've entered above. It's not a file upload, you have to actually input the path to the file on the local filesystem.

mnmoustafa:
I see. Thank you. But how about loading a prebuilt data db?
On the "new image classification model" page, I see that it accepts only the actual jpg images. 
My use case is as follows: I started from the CIFAR10 dataset (which is distributed as binaries not jpg images) and trained a network using caffe command-line. Can I continue fine tuning and testing individual images on this network with DIGITS?
Thanks,

lukeyeager:
You can definitely import your pretrained model and fine-tune it in DIGITS on other images. If you want to use the CIFAR10 dataset again, you'll have to extract the images from the distributed binary format into a folder of folders of images, like DIGITS expects, and create a dataset in DIGITS.

mnmoustafa:
Fair enough. Thank you lukeyeager for the prompt response. I appreciate if you consider adding an option in the future to import prebuilt leveldb data for instance.
Thanks,

lukeyeager:
> consider adding an option in the future to import prebuilt data

Will do, thanks for the suggestion!

kajwes:
Hi, I've got an additional question on this topic. I would like to create a model using custom network + pretrained model, but without having to specify a dataset. Since the model is already pretrained, I would like to be able to run the "Classify one image" or "classify many images" without having to train the network again.
Is this possible, or is there a simple way to create a "dummy dataset" which will work for let's say the ImageNet pretrained model?

lukeyeager:
> I would like to be able to run the "Classify one image" or "classify many images" without having to train the network again.

DIGITS is not a *deployment solution*, it is a **training system**. As such, it has been designed assuming that datasets are created and models are trained within DIGITS. Also, the emphasis has been on making training efficient, not inference.

[...]

PR #498, opened 2016-01-08
Add option to use multiple pretrained networks

IgorX2:
I've added an option to copy weights from multiple pretrained networks so that ensembles can be built easily. Multiple pretrained networks can be added as semicolon delimited paths.

### DIFF
--- a/digits/model/forms.py
+++ b/digits/model/forms.py
@@ -256,7 +256,7 @@ class ModelForm(Form):
         if form.method.data == 'custom':
             snapshot = field.data.strip()
             if snapshot:
-                if not os.path.exists(snapshot):
+                if not all(map(lambda x: os.path.exists(x), snapshot.split(';'))):
                     raise validators.ValidationError('File does not exist')
 
     # Select one of several GPUs
--- a/digits/model/forms.py
+++ b/digits/model/forms.py
@@ -248,7 +248,7 @@ class ModelForm(Form):
             )
 
     custom_network_snapshot = utils.forms.TextField('Pretrained model(s)',
-                tooltip = "Semicolon delimited paths to pretrained model files. Only edit this field if you understand how fine-tuning works in caffe."
+                tooltip = "Colon delimited paths to pretrained model files. Only edit this field if you understand how fine-tuning works in caffe."
             )
 
 
--- a/digits/model/forms.py
+++ b/digits/model/forms.py
@@ -254,9 +254,9 @@ class ModelForm(Form):
 
     def validate_custom_network_snapshot(form, field):
         if form.method.data == 'custom':
-            snapshot = field.data.strip()
+            snapshot = ':'.join(map(lambda x: x.strip(), field.data.split(':')))
             if snapshot:
-                if not all(map(lambda x: os.path.exists(x.strip()), snapshot.split(':'))):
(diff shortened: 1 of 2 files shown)


lukeyeager:
That's a neat idea. Can you link to a blog post or walkthrough that would let me test this out?

1. You should update the tooltip to let users know about this feature
2. Is there any way to add a test for this?

lukeyeager:
@gheinrich is this feasible for Torch as well?

IgorX2:
I needed this a few days ago so I added it and trained a MNIST network with it. I don't have it written down, but I could link a set of simple MNIST classifiers that can be used.

As for the tooltip, I'll try to update it tomorrow.

gheinrich:
@lukeyeager it is a challenge to implement this for Torch since Torch doesn't have named layers/parameters.

IgorX2:
@lukeyeager I have updated the tooltip and here is a link to three models for a neural network which can be used to test the modifications.

IgorX2:
@lukeyeager I have updated the tooltip. Here is a link to three neural network models that can be used to test the added functionality: http://www.igorsevo.com/File.ashx?type=res&resType=0&title=source+code%5cMNIST-Double.zip.

digits/model/forms.py
```diff
@@ -256,7 +256,7 @@ def validate_custom_network_snapshot(form, field):
         if form.method.data == 'custom':
             snapshot = field.data.strip()
             if snapshot:
-                if not os.path.exists(snapshot):
+                if not all(map(lambda x: os.path.exists(x), snapshot.split(';'))):
```
lukeyeager: Let's strip the whitespace around the separator.

lukeyeager: And use `:` instead of `;`.

IgorX2:
@lukeyeager I added the whitespace stripping and changed the delimiter and the tooltip.

lukeyeager:
This looks OK to me. Thanks for the updates!

I don't have a signed CLA from you yet, do I? Please follow the instructions here:

https://github.com/NVIDIA/DIGITS/blob/master/CONTRIBUTING.md#pull-requests

lukeyeager:
Oops, I forgot to ask you to squash your commits into one! Next time.

lukeyeager:
> And use : instead of ;.
@lukeyeager https://github.com/NVIDIA/DIGITS/pull/498#discussion_r49655374

Rats. That was a bad choice of separator on Windows machines. My bad.

> Oh, that's unfortunate, DIGITS is confusing the colon in D:\TEST_DIGITS with the separator we use in the presence of multiple pre-trained models.
@gheinrich https://github.com/NVIDIA/DIGITS/issues/49#issuecomment-226436025

We should probably use `os.path.pathsep` instead.
```py
# Linux
>>> os.path.pathsep
':'
# Windows
>>> os.path.pathsep
';'
```

IgorX2:
@lukeyeager: Fixed this and created a pull request #854.

gheinrich:
Oh, @lukeyeager has already pushed a fix with #851...

IgorX2:
I synced with the master branch, but the changes were not there. Also, there were issues when trying to use this because someone had overwritten the modifications to the `validate_custom_network_snapshot`  method. I corrected that here.

PR #499, opened 2016-01-11
Torch model fine-tuning

gheinrich:

PR #500, opened 2016-01-11
Fine-tuning tutorial

gheinrich:


gheinrich:
The Torch example depends on #499

examples/fine-tuning/create_dataset.sh
```diff
@@ -0,0 +1,22 @@
...
+NEW_DATASET_DIR=$1
+MNIST_DIR=$2
+
+# create new dataset directories
+mkdir -p $NEW_DATASET_DIR/odd
+mkdir -p $NEW_DATASET_DIR/even
+
+# create symbolic links
+for digit in 0 2 4 6 8
+   do cp $MNIST_DIR/$digit/*.png $NEW_DATASET_DIR/even/
```
lukeyeager: You could use symlinks here instead of copying the data.

https://github.com/NVIDIA/DIGITS/blob/v3.0.0-rc.3/docs/ImageFolderFormat.md#example-use-case---imagenet-subset

The only concern would be if symlinks aren't supported on an os, but if you're running `/bin/sh` I would expect symlinks to be supported.

May not be worth the effort, just thought I'd mention it in case you didn't know or hadn't thought of it.

lukeyeager:
Cool tutorial and nice write-up! I like the even/odd number problem.

It's a bit unfortunate that Caffe can solve this problem instantly without fine-tuning. If I point the standard Caffe LeNet network at the even/odd dataset, it still solves it within an epoch or two.

So I guess we're back to the weight initialization thing again? Is there any way to make the problem hard enough that Caffe can't solve it without fine-tuning, but easy enough that Torch still can solve it with fine-tuning?

examples/fine-tuning/create_dataset.sh
gheinrich: I actually used symlinks in my first implementation but noticed that shuffling didn't work across folders (?) so I was effectively only getting zeros for even numbers and ones for odd numbers... therefore I decided to do a hard copy in order to have a flat file structure. Does it make sense?

lukeyeager: Why do you say that shuffling wasn't working across folders? It looks like it's working to me ...

Are you looking at the file that you get when clicking on "Input File (before shuffling)"? As it says, that's the list of files before shuffling. You can look at the actual order in the database with the "Explore the db" tool.

gheinrich: If I create the dataset doing:
```
# create symbolic links
for digit in 0 2 4 6 8
   do ln -s $MNIST_DIR/$digit $NEW_DATASET_DIR/even/$digit
done
for digit in 1 3 5 7 9
   do ln -s $MNIST_DIR/$digit $NEW_DATASET_DIR/odd/$digit
done
```
Then the dataset tree has one sub-folder per digit and if I set the maximum number of samples per class to a small number I get only one type of digit per class (I just tried again and all even numbers of `8`s). Hence my suspicion that shuffling didn't work across folders but maybe this is only a problem in the presence of a "maximum samples per class"...

gheinrich:
Indeed the improvement isn't so dramatic when using Caffe... I just tried Caffe's LeNet without Xavier weight initialization and it behaves very much like Torch (accuracy stuck at 50%). Thanks for the feedback I'll see if I can find the right balance!

examples/fine-tuning/create_dataset.sh
lukeyeager: Oh I see, you're right. [The max_per_category check](https://github.com/NVIDIA/DIGITS/blob/v3.0.0-rc.3/tools/parse_folder.py#L393-L397) comes before [the shuffling](https://github.com/NVIDIA/DIGITS/blob/v3.0.0-rc.3/tools/parse_folder.py#L417).

Maybe I didn't notice it because I used the MNIST test images, not the training images?

gheinrich:
I tried doing something similar with Alexnet on CIFAR10 and a decomposition into animals or vehicles but I can't get Caffe to completely fail to learn anything unless I turn off Xavier weight initialization.

On MNIST odd and even numbers, there is some small benefit in doing fine tuning even with Caffe in that the accuracy is about 10% higher after fine tuning and the train/val loss functions are less erratic and exhibit less overfit on the training set.

Without fine tuning:
caffe-from-scratch

With fine tuning:
caffe-fine-tuned

Admittedly this is less than impressive... is it OK if I just make a note about this in the tutorial? I fear that it might take a while to come up with the perfect toy example for fine tuning.

lukeyeager:
> Admittedly this is less than impressive... is it OK if I just make a note about this in the tutorial? I fear that it might take a while to come up with the perfect toy example for fine tuning.

Yes for sure. It doesn't need to be perfect.

I do feel like we should document the difference between Caffe and Torch initialization somewhere through (maybe in `docs/GettingStartedTorch.md`?) and link to it from this tutorial.

gheinrich:
Thanks, so I've added a note and a screenshot to show how the default Caffe model (with Xavier initialization) is doing when trained from scratch. I think the example still shows the benefits of fine tuning (much less overfit and much higher accuracy after fine tuning).

PR #501, opened 2016-01-12
Fixup imports

lukeyeager:
This is a noop change. It simply improves the coding style, which should make development a bit easier.

1. Use `absolute_import` - [PEP 328](https://www.python.org/dev/peps/pep-0328/#rationale-for-absolute-imports)
  * Improves readability of imports
  * Allows us to name files with standard library names if we want (like `utils.logging`, for example)
  * Required if we ever want to upgrade to Python3

2. Alphabetize and re-order imports - [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports)
  * Each import in its own, rightful place (as far as possible)
  * Should minimize merge conflicts

### DIFF
--- a/digits-devserver
+++ b/digits-devserver
@@ -1,12 +1,11 @@
 #!/usr/bin/env python2
 # Copyright (c) 2014-2016, NVIDIA CORPORATION.  All rights reserved.
 
-import sys
 import argparse
-import digits
-
-# Only required if using SocketIO.run() and app.debug = False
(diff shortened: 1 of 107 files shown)


gheinrich:
Phew, this is a tough one to review. I can't say I have checked everything but in principle this looks OK to me.

lukeyeager:
> Phew, this is a tough one to review. I can't say I have checked everything but in principle this looks OK to me.

Indeed it is. Thanks for taking a look!

PR #507, opened 2016-01-13
Use flask blueprints to cleanup routing code

lukeyeager:
> Blueprints can greatly simplify how large applications work and provide a central means for Flask extensions to register operations on applications.
http://flask.pocoo.org/docs/0.10/blueprints/

This just helps with code quality and readability. This doesn't make any changes that are externally visible in the application.
* Add routes deliberately in `webapp.py` rather than implicitly through each of the `views.py` files
* Makes the `url_for()` statements match the actual code layout more closely

### DIFF
--- a/digits/dataset/images/classification/views.py
+++ b/digits/dataset/images/classification/views.py
@@ -15,10 +15,10 @@ from digits import utils
 from digits.dataset import tasks
 from digits.utils.forms import fill_form_if_cloned, save_form_to_job
 from digits.utils.routing import request_wants_json, job_from_request
-from digits.webapp import app, scheduler
+from digits.webapp import scheduler
 
 
-NAMESPACE = '/datasets/images/classification'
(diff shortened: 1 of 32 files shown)


digits/dataset/images/classification/test_views.py
```diff
@@ -129,9 +129,10 @@ def create_dataset(cls, **kwargs):
             s = BeautifulSoup(rv.data, 'html.parser')
             div = s.select('div.alert-danger')
             if div:
-                raise RuntimeError(div[0])
+                print div[0]
```
gheinrich: is this related to this change?

lukeyeager: You're right, it wasn't. New PR at #509.

gheinrich:
All the routes I tried worked for me, including a number of JSON routes.

PR #517, opened 2016-01-15
Allow analyze_db tool to read Datums without shape

### LINKED ISSUE #514: With encoded datums, Image Dimensions are set to 0x0x0
When generating an LMDB on the command line with the "-encode_type" parameter set (e.g., png, jpg), the associated datums are stored with width/height/channel all equal to zero.

image

This causes a set of other bugs:
a) random cropping isn't supported (Digits thinks that the patches are 0x0x0, so any crop size is invalid)
b) "Show visualizations and statistics" return an error "ValueError: unsupported number of channels: 0"

I can think of 3 solutions
a) modify convert_imageset.cpp (or more likely: io.cpp) in caffe such that it correctly sets the patch dimensions 
b) modify analyze_db.py such that if it detects datum.encoded=True and datum.width=0, load the image to determine the correct dimensions and then output those instead of the raw values 
c) allow the user specify it directly in the UI

I assume the first approach is the optimal one, but i'm unsure if it was left that way intentionally?

lukeyeager:
*Close #514*

### DIFF
--- a/tools/analyze_db.py
+++ b/tools/analyze_db.py
@@ -9,7 +9,14 @@ import os.path
 import sys
 import time
 
+# Find the best implementation available
+try:
+    from cStringIO import StringIO
+except ImportError:
+    from StringIO import StringIO
+
 import lmdb
+import PIL.Image
 
 # Add path for DIGITS package
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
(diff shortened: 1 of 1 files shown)


tools/analyze_db.py
```diff
@@ -108,7 +115,25 @@ def analyze_db(database,
         datum = caffe_pb2.Datum()
         datum.ParseFromString(value)
 
-        shape = '%sx%sx%s' % (datum.width, datum.height, datum.channels)
+        if (not datum.HasField('height') or datum.height == 0 or
```
gheinrich: maybe we can skip this the second time if `enforce_same_shape==False`?

lukeyeager: Oh yeah, duh. Thanks for catching that!

lukeyeager: Updated. Now it prints this if you don't add `--force-same-shape`:
```
2016-01-15 13:02:49 [INFO ] 79 entries found with shape ?x?x? (WxHxC)
2016-01-15 13:02:49 [INFO ] 1 entries found with shape 28x28x3 (WxHxC)
```

Issue #515, opened 2016-01-15
Siamese network example


danst18:
I have followed all the steps of the Siamese network example but the loss function keeps oscillating:

![capture](https://cloud.githubusercontent.com/assets/6713977/12354026/f89df0dc-bb8a-11e5-934b-ce59958a346d.JPG)

It is worth mentioning that I am running DIGITS on Windows, however I've never had any problem.

I think that the issue is in the dataset creation with the `create_db.py` script. I first ran the scrip on Windows and it took more than 6 hours to finish. The `train_mean.png` image looks like this, which makes me think that something is wrong (a 3 x 28 pixels image with black pixels in the first column and white the rest):

![train_mean](https://cloud.githubusercontent.com/assets/6713977/12354113/b5f010e8-bb8b-11e5-86fe-6482d2a0d5c5.png)

I've also run `create_db.py` on a Ubuntu virtual machine and, although it just takes 15 min to finish, the result is still the same (the same mean image and the training behaves in the same way). 

I've also tried to run the example using plain Caffe (without DIGITS) with no success.

gheinrich:
Hi @danst18 there is indeed an issue in the creation of the mean `.png` files since the channel transpose is done twice. If you comment out these [lines](https://github.com/NVIDIA/DIGITS/blob/master/examples/siamese/create_db.py#L155-L158) it should work better, although I don't think that is the root cause for your issue since these mean images are not used during training (we're only using `train_mean.binaryproto` - did you provide this file when you created the dataset?)

I just tried running this tutorial on the tip of the master branch and the loss function is similar to that of the tutorial (the validation loss goes down to 0.05:
![siamese-master-loss](https://cloud.githubusercontent.com/assets/3889770/12358057/b569e7c8-bbab-11e5-88a3-eb734a099bb3.png)

Besides, when I perform inference on the generated test images the result looks good (images of the same class produce similar vectors and images from different classes produce different vectors).

Perhaps you could try reducing your learning rate and let training run for a little while longer?

You can also check that the mean is correctly subtracted by enabling vizualisations during inference. Your data layer should look like this (the distribution of values should be mostly symmetric around zero):
![siamese-data-layer](https://cloud.githubusercontent.com/assets/3889770/12358253/9dec6cc8-bbac-11e5-92d5-f486428efeed.png)

lukeyeager:
Can this issue be closed now that #519 is merged, or is this a separate issue?

danst18:
[...]

PR #486, opened 2015-12-23
Siamese network tutorial for Torch

gheinrich:

PR #519, opened 2016-01-17
Siamese DB creation fixes

gheinrich:
> Enable batch writing in order to speed up DB creation.
fix slowliness issue reported in #515 and #486

> Remove extra channel transpose - already done in _save_image()
Fixes one issue reported in #515


lukeyeager:
LGTM.

PR #522, opened 2016-01-18
Weight initialization example

gheinrich:
Added examples showing the effects of various weight initialization schemes so I can make a reference to this in #500


examples/weight-init/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+Table of Contents
+=================
+* [Introduction](#introduction)
+* [Selecting a weight initialization strategy](#selecting-a-weight-initialization-strategy)
+* [Experiments with LeNet on MNIST](#experiments-with-lenet-on-mnist)
+
+## Introduction
+
+Weight initialization can critically affect the speed at which a neural network is able to learn. Under certain circumstances,
```
lukeyeager: Typically, you put each sentence on a different line in a Markdown file rather than chopping off the sentences at 80 characters. The former scheme makes it easier to review changes later, while the latter scheme makes it harder. Neither should affect the way the paragraph is actually displayed.

examples/weight-init/README.md
```diff
@@ -0,0 +1,93 @@
...
+      value: 0.2
+    }
+```
+
+The initialization of the bias terms can also be specified through a `bias_filler` field, although this tends to have a lesser impact
+on learning so we will focus on the initialization of weights in the following examples.
+
+### Under Torch7
+
+Most layers make it straightforward to initialize weights randomly from a uniform distribution over `[-std, std]` by calling their `reset(std)` method. However the determination
```
lukeyeager: Can you post a little bit of code here to show how to actually do this for a layer in your network (like you did above for Caffe)? I would hope to be able to read this guide and then know enough to go in and edit the default Torch AlexNet model to change the weight initialization.

gheinrich: Sure, good idea.

examples/weight-init/README.md
lukeyeager: Sure! I did a little test to try it out myself:
https://github.com/lukeyeager/github-testing/blob/2d2fc50ff9/CONTRIBUTING.md

gheinrich:
Thanks I have updated my commit.

PR #523, opened 2016-01-19
Use `cls` for classmethod instead of `self`.

dongjoon-hyun:


lukeyeager:
Whoops, looks like I got a little sloppy with my refactoring at some point. Thanks!

As you know, using `self` here isn't a functional error, but a style error ([PEP 8](http://legacy.python.org/dev/peps/pep-0008/#function-and-method-arguments)).

dongjoon-hyun:
Sure! It's just for avoiding confusion. :) 

By the way, @lukeyeager , could you add `@classmethod` to `prompt_message(cls)` in `digits/config/log_file.py`, too. It's vice versa of this PR. I found a few minutes ago.

lukeyeager:
> could you add `@classmethod` to prompt_message(cls) in digits/config/log_file.py, too?

Looks like it, yes. Good catch!

dongjoon-hyun:
Thanks!

PR #530, opened 2016-01-20
Fix nginx redirection to ports other than 80

### LINKED ISSUE #528: Port reconfiguration not working (deb package)
Hi

I have use this command sudo python -m digits.config.edit -v and changed the DIGITS port to 5000 
Then i have configure Data set Configuration 
Its URL is bellow
http://localhost:5000/datasets/images/classification/new

after click on create its redirect to http://localhost/datasets/20160120-104536-a3ef and Browser shows ERR_CONNECTION_REFUSED

Its seems to be not redirecting to port 5000 . Then i change url manually to

http://localhost:5000/datasets/20160120-104536-a3ef Its shows me the output.

lukeyeager:
*Fix #528*

After using `dpkg-reconfigure` to change the reverse proxy front-end port from 80 to something else (e.g. 5000), redirects were still going back to port 80.

~~Solution based on answers to [this SO post](http://serverfault.com/questions/576365/use-nginx-reverse-proxy-for-redirection).~~

/cc @dadap @3XX0 - anyone know of a more elegant fix?

### DIFF
--- a/nginx.site
+++ b/nginx.site
@@ -8,7 +8,7 @@ server {
         proxy_pass http://127.0.0.1:34448;
         proxy_redirect off;
 
-        proxy_set_header Host $host;
+        proxy_set_header Host $host:$server_port;
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 


lukeyeager:
Thanks for the suggestion @3XX0, this is clearer.

PR #534, opened 2016-01-21
Allow non-integer step size in visualize_lr()

### LINKED ISSUE #532: Error when pressing Visualize LR if Step Size given with decimal point
While creating a new model from an old one with clone model, DIGITS (Version 3.1-dev) sets the Step Size format on its own to decimal (e.g. from 5 to 5.0). This then throws an error when Visualize LR is pressed. If corrected back to pure integer value, the error goes away.  

500: INTERNAL SERVER ERROR
ValueError
invalid literal for int() with base 10: '5.0'

gheinrich:
closes #532

### DIFF
--- a/digits/model/views.py
+++ b/digits/model/views.py
@@ -165,7 +165,7 @@ def visualize_lr():
     if policy == 'fixed':
         pass
     elif policy == 'step':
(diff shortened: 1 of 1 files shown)


lukeyeager:
LGTM

PR #539, opened 2016-01-22
Allow standard Alexnet model for Torch to train on more than 2 GPUs

gheinrich:


lukeyeager:
Any reason why this wasn't turned on earlier?

gheinrich:
This is a remnant from the model I imported from [there](https://github.com/soumith/imagenet-multiGPU.torch/blob/5f8a7b28be9690de92c531316a36375bab0ecb25/models/alexnet_cudnn.lua#L3). Now this limitation does not apply anymore. I forgot to remove this as I had only tested up to 2 GPUs (my workstation has only 2).

lukeyeager:
Gotcha. LGTM!

PR #541, opened 2016-01-25
Only look in one location for config files

lukeyeager:
Keep using `digits/digits.cfg` but don't look at `$HOME/.digits.cfg` or `/etc/digits.cfg`.

**Why**
It's not that useful after all and creates too many opportunities for confusing configuration errors.

**How**
This is the quickest way to make the change. It ends up leaving a bunch of cruft in the `digits/config/*` code to handle multiple config files. That could be addressed later.

### DIFF
--- a/digits/config/config_file.py
+++ b/digits/config/config_file.py
@@ -114,24 +114,7 @@ class ConfigFile(object):
             cfg.write(outfile)
 
 
-class SystemConfigFile(ConfigFile):
-    def __init__(self):
-        if platform.system() in ['Linux','Darwin']:
-            filename = '/etc/digits.cfg'
-        else:
-            filename = None
-        super(SystemConfigFile, self).__init__(filename)
-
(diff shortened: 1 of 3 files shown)


lukeyeager:
I'm getting a bunch of Torch errors in the build. @gheinrich do those look familiar? I just built from ToT on my box and didn't have a problem.

gheinrich:
It looks like all LMDB/Torch tests are failing with a SEGFAULT (HDF5 tests pass). This seems to be a regression in the current version of the lightningmdb wrapper on Luarocks registry. The latest version of Github is working. I have filed an issue on https://github.com/shmul/lightningmdb/issues/10.

lukeyeager:
Thanks for looking into it! I was having the hardest time trying to reproduce the problem.

lukeyeager:
Let's move the discussion about the luarocks package to https://github.com/NVIDIA/DIGITS/issues/544.

PR #550, opened 2016-01-27
Fix Torch handling of multistep LR policy

### LINKED ISSUE #547: Error when use "Step Down (arbitrary steps)" in learning rate policy
When I use "Step Down" (arbitrary steps) policy in the learinng rate,  set to "6,50", the digits generated error: 
ValueError: could not convert string to float: ,

gheinrich:
Add unit test and fix Torch handling of multistep LR policy

fix #547


lukeyeager:
Awesome, thanks for the tests!

PR #465, opened 2015-12-10
Cleanup messages about upgrading/loading previous jobs

lukeyeager:
Most users shouldn't run into these issues because they should never have jobs lying around with invalid pickle files.

This PR makes error-handling a little more robust (for other errors that will occur in the future) and a little more verbose (so we are motivated to keep adding robustness). It shouldn't change any real functionality.

Before (on my machine):
```
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
Upgrading CaffeTrainTask to version 3 ...
2015-12-10 15:33:00 [WARNING] Failed to load 21 jobs.
```
After removing print messages and setting `scheduler.verbose=True`:
```
2015-12-10 15:37:06 [INFO ] Loaded 108 jobs.
2015-12-10 15:37:06 [WARNING] Failed to load 21 jobs.
2015-12-10 15:37:06 [DEBUG] 20150910-144319-e67c - AttributeError: 'CaffeTrainTask' object has no attribute 'caffe_log_file'
2015-12-10 15:37:06 [DEBUG] 20150911-082638-757c - AttributeError: 'CaffeTrainTask' object has no attribute 'caffe_log_file'
2015-12-10 15:37:06 [DEBUG] 20150911-082652-4008 - AttributeError: 'CaffeTrainTask' object has no attribute 'caffe_log_file'
2015-12-10 15:37:06 [DEBUG] 20150911-083943-a65b - AttributeError: 'CaffeTrainTask' object has no attribute 'caffe_log_file'
[...]

PR #551, opened 2016-01-27
Fix error-handling on loading jobs with no job_id

lukeyeager:
I introduced an obscure bug in https://github.com/NVIDIA/DIGITS/pull/465. If you somehow created a pickle file without a job_id, then the `load_past_jobs()` function failed to catch the error (or rather, only caught it the first time), and the server failed to start. This should fix that.

/cc @jmancewicz


jmancewicz:
That fixed the load error.

PR #552, opened 2016-01-28
Make Torch regression example more robust

gheinrich:
Check if `inputShape` field of external parameters is not nil before calculating dimension of input tensor.
Set dimension to `1` by default for network visualization.


gheinrich:
Verified by Ryan

lukeyeager:
Do we want to backport this into `digits-3.0`?

PR #558, opened 2016-01-29
Add documentation for installing CUDA

lukeyeager:
Enough questions have come up that I feel like this deserves its own doc.

### DIFF
--- a/docs/InstallCuda.md
+++ b/docs/InstallCuda.md
@@ -1,0 +1,124 @@
+# CUDA Installation
+
+Getting CUDA and the NVIDIA driver installed correctly on your machine can be tricky.
+This guide should tell you everything you need to know (on Ubuntu, at least!).
+
+Another good resource is the [CUDA installation guide for Linux](http://docs.nvidia.com/cuda/cuda-installation-guide-linux).
+
+## GPU
+
+You will need an NVIDIA GPU if you want to use CUDA.
+**If you want to use [cuDNN](https://developer.nvidia.com/cudnn), you will need a GPU with compute capability >= 3.0.**
+To find out what the compute capability of your card is, check one of these websites:
+
+* https://developer.nvidia.com/cuda-gpus
+* https://en.wikipedia.org/wiki/CUDA#Supported_GPUs
+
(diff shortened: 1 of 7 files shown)


docs/InstallCuda.md
```diff
@@ -0,0 +1,124 @@
...
+On Ubuntu, you can install a driver in two ways: with a `run` file or with a Debian package.
+
+It is recommended that you use a Debian package to install your driver, unless you have a new GPU that requires a newer driver version.
+Debian packages are simpler to install, uninstall and upgrade, while `run` file installers are useful if you need a newer driver version.
+
+To install with a `run` file, download one from the [NVIDIA Driver Downloads](http://www.nvidia.com/Download/index.aspx) website and follow the instructions.
+If you run into any problems, look at the "Additional Information" section.
+
+> **IMPORTANT**: If you use a `run` file to install your driver, don't install the `cuda` Debian package.
+More information [below](#deb-packages).
```
gheinrich: shouldn't it be `#debian-packages`?

gheinrich:
Thanks for the doc!

docs/InstallCuda.md
lukeyeager: Yeah we keep waffling on this throughout the documentation. Scanning through our other documentation [here](https://developer.nvidia.com/cuda-downloads) and [here](http://docs.nvidia.com/cuda/cuda-installation-guide-linux), it looks like we're not consistent there, either. Although Andy does overall seem to prefer "RPM vs. Deb" to "RPM vs. Debian".

In my mind, "deb" is an archive format. You can use some packaging tools to build "Debian deb packages" for a Debian distro, or "Ubuntu deb packages" for a Ubuntu distro.

It seems like a matter of opinion to me - I can't find anything that makes a case for "the right way" to talk about it.

gheinrich: Sure I don't have a strong opinion either - actually what I meant really is the link is broken :-)

lukeyeager: I decided to go with "Deb" like we do in the CUDA installation docs.

Issue #536, opened 2016-01-21
Problem with Batched Inference in example.py script


michaelholm-ce:
I may be missing something here, or I have found a bug.

I believed that changing the batch_size to something greater than 1 (i.e. batched inference) should give the same individual scores, just maybe complete the task faster.  However, I have found that changing the batch_size actually changes the resulting scores (they are no longer correct).

Am I in need of a deeper lesson on neural networks, or have I found a bug?

Thanks,

Michael

lukeyeager:
I assume you're talking about the script at `examples/classification/example.py`.

Are you running into this bug - #409? Which version of DIGITS do you have?
```sh
$ ./digits-devserver --version
3.1-dev
```

michaelholm-ce:
I'm running the script at examples/classification/example.py with digits version 3.1-dev.

I did read issue #409 and made the deep copy scores change, but the problem persists.

michaelholm-ce:
More specifically, I changed the example.py script only in the following way:

I replaced "image_files = [args['image']]" with 
 im_path = args['image']
 if os.path.isdir(im_path):
      image_files = [os.path.join(im_path,f) for f in os.listdir(im_path)]
  elif os.path.isfile(im_path):
      image_files = [im_path]

I have a directory with N images.  When I run the images individually, I get the correct scores.  When I run the whole directory, the first image scores correctly and the remainder score incorrectly.

gheinrich:
I am wondering why you had to change the code to to a deep copy of scores since the change should already be there on 3.1-dev? (see this [line](https://github.com/NVIDIA/DIGITS/blob/master/examples/classification/example.py#L143)). Or am I missing anything?

michaelholm-ce:
You're right @gheinrich -- my mistake.  The truth is that I first saw the problem when working with an older version (when I did make the deep copy change).  When the problem persisted, I updated to version 3.1-dev (where I made the other changes listed above) and the problem persisted.  Thanks for pointing that out.

gheinrich:
I tried making the same changes as you did however it feels to me like predictions are correct. Did you try using `use_archive.py`?

michaelholm-ce:
To clarify @gheinrich , are you saying that you ran `example.py` on a full directory (even as small as two images), and the output scores were identical to the scores you get when you run `example.py` on the images individually?

[...]

PR #559, opened 2016-01-29
Improvements to examples/classification/

lukeyeager:
This lets you classify multiple images without needing to edit the example at all. I found it useful while investigating https://github.com/NVIDIA/DIGITS/issues/536.

* Allow multiple input image files
* Expose batch_size as an option

### DIFF
--- a/examples/classification/example.py
+++ b/examples/classification/example.py
@@ -106,7 +106,7 @@ def load_image(path, height, width, mode='RGB'):
     image = scipy.misc.imresize(image, (height, width), 'bilinear')
     return image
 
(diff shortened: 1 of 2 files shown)


lukeyeager:
FYI, you can use a command like this to:

1. Read a list of images from a text file
2. Strip off any label information at the end of each line
3. Pass the image files as command-line arguments to `use_archive.py`:

```sh
cat list-of-images.txt | awk '{print $1}' | \
    xargs ./use_archive.py 20160129-133556-a640_epoch_3.0.tar.gz --batch-size=10
```

PR #561, opened 2016-02-01
Fix unmatched local variable initialization.

dongjoon-hyun:
This is just about removing redundant variable `v` in lua code.

### DIFF
--- a/tools/torch/data.lua
+++ b/tools/torch/data.lua
@@ -28,7 +28,7 @@ end
 local function all_keys(cursor_,key_,op_)
     return coroutine.wrap(
         function()
-            local k = key_,v
+            local k = key_
             repeat
                 k,v = cursor_:get(k,op_ or MDB.NEXT)
                 if k then
--- a/tools/torch/data.lua
+++ b/tools/torch/data.lua
@@ -29,6 +29,7 @@ local function all_keys(cursor_,key_,op_)
     return coroutine.wrap(
         function()
             local k = key_
+            local v
             repeat
                 k,v = cursor_:get(k,op_ or MDB.NEXT)
                 if k then


dongjoon-hyun:
The test fails on downloading packages related Caffe installation step. 
```
W: Failed to fetch http://security.ubuntu.com/ubuntu/dists/precise-security/main/binary-i386/Packages  Hash Sum mismatch
```
Should I amend the commit to trigger Trivis-CI again?

lukeyeager:
I'll restart it manually

dongjoon-hyun:
Thank you!

gheinrich:
Thanks @dongjoon-hyun - what about my concern that this change risks overwriting a global variable? Isn't it better to use local variables? Sorry I don't understand what problem this change addresses.

dongjoon-hyun:
Oh, I see what is your point. The new one works the same way. This is just about a syntax fix. 
In fact, the original one is also overwriting a global variable since it use `v` as a *rvalue*, not a *lvalue* in here. Isn't it?
```
local k = key_,v
```

dongjoon-hyun:
In this assignment, `v` is not used actually. Only `key_` is assigned into local variable `k`.

gheinrich:
I see thanks, `v` is used as rvalue however since it is `nil` it isn't written into `k`...
So I think the fix is to change:
```
local k = key_,v
```
to:
```
local k = key_
local v
```
Do you agree?

dongjoon-hyun:
Right. I agree that's a more correct way. I just thought a fix on syntax, but you improves it. :) 
I'll improve like that, then. :)

dongjoon-hyun:
I've changed. Thank you for your thoughtful review, @gheinrich .

gheinrich:
That looks good to me, thanks!

dongjoon-hyun:
Thank you, @gheinrich !

PR #568, opened 2016-02-04
Fix links to jobs on job details page

### LINKED ISSUE #567: Bad links to models in "Models" page
In Models page (accessed through "View details") when clicked on model name the resulting page is not found. The URL is `localhost/jobID` and I guess it should be `localhost/models/jobID` as then it works as expected.

gheinrich:
closes #567


lukeyeager:
LGTM

Issue #479, opened 2015-12-17
gunicorn + nginx issues


wghilliard:
1. I am currently using digits on large data samples (225000+) and after I train a model I am trying to tune my data samples by using classification stats via the API's classify many. I have noticed that when I attempt to classify >4500 images, the server will timeout and produce a 502 error. If I append the gunicorn call in /etc/init/nvidia-digits-server.conf with "--timeout=300" the server will timeout after 60 seconds and produce a 504 error. After some reading I found out nginx default is 60 seconds and you must append the /etc/nginx/sites-available/digits.site with "proxy_read_timeout 300;". Maybe it was due to my naivety, but I didn't find this fix immediately obvious. Should the docs be modified to cite this issue for users of my skill level?
2. I noticed that when posting to the API's classify many that the webapp will hang and result in either a timeout or an extremely delayed response. I would attribute this to only having a single gunicorn worker. I tried appending "--workers=4" to the aforementioned gunicorn call which resulted in intermittent 404 errors when you tried to look at jobs via the webapp along with failed model training. With this in mind, my question is: Is Digits designed to run with multiple worker instances? If not, why wrap Flask with gunicorn?

lukeyeager:
Re (1):

Thanks for reporting the issue! This is a known issue and needs to be addressed in the release notes/documentation. Here's a previous discussion of the issue:

> I think the best way to solve this would be with the intermediate results / progress bar feature I mentioned earlier. Then the page can return quickly with a "0/1,300,000 images processed" page, and then slowly return the data as it comes through. That's a lot of work, and won't get implemented quickly. For now I'd suggest breaking up your huge textfile into manageable chunks.
https://github.com/NVIDIA/DIGITS/issues/70#issuecomment-93782536

The basic issue is that inference is that the request doesn't return a response until all of the classifications are complete (which could take a very long time). We've prioritized other things for the v3.0 release and didn't have time to get around to it.

So far, we've been treating DIGITS as primarily a deployment solution without working to make it a quality inference solution:

> DIGITS is not a deployment solution, it is a training system. As such, it has been designed assuming that datasets are created and models are trained within DIGITS. Also, the emphasis has been on making training efficient, not inference.
https://github.com/NVIDIA/DIGITS/issues/49#issuecomment-99533854

But realistically, people are using it for inference and it needs to work - or at least not crash the server.

[...]

Issue #70, opened 2015-04-15
Output the testing results of many images as text file


tmquan:
Thanks to digits, the new feature of "image_classification_model_classify_many" works well. 

I just wonder is there any way to retrieve that result as text file from this part, to systematically analyze the output using pandas? It would be nice if this can save the dataframe of testing images somewhere in its job's directory
    
    classifications = []
        for image_index, index_list in enumerate(indices):
            result = []
            for i in index_list:
                # `i` is a category in labels and also an index into scores
                result.append((labels[i], round(100.0*scores[image_index, i],2)))
            classifications.append(result)

michael-george-hart:
Did you have a look at my code in  #61... That is how I directly store the results to a csv file....

tmquan:
Thanks to michael-geoge-hart. 

I did the same strategy to output the result to text file or csv, but it is really messy and I dont want to alter the beautiful structure of DIGITS. 

Just wonder to have this feature in the next version because it is very convenient to analyse on the test dataset. 

On the other hand, since the list of test data is oftent quite large (mine has around 1 milion images). It would be nice if we can have a status directly on the web browser, along side with the terminal info log. 

Regards,

michael-george-hart:
Totally agree .... @lukeyeager has done a really good job on DIGITS in terms of clear though and design/implementation of DIGITS .... it is really not that hard to add the additional features you desired if you look at #61

lukeyeager:
@tmquan, so you'd like to be able to download the "Classify Many" results as a `.csv`? That sounds pretty doable. As I mentioned [here](https://github.com/NVIDIA/DIGITS/issues/61#issuecomment-92518289), you should be able to copy and paste the text into excel and get your data fairly easily. But I agree that this would be a nice enhancement.

@michael-george-hart, you sound eager to help with this. If you could embed the data on the results page in the CSV format and provide a "Download Results" button that would be awesome. That would be much more user friendly than Tran's suggestion: 

> save the dataframe of testing images somewhere in its job's directory

lukeyeager:
> On the other hand, since the list of test data is often quite large (mine has around 1 million images). It would be nice if we can have a status directly on the web browser, along side with the terminal info log.

Are you talking about returning intermediate results? Or some kind of progress bar? @Sravan2j, this sounds like the kind of thing you were talking about. I agree, that would be nice.

michael-george-hart:
[...]

PR #573, opened 2016-02-10
Inference jobs

gheinrich:
Move inference to separate job

### Motivation

* Allow inference to run on different machine within cluster
* Allow long inference jobs to report progress through socketio
* Allow inference to reserve resources (GPUs)
* Image resizing done outside of server context
* Provide command-line inferface for inference using *exactly* the same path as DIGITS

### Features 

No new feature, will implement socketio updates in separate pull request

### Summary of changes

* New job and task created for inference
* To limit changes in source code, re-use existing inference API from model job/task
* No code duplication between Caffe/Torch, Classification/Generic
* New inference.py tool performs image resizing and inference
* Results are communicated back to DIGITS through a HDF5 file on filesystem

### Progress

- [x] Caffe single-image classification
- [x] Torch single-image classification
- [x] Caffe single-image inference
- [x] Torch single-image inference
- [x] Caffe multiple-image classification
- [x] Torch multiple-image classification
- [x] Caffe multiple-image inference
- [x] Torch multiple-image inference
- [x] Caffe layer visualizations
- [x] Torch layer visualizations
- [x] Delete job when done
- [x] Code Clean-up
- [x] Pass nostests
- [ ] Fix code coverage in presence of sub-processes
- [ ] Fix nosetest slowliness

### DIFF
--- a/digits/inference/errors.py
+++ b/digits/inference/errors.py
@@ -1,0 +1,21 @@
+# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
+from __future__ import absolute_import
+
+from digits.utils import subclass, override
+
+@subclass
+class Error(Exception):
+    pass
+
+@subclass
+class InferenceError(Error):
+    """
+    Errors that occur during inference
+    """
+    def __init__(self, message):
+        self.message = message
+
+    @override
+    def __str__(self):
+        return repr(self.message)
+
--- a/digits/inference/job.py
+++ b/digits/inference/job.py
@@ -1,0 +1,65 @@
+# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
+from __future__ import absolute_import
+
+from . import tasks
+import digits.frameworks
+from digits.job import Job
+from digits.utils import subclass, override
+
+@subclass
+class InferenceJob(Job):
+    """
+    A Job that exercises the forward pass of a neural network
+    """
+
+    def __init__(self, model, images, epoch, layers, **kwargs):
+        """
+        Arguments:
+        model   -- job object associated with model to perform inference on
+        images  -- list of image paths to perform inference on
+        epoch   -- epoch of model snapshot to use
+        layers  -- layers to import ('all' or 'none')
+        """
+        super(InferenceJob, self).__init__(**kwargs)
+
+        # get handle to framework object
+        fw_id = model.train_task().framework_id
+        fw = digits.frameworks.get_framework_by_id(fw_id)
+
+        # create inference task
+        self.tasks.append(fw.create_inference_task(
+            job_dir   = self.dir(),
+            model     = model,
+            images    = images,
+            epoch     = epoch,
+            layers    = layers))
+
+    @override
+    def __getstate__(self):
+        fields_to_save = ['_id', '_name']
+        full_state = super(InferenceJob, self).__getstate__()
+        state_to_save = {}
+        for field in fields_to_save:
+            state_to_save[field] = full_state[field]
+        return state_to_save
+
+    def inference_task(self):
+        """Return the first and only InferenceTask for this job"""
+        return [t for t in self.tasks if isinstance(t, tasks.InferenceTask)][0]
+
+    @override
+    def __setstate__(self, state):
+        super(InferenceJob, self).__setstate__(state)
+
+    def get_data(self):
+        """Return inference data"""
+        task = self.inference_task()
+        return task.inference_inputs, task.inference_outputs, task.inference_layers
+
+    @override
+    def is_read_only(self):
+        """
+        Returns True if this job cannot be edited
+        """
+        return True
+
--- a/digits/inference/tasks/inference.py
+++ b/digits/inference/tasks/inference.py
@@ -1,0 +1,207 @@
+# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
+from __future__ import absolute_import
+
+import base64
+import h5py
+import os.path
+import tempfile
+import re
+import sys
+
+import digits
+from digits import device_query
+from digits.task import Task
+from digits.utils import subclass, override
+from digits.utils.image import embed_image_html
+from digits.status import Status
+from ..errors import InferenceError
+
+@subclass
+class InferenceTask(Task):
+    """
+    A task for inference jobs
+    """
+
+    def __init__(self, model, images, epoch, layers, **kwargs):
+        """
+        Arguments:
+        model  -- trained model to perform inference on
+        images -- list of images to perform inference on
+        epoch  -- model snapshot to use
+        layers -- which layers to visualize (by default only the activations of the last layer)
+        """
+        # memorize parameters
+        self.model = model
+        self.images = images
+        self.epoch = epoch
+        self.layers = layers
+
+        self.image_list_path = None
+        self.inference_log_file = "inference.log"
+
+        # resources
+        self.gpu = None
+
+        # generated data
+        self.inference_data_filename = None
+        self.inference_inputs = None
+        self.inference_outputs = None
+        self.inference_layers = []
+
+        super(InferenceTask, self).__init__(**kwargs)
+
+    @override
+    def name(self):
+        return 'Infer Model'
+
+    @override
+    def __getstate__(self):
+        state = super(InferenceTask, self).__getstate__()
+        if 'inference_log' in state:
+            # don't save file handle
+            del state['inference_log']
+        return state
+
+    @override
+    def __setstate__(self, state):
+        super(InferenceTask, self).__setstate__(state)
+
+    @override
+    def before_run(self):
+        super(InferenceTask, self).before_run()
+        # create log file
+        self.inference_log = open(self.path(self.inference_log_file), 'a')
+        # create a file to pass the list of images to perform inference on
+        imglist_handle, self.image_list_path = tempfile.mkstemp(dir=self.job_dir, suffix='.txt')
+        for image_path in self.images:
+            os.write(imglist_handle, "%s\n" % image_path)
+        os.close(imglist_handle)
+
+    @override
+    def process_output(self, line):
+        self.inference_log.write('%s\n' % line)
+        self.inference_log.flush()
+
+        timestamp, level, message = self.preprocess_output_digits(line)
+        if not message:
+            return False
+
+        # progress
+        match = re.match(r'Processed (\d+)\/(\d+)', message)
+        if match:
+            self.progress = float(match.group(1))/int(match.group(2))
+            return True
+
+        # path to inference data
+        match = re.match(r'Saved data to (.*)', message)
+        if match:
+            self.inference_data_filename = match.group(1).strip()
+            return True
+
+        return False
+
+    @override
+    def after_run(self):
+        super(InferenceTask, self).after_run()
+
+        # retrieve inference data
+        inputs = None
+        visualizations = []
+        outputs = {}
+        if self.inference_data_filename is not None:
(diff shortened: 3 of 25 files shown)


gheinrich:
A couple of issues:

#### Moving inference to a dedicated job adds a significant amount of latency

Loading the model and classifying one image on LeNet takes less than 0.2s on my machine however the whole process of running the job and reporting results back to the user takes over three seconds.

First, there is a 1-second delay (when not in test mode) before starting the job, followed by a couple of random 0.3s-0.5s delays to move the scheduler state machine to a state where tasks can be started. Then the inference sub-process takes 1.1s - of which 0.5s is spent loading DIGITS config. Most of the remaining time is spent starting the Python process and importing packages.

Even in test mode, the overhead is significant as it takes over 20 minutes to run the nose test suite (Travis test time is now about 40 minutes).

Possible areas of improvement would be: reduce delays, rework code to not have to load DIGITS config from inference sub-process. Is this something we need to do before merging this Pull Request?

#### Code coverage down as sub-processes not taken into account

Two solutions are presented [there](http://coverage.readthedocs.org/en/coverage-4.0.3/subprocess.html). None sound particularly elegant.

lukeyeager:
Neither of those issues seem like showstoppers to me.

##### Speed

That all sounds correct to me. Nice profiling work, what did you use to get the timings?

* 1-second delay - we could try to work around this with some more clever logic for dealing with SocketIO message race conditions. The delay is a hack we should eliminate at some point anyway.
* 0.3-0.5 second delays - It'd be cool to move to a more event-based scheduler that didn't require all those calls to `sleep()`.
* Subprocess loading delays - I don't think there's any way around that, unfortunately.
* Docker startup time - we haven't even gotten to this yet, but I expect it may be substantial  despite what @3XX0 promises me :wink:.

##### Code coverage

Bummer, I hadn't thought of that. Have you tried any of those hacks they suggested? They would be nice to have for our other worker processes (like `create_db` or `parse_folder`), too.

gheinrich:
> Nice profiling work, what did you use to get the timings?

Oh, nothing particularly clever, sorry! Just a code review and a few `print` statements. I'm feeling dumb :-)

lukeyeager:
> Oh, nothing particularly clever, sorry! Just a code review and a few print statements.

Haha that works too :smirk:

digits/inference/errors.py
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
+from __future__ import absolute_import
+
+from digits.utils import subclass
+
+@subclass
```
lukeyeager: FYI `@subclass` doesn't really do anything if you're not also using `@override`. But I'm fine with leaving it in if you think it makes the code more readable.

digits/inference/job.py
```diff
@@ -0,0 +1,56 @@
...
+import digits.frameworks
+from digits.job import Job
+from digits.utils import override
+
+class InferenceJob(Job):
+    """
+    A Job that exercises the forward pass of a neural network
+    """
+
+    @override
```
lukeyeager: Similarly, `@override` is a no-op on classes without `@subclass`. And you can always override `__init__` - there's no need to check that the parent class has that function.

gheinrich: Thanks for the info I've been using those `@override` and `@subclass` without really understanding how they worked.

lukeyeager:
You've got some UI issues with `classify_many`:
separate-inference-ui-bug

lukeyeager:
Let's remove the `Computing visualizations for` print statement to avoid all these warnings:
```
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "scale" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "slice_triplet" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "conv1_left" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "pool1_left" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "conv2_left" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "pool2_left" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "ip1_left" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "relu1" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "ip2_left" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "feat_left" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "conv1_right" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "pool1_right" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "conv2_right" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "pool2_right" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "ip1_right" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "relu1_right" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "ip2_right" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "feat_right" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "discard" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [WARNING] Infer Model unrecognized output: Computing visualizations for "loss" ...
2016-02-16 17:52:28 [20160216-175223-c0c4] [INFO ] Infer Model task completed.

```

lukeyeager:
I've got two related requests, but we may decide to push them back to a later PR.

1. Can we show the inference page immediately, like we do for other jobs? That would require sending the resulting data to the page through SocketIO.
2. If we're already sending data over SocketIO, can we send it incrementally as we get it?
  * Would solve https://github.com/NVIDIA/DIGITS/issues/70#issuecomment-93782536 and https://github.com/NVIDIA/DIGITS/issues/479#issuecomment-165810024

digits/inference/tasks/inference.py
```diff
@@ -0,0 +1,189 @@
...
+            db.close()
+        self.inference_log.close()
+
+        # save inference to data for further use
+        self.inference_inputs = inputs
+        self.inference_outputs = outputs
+        self.inference_layers = visualizations
+
+    @override
+    def offer_resources(self, resources):
```
lukeyeager: Can we reserve a GPU for inference, since we are in fact using one? That seems logical, and would have saved me from some confusing out-of-memory errors. I think you can tell pycaffe which one you want to use.

It'd be cool if we could fall back to CPU-only if there are no GPUs available, so you don't have to wait for all the long-running training jobs to finish first. We could still use the `inference_task_pool` resource to limit the number of CPU inference tasks running at once.

gheinrich:
Hi Luke, I have pushed a few more commits to address your remarks. I will squash everything once that is OK.
If you mind I'd like to address the incremental inference notifications in a separate PR as this will require more work.

digits/model/tasks/caffe_train.py
```diff
@@ -54,6 +54,14 @@ def upgrade_network(network):
         #TODO
         pass
 
+    @staticmethod
+    def set_mode(gpu):
+        if gpu>=0:
```
lukeyeager: Can we use `gpu=None` and `gpu is None` instead of `gpu=-1` and `gpu>=0`? Sorry for being nitpicky, but I feel like it's a little less cryptic and more pythonic.

digits/inference/job.py
lukeyeager: No problem. In general I wouldn't use `@override` for any "dunder" functions (i.e. `__init__`, `__str__`, `__getstate__`), but I wouldn't worry about changing it in this PR - it doesn't hurt anything.

There's really not much to these functions at all:
https://github.com/NVIDIA/DIGITS/blob/v3.1.0/digits/utils/__init__.py#L75-L99

digits/model/tasks/caffe_train.py
```diff
@@ -1049,61 +1057,15 @@ def can_infer_one(self):
...
-                    snapshot_epoch=snapshot_epoch,
-                    layers=layers,
-                    )
-        elif isinstance(self.dataset, dataset.GenericImageDatasetJob):
-            return self.infer_one_generic(data,
+    def infer_one(self, data, snapshot_epoch=None, layers=None, gpu=-1):
+        if isinstance(self.dataset, dataset.ImageClassificationDatasetJob) or isinstance(self.dataset, dataset.GenericImageDatasetJob):
+            return self.infer_one_image(data,
                     snapshot_epoch=snapshot_epoch,
                     layers=layers,
```
lukeyeager: You're missing a `gpu=gpu` here which is causing you to never call `set_mode_gpu()`.

lukeyeager:
> If you mind I'd like to address the incremental inference notifications in a separate PR as this will require more work.

Yeah that's fine. We'll have to decide what to do about the JSON interface, too.

lukeyeager:
While we're fixing stuff, can you remove this line:

https://github.com/gheinrich/DIGITS/blob/8b6022162c/digits/model/tasks/caffe_train.py#L1242

As @anuragphadke discovered in https://github.com/NVIDIA/DIGITS/issues/581, it's unnecessary.

digits/inference/tasks/inference.py
lukeyeager: Looks great now, thanks!

digits/model/tasks/caffe_train.py
```diff
@@ -54,6 +54,14 @@ def upgrade_network(network):
         #TODO
         pass
 
+    @staticmethod
+    def set_mode(gpu):
+        if gpu is not None:
+            caffe.set_mode_gpu()
```
lukeyeager: I think you need to put `set_device()` BEFORE `set_mode_gpu()`. Odd.

https://groups.google.com/d/msg/caffe-users/HqdziMqpu6o/2zgy7MK3-hAJ
https://github.com/BVLC/caffe/pull/507

gheinrich:
Rebased/squashed and I switched the order of `set_device()` and `set_mode_gpu()`

lukeyeager:
The UI leaves something to be desired. The "Edit Name", "Edit Notes" and "Delete Job" buttons all return errors if you try to use them because the job is already deleted. However, I like the efficiency of re-using the job template, and the "Abort" button will make sense once we implement https://github.com/NVIDIA/DIGITS/issues/70#issuecomment-93782536.

lukeyeager:
Ok, with those last few fixes this looks good to me!

Please squash and merge.

PR #576, opened 2016-02-11
Drop error if a Caffe layer references an invalid bottom

gheinrich:
This should help address the case when a layer is added to manipulate e.g. the label blob. In that case the layer must include a directive like:

include {
 phase: TRAIN
}

Otherwise Caffe will drop an error during inference.

A sanity check is added to DIGITS to catch these errors earlier in the process (i.e. before training).

### DIFF
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -33,6 +33,15 @@ CAFFE_SNAPSHOT_PREFIX = 'snapshot'
 CAFFE_DEPLOY_FILE = 'deploy.prototxt'
 
 @subclass
+class Error(Exception):
+    pass
+
+@subclass
+class CaffeTrainSanityCheckError(Error):
+    """A sanity check failed"""
+    pass
+
+@subclass
 class CaffeTrainTask(TrainTask):
     """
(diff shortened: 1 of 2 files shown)


digits/model/tasks/caffe_train.py
```diff
@@ -1490,3 +1501,38 @@ def get_network_desc(self):
         return text description of model
         """
         return text_format.MessageToString(self.network)
+
+    def net_sanity_check(self, net, phase):
```
lukeyeager: Can we make this a `classmethod` so it's easy to write tests for?

gheinrich: Sure, or perhaps a `staticmethod` since we're not making any reference to the class?
And I need to write tests too, thanks for reminding me :-)

lukeyeager: Yeah you're right - I meant `staticmethod`.

gheinrich:
I have updated the patch to make the new method static and I have added tests.

lukeyeager:
> I have updated the patch to make the new method static and I have added tests.

Nice!

Issue #581, opened 2016-02-17
np.array performance bottleneck w/fix


anuragphadke:
there seems to be a performance issue while profiling example.py for batch of images. 
https://gist.github.com/anuragphadke/c0d8227ee4e27b9710c6
line #133:
caffe_images = np.array(caffe_images)
This has couple of issues:
1. it accounts for ~40% of execution time for a batch of images.
2. np.array() seems un-necessary as images are already loaded as np.array in load_image() call, line 104.

removing line #133 reduces execution time by 40%. 
performance metrics below:
current:
np.array took 2.009542 seconds.
Script took 6.449386 seconds.

proposed (after removing line #133)
np.array took 0 seconds.
Script took 4.233299 seconds.

Improvement: 
~2.2 seconds or about 40%

lukeyeager:
You're certainly right that the extra call to `np.array()` is unnecessary. Let's remove that.

FYI, the amount of wasted time isn't nearly as bad for me. Here are some numbers for GoogLeNet, batchsize=10:

Mode | Total time (sec) | Forward pass (sec) | `np.array()` (sec) | Wasted time
------------ | ------------- | ------------- | ------------- | -------------
GPU | 1.18 | 0.07 | 0.16 | 13%
CPU | 6.05 | 5.01 | 0.16 | 2.6%

Reference command:
```sh
head -n10 20-animals.txt | xargs \
  ./example.py model/snapshot_iter_33450.caffemodel model/deploy.prototxt -m model/mean.binaryproto  -l model/labels.txt  --batch-size 10 --nogpu
```

PR #582, opened 2016-02-17
Classification example np.array performance bottleneck

anuragphadke:
performance enhancement as per https://github.com/NVIDIA/DIGITS/issues/581


lukeyeager:
I must have assumed that Caffe wouldn't accept a `list` of `np.ndarrays`, but apparently it can.

LGTM. Thanks for the perf improvement!

PR #583, opened 2016-02-17
np.array performance bottleneck while training

anuragphadke:
https://github.com/NVIDIA/DIGITS/issues/581


lukeyeager:
See https://github.com/NVIDIA/DIGITS/pull/573#issuecomment-185357816 - we're refactoring all of that right now.

PR #307, opened 2015-09-14
Address protobuf PicklingErrors on all platforms

lukeyeager:
Close #306

PR #585, opened 2016-02-18
Upgrade caffe_train_task.network object type

lukeyeager:
*Caused by #307*

Convert from NetParameter to string back to NetParameter to avoid this error:

> TypeError: Parameter to MergeFrom() must be instance of same class:
expected caffe_pb2.NetParameter got caffe.proto.caffe_pb2.NetParameter.

/cc @ajtao

### DIFF
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -24,7 +24,7 @@ import caffe
 import caffe_pb2
 
 # NOTE: Increment this everytime the pickled object changes
-PICKLE_VERSION = 3
+PICKLE_VERSION = 4
 
 # Constants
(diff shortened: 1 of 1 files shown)


lukeyeager:
To reproduce: try fine-tuning a model that was created with DIGITS before #307.

lukeyeager:
@gheinrich can you try using this to upgrade some old jobs? I think it's correct but if something goes wrong with this sort of change it could potentially make someone lose all their old jobs.

gheinrich:
I tried to upgrade old jobs (created prior to #307) and I could use them to classify images.

lukeyeager:
Thanks for the review @gheinrich.

I've re-tested and still think this is the correct solution and a potentially useful fix. Plus, it will be a no-op unless people still have super old jobs lying around.

PR #586, opened 2016-02-18
Fix file serving for files with absolute paths

lukeyeager:
Bug reported at https://groups.google.com/d/msg/digits-users/MW_kHbOEqNo/y7wXpvlSFAAJ

To reproduce:
* Create a dataset from textfiles on the server (i.e. enter the path on the server manually)
* Click on the textfile from the DatasetJob page
* Get a 404 error because the path is wrong

### DIFF
--- a/digits/task.py
+++ b/digits/task.py
@@ -133,8 +133,8 @@ class Task(StatusCls):
             path = filename
         else:
             path = os.path.join(self.job_dir, filename)
-        if relative:
-            path = os.path.relpath(path, config_value('jobs_dir'))
+            if relative:
+                path = os.path.relpath(path, config_value('jobs_dir'))
         return str(path).replace("\\","/")
 
     def ready_to_queue(self):
--- a/digits/templates/helper.html
+++ b/digits/templates/helper.html
@@ -51,3 +51,12 @@
         {{ ' has-warning' if form.warnings else '' }}
     {% endfor %}
 {% endmacro %}
+
+{% macro serve_file(obj, path) %}
+    {% if path[0] == '/' %}
+        {# Absolute path - don't serve file for security reasons #}
+        {{path}}
+    {% else %}
+        <a href="{{url_for('digits.views.serve_file', path=obj.path(path, relative=True))}}">{{path}}</a>
+    {% endif %}
+{% endmacro %}
--- a/digits/job.py
+++ b/digits/job.py
@@ -123,8 +123,8 @@ class Job(StatusCls):
             path = filename
         else:
             path = os.path.join(self._dir, filename)
-        if relative:
-            path = os.path.relpath(path, config_value('jobs_dir'))
+            if relative:
+                path = os.path.relpath(path, config_value('jobs_dir'))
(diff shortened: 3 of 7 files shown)


gheinrich:
I don't see this issue on `master`, is this because I'm using a devserver and the source install?
For example when I click on a path like `/home/greg/ws/digits/digits/jobs/20160117-131355-ba71/train.txt` I seem to be correctly directed to `http://localhost:5000/files/20160117-131355-ba71/train.txt` and I can see the contents of the file.

lukeyeager:
That still works because the file you're using is in the jobs directory. Try one on your desktop, for example.

gheinrich:
Indeed! Moving the `.txt` files to `/tmp/text/*.txt`, now with your fix at 08f73e5 I don't get links to text files any more on the dataset job page. Is it expected?
job-local-file

lukeyeager:
> I don't get links to text files any more on the dataset job page. Is it expected?

That is what I meant to do, yes. But I admit it's a little confusing/disappointing.

The `serve_file` route only serves up files from within the jobs directory. That's for security and because it integrates with nginx more easily. See this comment:

```python
# https://github.com/NVIDIA/DIGITS/blob/v3.2.0/digits/views.py#L319-L328
def serve_file(path):
    """
    Return a file in the jobs directory
    If you install the nginx.site file, nginx will serve files instead
    and this path will never be used
    """
    jobs_dir = config_value('jobs_dir')
    return flask.send_from_directory(jobs_dir, path)
```

digits/task.py
```diff
@@ -133,8 +133,8 @@ def path(self, filename, relative=False):
             path = filename
         else:
             path = os.path.join(self.job_dir, filename)
-        if relative:
-            path = os.path.relpath(path, config_value('jobs_dir'))
+            if relative:
+                path = os.path.relpath(path, config_value('jobs_dir'))
```
gheinrich: unrelated with this change but that code duplication makes it look like `Task` and `Job` classes could share a common ancestor

digits/templates/helper.html
```diff
@@ -51,3 +51,12 @@
         {{ ' has-warning' if form.warnings else '' }}
     {% endfor %}
 {% endmacro %}
+
+{% macro serve_file(obj, path) %}
+    {% if path[0] == '/' %}
+        {# Absolute path - don't serve file for security reasons #}
+        {{path}}
+    {% else %}
```
gheinrich: Sorry for being picky... if the `.txt` file is in your DIGITS installation, say `$DIGITS_HOME/test/train.txt` and you specify the relative path to the file then a broken link will show on the dataset page (in addition to the fact that the job fails as the create DB script can't locate the relative text file). I don't think that is a major issue but should we not ban relative server paths?

lukeyeager: Good point. Filed a bug at https://github.com/NVIDIA/DIGITS/issues/593.

gheinrich:
Beside #593 the change looks good to me and is working for me.

PR #589, opened 2016-02-19
Update GettingStarted.md

ibmua:
Fixed


lukeyeager:
Thanks for the suggestion, but not everyone installs DIGITS with a deb package. Also, check out the line above the one you're editing.

ibmua:
But still you have to write "python " before the address, don't you?

lukeyeager:
Nope, that's what this line is for:
https://github.com/NVIDIA/DIGITS/blob/v3.0.0/tools/download_data/main.py#L1

PR #592, opened 2016-02-22
Use image limit value for "Classify/Test Many"

lukeyeager:
As requested, now the "Classify Many" and "Test Many" tools can use the image limit value.
* Use `num_test_images` for "Classify Many"
* Add `num_test_images` to generic path for "Test Many"
* Default to "All" instead of 100
* Remove a print debugging statement

---

highlight-image-limit

### DIFF
--- a/digits/model/images/classification/views.py
+++ b/digits/model/images/classification/views.py
@@ -353,6 +353,11 @@ def classify_many():
     else:
         image_folder = None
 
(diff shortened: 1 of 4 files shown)


lukeyeager:
I'll wait to merge this until I can rebase after @gheinrich merges #573, which has conflicting changes.

lukeyeager:
Afterwards, I want to fix the case where top_n > num_images. That's not handled well at the moment.

PR #594, opened 2016-02-23
Fix note editing permissions

gheinrich:


lukeyeager:
LGTM. It looks like the current job list on the homepage may need a tweak as well...

PR #596, opened 2016-02-23
Fix permissions for homepage Abort/Delete buttons

lukeyeager:


lukeyeager:
I had to add a strange fix (suggested in [this SO post](http://stackoverflow.com/a/9932189/2404152)) for this error:
```
ERROR: digits.test_scheduler.TestSchedulerFlow.test_add_remove_job
    RuntimeError: working outside of request context
```

Actually, this is required to run `digits/test_scheduler.py` by itself even without this PR. Otherwise, you get this error:
```
FAIL: digits.test_scheduler.TestScheduler.test_start_twice
    AssertionError: failed to stop
```

digits/templates/job_row.html
```diff
@@ -51,10 +51,12 @@ <h4 class="list-group-item-heading">
...
-            <a id="abort" class="btn btn-xs btn-warning" {{ ' style=display:none;' if not job.status.is_running() }} onClick="return abortJob('{{job.id()}}', '{{ abort_url }}');">Abort
-            </a>
-            <a id="delete" class="btn btn-xs btn-danger" {{ ' style=display:none;' if job.status.is_running() }} onClick="return deleteJob('{{job.id()}}');">Delete
-            </a>
+            {% if job | has_permission('abort') %}
+            <a id="abort" class="btn btn-xs btn-warning" {{ ' style=display:none;' if not job.status.is_running() }} onClick="return abortJob('{{job.id()}}', '{{ abort_url }}');">Abort</a>
+            {% endif %}
+            {% if job | has_permission('delete') %}
+            <a id="delete" class="btn btn-xs btn-danger" {{ ' style=display:none;' if job.status.is_running() }} onClick="return deleteJob('{{job.id()}}');">Delete</a>
+            {% endif %}
```
gheinrich: unrelated but why does the job have to not be running to display the delete button? On the job page there is no such limitation.

gheinrich:
Looks good to me and is working for me.

digits/templates/job_row.html
lukeyeager: Good question. Maybe it fits better? @jmancewicz?

jmancewicz: I think I had it that way, because initially jobs in that list were removed once they completed.  I was probably looking to avoid having Abort and Delete visible, and those conditions worked in the initial case.

lukeyeager: Ok, I'll change it in a separate PR.

PR #603, opened 2016-02-25
Deep copy in infer_many

### LINKED ISSUE #599: Test Many issue
The *Test Many* feature is not working properly for me when using generic inference.

It seems that I can't test more images than the batch size that I used for training. For example, if I use a batch size of 16 in the siamese network example from the docs and I try to test more than 16 images this is what I get:

screenshot from 2016-02-24 15 21 38

The result for the first four images is not correct. As you can see in the screenshot below, when I remove the last ten images the result for the first four images is correct.

screenshot from 2016-02-24 15 22 40

lukeyeager:
*Fix #599*

Fix a few problems:

1. Shallow copying - `infer_many` results are wrong for some images

  * Classification - https://github.com/NVIDIA/DIGITS/pull/361 snuck back in with https://github.com/NVIDIA/DIGITS/pull/573
  * Generic - has been wrong all along ([see here](https://github.com/NVIDIA/DIGITS/blob/f42f473710/digits/model/tasks/caffe_train.py#L1241))

2. ~~Output ordering - the wrong network output is displayed (e.g. for GoogLeNet)~~

  * ~~With https://github.com/NVIDIA/DIGITS/pull/573, network output order is lost and the wrong values are displayed~~

@gheinrich can you verify?

### DIFF
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2014-2016, NVIDIA CORPORATION.  All rights reserved.
 from __future__ import absolute_import
 
+import copy
 import math
 import operator
 import os
@@ -1258,7 +1259,7 @@ class CaffeTrainTask(TrainTask):
                         'data', image)
             o = net.forward()
             if outputs is None:
-                outputs = o
+                outputs = copy.deepcopy(o)
             else:
                 for name,blob in o.iteritems():
                     outputs[name] = np.vstack((outputs[name], blob))


digits/model/tasks/caffe_train.py
```diff
@@ -1256,15 +1257,16 @@ def infer_many_images(self, images, snapshot_epoch=None, gpu=None):
...
-                outputs = o
-            else:
-                for name,blob in o.iteritems():
-                    outputs[name] = np.vstack((outputs[name], blob))
-            print 'Processed %s/%s images' % (len(outputs[outputs.keys()[0]]), len(caffe_images))
+            outputs = net.forward()
+            for output_name in net.outputs:
+                blob = outputs[output_name]
+                if output_name in results:
+                    results[output_name] = np.vstack(results[output_name], blob)
```
gheinrich: aren't you missing a pair of brackets? Should it be:
`results[output_name] = np.vstack((results[output_name], blob))`

lukeyeager: Whoops, that was sloppy. Thanks for the careful review!

digits/model/tasks/caffe_train.py
```diff
@@ -1256,15 +1257,16 @@ def infer_many_images(self, images, snapshot_epoch=None, gpu=None):
...
-                outputs = o
-            else:
-                for name,blob in o.iteritems():
-                    outputs[name] = np.vstack((outputs[name], blob))
-            print 'Processed %s/%s images' % (len(outputs[outputs.keys()[0]]), len(caffe_images))
+            outputs = net.forward()
+            for output_name in net.outputs:
+                blob = outputs[output_name]
+                if output_name in results:
+                    results[output_name] = np.vstack((results[output_name], blob))
```
gheinrich: This is working for me, thanks!

Question: I was wondering why you didn't iterate with `for name,blob in net.forward().iteritems()`, which felt more concise and easy to read. I understand that `net.forward()` returns an unordered dictionary while `net.outputs` is an (ordered) array and this is why you need to iterate over `net.outputs`, right?

gheinrich:
I believe this change addresses the corrupt outputs on first batch but I don't think this addresses the ordering of outputs though... for one thing, does the `infer_one` not need to be changed as well?

Also, it feels like even before inference jobs were introduced Googlenet was working out of "luck"... on 5efec653 (just before #573) if you change the name of `loss1/classifier` into `z1` (any string greater than `"prob"` should do) then you get wrong results like:
googlenet-z1

With the inference jobs, since we're using HDF5 to transfer data from the inference process to the web app, I think we need some more logic to enforce the ordering of outputs.

lukeyeager:
> since we're using HDF5 to transfer data from the inference process to the web app, I think we need some more logic to enforce the ordering of outputs.

You're right. I'll just fix the deepcopy bug this this PR and we can handle the multiple outputs in another.

Issue #601, opened 2016-02-25
New sanity check for "invalid bottom" blobs breaks valid networks


Shadfield:
Revision d3cbdca5d4792e3a067094d6ce5e022279398826 introduced a new sanity check into digits/model/tasks/caffe_train.py at Line 1515 onwards.

I don't understand the code well enough to pinpoint the error further than this, but there seems to be some problem with the implementation of the sanity check.

Below is an MWE network definition. In previous revisions of digits this executes successfully, and plots out the performance on the train and validation (test) data. Obviously this MWE has nothing to learn, so the error would be constant for all epochs.

In the new revision of DIGITS the training process quits before it even starts, with the error message:
```
ERROR: Layer 'flatlabel' references bottom 'label' at the TEST stage however this blob is not included at that stage. Please consider using an include directive to limit the scope of this layer.

Traceback (most recent call last):
  File "/home/mofo/digits_new/digits/scheduler.py", line 482, in run_task
    task.run(resources)
  File "/home/mofo/digits_new/digits/task.py", line 179, in run
    self.before_run()
  File "/home/mofo/digits_new/digits/model/tasks/caffe_train.py", line 133, in before_run
    self.save_files_generic()
  File "/home/mofo/digits_new/digits/model/tasks/caffe_train.py", line 669, in save_files_generic
    CaffeTrainTask.net_sanity_check(deploy_network, caffe_pb2.TEST)
  File "/home/mofo/digits_new/digits/model/tasks/caffe_train.py", line 1451, in net_sanity_check
    layer.name, bottom, "TRAIN" if phase == caffe_pb2.TRAIN else "TEST"))
CaffeTrainSanityCheckError: Layer 'flatlabel' references bottom 'label' at the TEST stage however this blob is not included at that stage. Please consider using an include directive to limit the scope of this layer.
```
The sanity check is complaining that the flatten layer requires the label blob, even though the label blob is available in both train and test modes. I've also tried including 2 flatten layers (one specified as "phase: TEST" and the other as "phase: TRAIN") with no luck.

If we remove the flatten layer from the MWE (so label is connected straight to the loss layer) it passes the sanity check, and correctly computes the validation loss on the test data. So the label blob is definitely available at test time, even in the most recent DIGITS revision.

MWE network definition copied from the train_val.prototxt that DIGITS creates, with filepaths anonymized.
```
layer {
  name: "data"
  type: "Data"
  top: "data"
  include {
    phase: TRAIN
  }
[...]

PR #605, opened 2016-02-26
Layer exclusion naming convention applied to classification nets [DON'T MERGE YET]

gheinrich:
In case someone hits a problem like that mentioned in #601 for a classification network.
Once Caffe implements input layers and phase control from Python we should be able to remove those workarounds.

### DIFF
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -33,6 +33,9 @@ CAFFE_TRAIN_VAL_FILE = 'train_val.prototxt'
 CAFFE_SNAPSHOT_PREFIX = 'snapshot'
 CAFFE_DEPLOY_FILE = 'deploy.prototxt'
 
+TRAIN_VAL_LAYER_EXCLUSION_PREFIX = "deploy_"
+DEPLOY_LAYER_EXCLUSION_PREFIX = "train_"
+
 @subclass
 class Error(Exception):
     pass
@@ -243,7 +246,7 @@ class CaffeTrainTask(TrainTask):
         bottoms = {}
         train_data_layer = None
         val_data_layer = None
-        hidden_layers = caffe_pb2.NetParameter()
+        hidden_layers = []
         loss_layers = []
         accuracy_layers = []
         for layer in self.network.layer:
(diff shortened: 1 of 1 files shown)


digits/model/tasks/caffe_train.py
```diff
@@ -411,7 +411,10 @@ def save_files_classification(self):
                     val_data_layer.hdf5_data_param.batch_size = constants.DEFAULT_BATCH_SIZE
 
         # hidden layers
-        train_val_network.MergeFrom(hidden_layers)
+        for layer in hidden_layers:
+            # don't add layers with names starting with "deploy_" to train/val network
+            if not layer.name.startswith('deploy_'):
+                train_val_network.layer.add().CopyFrom(layer)
```
lukeyeager: In the `generic` path, I'm stripping the prefix from the final layer name:
https://github.com/NVIDIA/DIGITS/blob/v3.0.0/digits/model/tasks/caffe_train.py#L556

Can we do the same here?

lukeyeager:
I tried using this to integrate my Python accuracy layer ([here](https://gist.github.com/lukeyeager/5ae183044d96ef1540f2)), and got this error:

> ERROR: Layer 'prob' references bottom 'accuracy' at the TEST stage however this blob is not included at that stage. Please consider using an include directive to limit the scope of this layer.

That's because this layer gets added to the deploy network:
```
layer {
  name: "prob"
  type: "Softmax"
  bottom: "accuracy"
  top: "prob"
}
```
The bottom should be `ip2` instead of `accuracy`.

Looking at the code, I don't see a great way around this. I spent some time trying to find a programmatic solution to all this mess a while back ([rough code here](https://github.com/lukeyeager/DIGITS/compare/666e9f2c37ecd297abfccafb7ce2557e9f4cfca9...4dac32c32efdd6410a667ed40f3a72ac89cc8d67)) and I gave up on it because there was simply no way to do it elegantly. I'd love to get https://github.com/BVLC/caffe/issues/1245 integrated and update our default networks so that the networks can be fully, correctly specified without needing to do any magic behind the user's back.

gheinrich:
Updated to strip prefix but as @lukeyeager mentioned there is an issue with the helper code in DIGITS which adds a `softmax` layer to the deploy network.
Needs more work.

lukeyeager:
Closed in favor of https://github.com/NVIDIA/DIGITS/pull/623 (which has been closed in favor of https://github.com/NVIDIA/DIGITS/pull/628 !).

PR #606, opened 2016-02-26
Fix inference image paths in presence of invalid paths

gheinrich:
Bug introduced by #573: if a path from the list is invalid this leads to a misalignment between image paths and predictions, like:
test-many-one-invalid-path

To fix this the image paths of images that were successfully loaded must be returned by the inference tool.

### DIFF
--- a/digits/inference/tasks/inference.py
+++ b/digits/inference/tasks/inference.py
@@ -105,7 +105,6 @@ class InferenceTask(Task):
         super(InferenceTask, self).after_run()
 
         # retrieve inference data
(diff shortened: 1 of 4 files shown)


lukeyeager:
I think you need to address the ground truths, as well.

gheinrich:
Oh yes of course, thanks!

gheinrich:
I have pushed a new commit which should address the ground truth too.

lukeyeager:
Works for me, thanks!

PR #608, opened 2016-02-29
 	Show accuracy and confusion matrix in classify_many

gheinrich:
Show some stats when ground truth is provided in text file:
- top-1 accuracy,
- top-5 accuracy,
- per-class accuracy,
- confusion matrix.

This may produce output like:
alexnet-classify-many-stats

### DIFF
--- a/digits/inference/tasks/inference.py
+++ b/digits/inference/tasks/inference.py
@@ -105,7 +105,6 @@ class InferenceTask(Task):
         super(InferenceTask, self).after_run()
 
         # retrieve inference data
-        inputs = None
         visualizations = []
         outputs = {}
         if self.inference_data_filename is not None:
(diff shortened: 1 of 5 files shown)


lukeyeager:
Very nice!

If we merge, more people will probably be interested in using `infer_many` in DIGITS. That raises the importance of solving https://github.com/NVIDIA/DIGITS/issues/611 before the next release so that Nginx users don't try running big jobs which shut down their server when it takes too long to respond.

lukeyeager:
Tested with 1000-class models and the UI fails pretty gracefully. The table just shoots off the screen to the right, which seems like the only reasonable option.

gheinrich:
The current code discards empty rows. I was wondering whether it would make sense to discard empty columns as well. When most rows and columns have non-zero numbers in them I didn't really want to implement things like "retain only the x most common classes" because that would be another configuration knob to add.

Perhaps we can limit the maximum number of labels we show on the result page, and add an option to download the full matrix, _à la_ Travis log. That would be slightly more difficult to implement as I think that would require storing inference results on disk.

If user really feels offended by the wide result page, they can just not add the ground truth to their file and none of those stats will show.

lukeyeager:
I don't think any of those things need to be addressed right now. This looks good to me - merging!

PR #612, opened 2016-02-29
Show image index for infer_many

lukeyeager:
This mostly helpful when debugging shallow copy errors because it's easy to see when the first 16 images are wrong and everything else is right, for example. Hopefully with #603 those issues will go away and never come back, but this seems like it could still be useful.

Here's a screenshot of this feature used in conjunction with #608 (adds numbering to the bottom table):

infer-many-loop-index

### DIFF
--- a/digits/templates/models/images/classification/classify_many.html
+++ b/digits/templates/models/images/classification/classify_many.html
@@ -29,6 +29,7 @@
 
 <table class="table">
     <tr>
+        <td></td>
         <th>Path</th>
         {% if show_ground_truth %}<th>Ground truth</th>{% endif %}
         <th colspan=10>Top predictions</th>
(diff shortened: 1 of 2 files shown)


gheinrich:
Thanks, this looks good to me.

Question: I was wondering whether we should show the index of images from the list in the file instead/too? The indices will differ if there are images we failed to load. Also, shall we show the paths to those images (with an error message instead of a set of predictions)?

lukeyeager:
> I was wondering whether we should show the index of images from the list in the file instead/too? The indices will differ if there are images we failed to load. Also, shall we show the paths to those images (with an error message instead of a set of predictions)?

That's a good idea. Sounds like that would require passing even more data around from the ModelJob to the InferenceJob and then back again - right?

gheinrich:
> That's a good idea. Sounds like that would require passing even more data around from the ModelJob to the InferenceJob and then back again - right?

I don't think that would be the case because the inference job already passes the list of path IDs it was able to successfully load through the `inputs['ids']` list, like so:
https://github.com/NVIDIA/DIGITS/blob/master/digits/model/images/classification/views.py#L419

PR #526, opened 2016-01-19
Allow optional specification of image folder during multiple inference

gheinrich:
closes #504

PR #614, opened 2016-03-02
Fix ignored image folder in Top-N classification

gheinrich:
This was forgotten in #526


lukeyeager:
I've verified that this solution works. Thanks for the fix!

I'm restarting the Travis job - it took more than 45 minutes... :cry:

PR #624, opened 2016-03-10
Handle missing graph data gracefully

lukeyeager:
Fixes two errors when you have missing graph data:

1. `TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'`

2. `Uncaught ReferenceError: None is not defined`


lukeyeager:
We can't calculate mAP for Epoch 0 since there's not a snapshot yet.

gheinrich:
The code looks OK to me. I haven't used it for the mAP PR (https://github.com/lukeyeager/DIGITS/pull/1) as I added a separate network output type for asynchronous outputs.

PR #628, opened 2016-03-11
Use layer stages for all-in-one nets

### LINKED ISSUE #335: Automatically prune GoogleNet useless branches in the deploy

lukeyeager:
*Close #605, close #623, close #335*

* Use layer.include.stage to specify train/val/deploy all in a single .prototxt description
* Stop automatically creating Softmax layers in deploy networks for classification jobs
* Only set inner_product_param.num_output for classification jobs if it was unset
* Update the standard networks

TODO:

- [x] Documentation

### DIFF
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -231,78 +231,38 @@ class CaffeTrainTask(TrainTask):
         layer.transform_param.mean_file = mean_file
 
     # TODO merge these monolithic save_files functions
-    # TODO break them up into separate functions
     def save_files_classification(self):
         """
         Save solver, train_val and deploy files to disk
         """
-        has_val_set = self.dataset.val_db_task() is not None
-
-        ### Check what has been specified in self.network
-
-        tops = []
-        bottoms = {}
-        train_data_layer = None
-        val_data_layer = None
-        hidden_layers = caffe_pb2.NetParameter()
-        loss_layers = []
-        accuracy_layers = []
-        for layer in self.network.layer:
-            assert layer.type not in ['DummyData', 'ImageData', 'MemoryData', 'WindowData'], 'unsupported data layer type'
-            if layer.type in ['Data', 'HDF5Data']:
-                for rule in layer.include:
-                    if rule.phase == caffe_pb2.TRAIN:
-                        assert train_data_layer is None, 'cannot specify two train data layers'
-                        train_data_layer = layer
-                    elif rule.phase == caffe_pb2.TEST:
-                        assert val_data_layer is None, 'cannot specify two test data layers'
-                        val_data_layer = layer
-            elif layer.type == 'SoftmaxWithLoss':
-                loss_layers.append(layer)
-            elif layer.type == 'Accuracy':
-                addThis = True
-                if layer.accuracy_param.HasField('top_k'):
-                    if layer.accuracy_param.top_k >= len(self.get_labels()):
-                        self.logger.warning('Removing layer %s because top_k=%s while there are are only %s labels in this dataset' % (layer.name, layer.accuracy_param.top_k, len(self.get_labels())))
-                        addThis = False
-                if addThis:
-                    accuracy_layers.append(layer)
-            else:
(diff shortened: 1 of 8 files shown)


lukeyeager:
Added some documentation:

custom-network-explanation

digits/model/tasks/caffe_train.py
```diff
@@ -444,23 +401,22 @@ def save_files_classification(self):
...
 
         # network sanity checks
         self.logger.debug("Network sanity check - deploy")
         CaffeTrainTask.net_sanity_check(deploy_network, caffe_pb2.TEST)
+        found_softmax = False
+        for layer in deploy_network.layer:
+            if layer.type == 'Softmax':
+                found_softmax = True
+                break
+        assert found_softmax, 'Your deploy network is missing a Softmax layer! Read the documentation for custom networks and/or look at the standard networks for examples.'
```
gheinrich: Did I miss the bit in the documentation where it is explained to the user that a `softmax` layer is needed to display a probability distribution?

lukeyeager: Oh yeah, good call

lukeyeager: Updated.

digits/model/tasks/caffe_train.py
```diff
@@ -1472,3 +1415,178 @@ def net_sanity_check(net, phase):
...
+    for layer in network.layer:
+        if 'Data' in layer.type:
+            assert layer.type in ['Data', 'HDF5Data'], \
+                'Unsupported data layer type %s' % layer.type
+
+        if layer.type == 'Accuracy':
+            # Check to see if top_k > num_categories
+            if ( layer.accuracy_param.HasField('top_k') and
+                    layer.accuracy_param.top_k >= num_categories ):
+                self.logger.warning(
```
gheinrich: `self` is not defined here and the layer isn't actually being removed

gheinrich:
Great PR! Looks good except for the small omission in the processing of accuracy layers. I've also pushed #632 to update examples.

digits/model/tasks/caffe_train.py
lukeyeager: Whoops that was sloppy. Thanks for the review!

digits/templates/models/images/classification/custom_network_explanation.html
```diff
@@ -1,36 +1,107 @@
...
 </p>
 
+<p>
+    Classification are special in a few ways:
+</p>
 <ul>
     <li>
-        The <i>num_output</i> for each <b>InnerProduct</b> layer which is a network output gets set to the number of labels in the chosen dataset.
+        The Deploy network <b>must contain a Softmax layer</b>.
+        This should produce the only network output.
```
gheinrich: I think GoogleNet does not abide by this principle. Do the auxiliary classifiers need to be sent to `SilenceLayer`s in the `deploy` network?

lukeyeager: Actually it does. I pruned the auxiliary classifiers from the deploy network (solving #335).

gheinrich: Oh yes indeed, my mistake, sorry.

digits/model/tasks/caffe_train.py
lukeyeager: Oops I misspoke. The way I implemented it will break if two subsequent layers both have an invalid top_k because the second one won't be processed.

I need to fix that tomorrow...

```python
>>> a = range(10)
>>> for i, x in enumerate(a):
...   print 'Processing %d (%d) ...' % (i, x)
...   if (x%3 == 0):
...     del a[i]
...
Processing 0 (0) ...
Processing 1 (2) ...
Processing 2 (3) ...
Processing 3 (5) ...
Processing 4 (6) ...
Processing 5 (8) ...
Processing 6 (9) ...
>>> a
[1, 2, 4, 5, 7, 8]
```

PR #631, opened 2016-03-14
Support for non-persistent jobs

gheinrich:
~~Volatile jobs are not persisted to disk. Besides, they are automatically deleted by the scheduler 1 hour after completion.~~

These jobs are not saved to disk. Besides, they are automatically deleted by the scheduler 1 hour after completion.

This should become useful if jobs are increasingly used for background processing, inference, etc.

### DIFF
--- a/digits/inference/job.py
+++ b/digits/inference/job.py
@@ -20,7 +20,7 @@ class InferenceJob(Job):
         epoch   -- epoch of model snapshot to use
         layers  -- layers to import ('all' or 'none')
         """
-        super(InferenceJob, self).__init__(**kwargs)
+        super(InferenceJob, self).__init__(persistent = False, **kwargs)
 
         # get handle to framework object
         fw_id = model.train_task().framework_id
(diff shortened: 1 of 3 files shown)


lukeyeager:
I'm not sure that "volatile" is the right word here. I'd go with a word like "temporary", "one-off", "non-persistent", "transitory", etc.

How about `persistent=False` (defaults to `True`)?

gheinrich:
> How about persistent=False (defaults to True)?

Yes, that would work too, no problem. I used the word `volatile` in reference to `volatile memory` (i.e. that requires the server to be ON to retain its contents).

lukeyeager:
Ah, I was thinking of volatile as meaning "data may change unpredictably" rather than "data may be deleted at restart". Both are probably valid, but I think "persistent" is clearer, maybe? Thanks for the change.

gheinrich:
I agree with you - "persistent" is more self-explanatory - "volatile" has different meanings and is ambiguous here.

PR #566, opened 2016-02-03
Save job information to a database [DON'T MERGE]

lukeyeager:
### Motivation

* Allows communication between server and workers on different nodes (coming soon!)
* May enable multi-threaded webserver (https://github.com/NVIDIA/DIGITS/issues/479)
  * Need to investigate compatibility with Flask-SocketIO
* Faster startup/shutdown times
* Browse/edit job data with SQL tools

### Features

* You can use SQLite, MySQL, PostgreSQL or anything else [supported by SQLAlchemy](http://docs.sqlalchemy.org/en/rel_0_8/dialects/index.html)
* Database schema changes are handled automatically through [Flask-Migrate](https://github.com/miguelgrinberg/Flask-Migrate)
* New CLI using [Flask-Script](https://github.com/smurfix/flask-script)
  * Start development server

    ```sh
$ ./manage.py runserver --help
usage: manage.py runserver [-h] [-t HOST] [-p PORT] [-d] [-r]

Runs the Flask-SocketIO development server

optional arguments:
  -h, --help            show this help message and exit
  -t HOST, --host HOST
  -p PORT, --port PORT
  -d, --no-debug
  -r, --reload
    ```
  * Create / upgrade the database

    ```sh
./manage.py db init
./manage.py db migrate
./manage.py db upgrade
    ```

### Progress

- [x] Define db schema
- [x] Load data into db from previous jobs (pickle files)
- [ ] Store data for new jobs to db
- [ ] Read job data from db
- [ ] Stop loading pickle files
- [ ] Stop saving pickle files
- [ ] Set up cascading deletion for appropriate relationships
- [x] Add configuration option for database dialect
- [ ] Document how to create and connect to a database

PR #633, opened 2016-03-15
refreshing digits home view will unload deleted and load newly added …

peterneher:
- scheduler not only loads jobs from disk but also removes jobs from its internal list that are not present on the disk anymore

- home view calls scheduler.load_past_jobs() on refresh

### DIFF
--- a/digits/scheduler.py
+++ b/digits/scheduler.py
@@ -108,14 +108,15 @@ class Scheduler:
 
         self.running = False
         self.shutdown = gevent.event.Event()
-
+        
     def load_past_jobs(self):
         """
         Look in the jobs directory and load all valid jobs
         """
         loaded_jobs = []
         failed_jobs = []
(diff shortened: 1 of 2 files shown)


lukeyeager:
Sorry but I'm not going to merge this. When you have hundreds of jobs on your machine, calling `scheduler.load_past_jobs()` can take several seconds. And most people don't delete job directories manually.

The right way to fix this is with something like https://github.com/NVIDIA/DIGITS/pull/566. I've had to push that off for a while behind other priorities, but that's the way we'll address the issue going forward.

Thanks for the PR, though! Even though I can't merge it into `master`, this could still be a useful workaround for others.

peterneher:
Sure it will take long? The jobs that are already loaded are not touched,
so most of the time nothing is loaded if the method is called a second
time. Only new ones are loaded.
Anyway. Happy to help :)

2016-03-15 17:46 GMT+01:00 Luke Yeager <notifications@github.com>:

PR #638, opened 2016-03-16
Support for inference on LMDB

### LINKED ISSUE #619: Support for inference on LMDB file
Original suggestion from Clement Payout.

It could be useful to allow users to provide their own LMDB to perform inference on:
- That would be one possible use of the test LMDB that DIGITS can create when generating a classification dataset;
- That would make it possible to work with image-like data that cannot be stored in image files (e.g. non-standard number of channels, ...). Today a user can create a generic model for those image-like data but DIGITS won't allow them to perform inference if the data are not stored as image files;
- This would allow users to skip the image file creation step.

### LINKED ISSUE #630: Optionally keep the information about the original image in LMDB
We've recently had an issue, we experimented with image similarity and instead returning similar images from lmdb dabatase (resized, croped and filled with noise), we wanted to return the original image, but key or value in lmdb database doesn't contain such information.

Now, we could hack the create_db.py file but we don't find this a good long term solution, so an option is to completely abandon digits or continue using it if such option becomes available.

The suggestion is to optionally provide a third column to input txt files provided to digits to fill the database, where third column would be image id. So for example key would look something like this:
b'00000234_992929_4' where 992929 would be some id provided by user, and 4 the category id. The third column should be optional for easier usage for users who do not need the info about original image.

Another option is that filepath is part of a key.

If you decide to change the key naming convention, also think about changing the first number to 9 digits so datasets of more than 100M images have valid keys, currently this looks like this (line 643 in create_db.py):

```key = '%08d_%d' % (image_count + i, datum.label)```

### PR DESCRIPTION
**Pull request by gheinrich**
A (small) step towards support for non-image data: this makes it possible to perform inference on any 3D blob. Blobs may not be "proper" images (e.g. because they don't have a standard number of channels).

close #619
close #630 

A new field and a new button were added to allow inference on LMDB database:

generic-inference-form

Results are shown as follows (using DB key instead of image path):

infer_db

# Progress #

- [x] Add "Test DB" form to generic model show page
- [x] Create new route
- [x] Update inference job to take in a path to a database
- [x] Update inference tool to parse an LMDB database
- [x] Add tests

### DIFF
--- a/digits/inference/tasks/inference.py
+++ b/digits/inference/tasks/inference.py
@@ -27,7 +27,7 @@ class InferenceTask(Task):
         """
         Arguments:
         model  -- trained model to perform inference on
-        images -- list of images to perform inference on
+        images -- list of images to perform inference on, or path to a database
         epoch  -- model snapshot to use
         layers -- which layers to visualize (by default only the activations of the last layer)
         """
@@ -72,11 +72,12 @@ class InferenceTask(Task):
         super(InferenceTask, self).before_run()
         # create log file
         self.inference_log = open(self.path(self.inference_log_file), 'a')
-        # create a file to pass the list of images to perform inference on
-        imglist_handle, self.image_list_path = tempfile.mkstemp(dir=self.job_dir, suffix='.txt')
-        for image_path in self.images:
-            os.write(imglist_handle, "%s\n" % image_path)
-        os.close(imglist_handle)
+        if type(self.images) is list:
+            # create a file to pass the list of images to perform inference on
+            imglist_handle, self.image_list_path = tempfile.mkstemp(dir=self.job_dir, suffix='.txt')
+            for image_path in self.images:
+                os.write(imglist_handle, "%s\n" % image_path)
+            os.close(imglist_handle)
 
     @override
     def process_output(self, line):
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by gheinrich** [comment]
Tests added in latest commit.

**Comment by ajcampisi** [comment]
I am interested in using Digits with my Multispectral images (4 image channels).  Since most multispectral images can be output as a separate 8 bit grayscale images (one for each channel), I was wondering if we could get Digits to accept each  multispectral images as separate channels images for both training and classification. For example the file-list could contain the image name, channel number and category.
Image1.jpg c1 1
Image1.jpg c2 1
Image1.jpg c3 1
Image1.jpg c4 1
Image2.jpg c1 1
Image2.jpg c2 1
Image2.jpg c3 1
Image2.jpg c4 1
Image3.jpg c1 2
Image3.jpg c2 2
Image3.jpg c3 2
Image3.jpg c4 2

**Comment by gheinrich** [comment]
Hi @ajcampisi DIGITS does not support the format you propose out of the box. Are you saying that each image channel - taken in isolation - is sufficient to classify the image? In that case why don't you convert the 4-channel image to grayscale with:
```
convert <input> -colorspace Gray <output>
```
and then feed those grayscale images to DIGITS?

ajcampisi:
Hi Greg,
Thanks for the reply.  No unfortunately it is not sufficient an individual channel image to be used for classification.  All four channels images need to be feed into the network as four separate network inputs and the combined image data from all four network channels can then be used by the network for both classification and training.

PR #651, opened 2016-03-22
Add pop-out explanation for Python layers

### LINKED ISSUE #636: Add better Python layer documentation
The documentation from https://github.com/NVIDIA/DIGITS/pull/447 is great and all, but we should have better documentation readily available within the app, like https://github.com/NVIDIA/DIGITS/pull/628#issuecomment-195137527.

lukeyeager:
*Close #636*

Cleanup Python layer upload UI and add some pop-out text similar to https://github.com/NVIDIA/DIGITS/pull/628#issuecomment-195137527.

Also add network outputs with names containing "accuracy" to the accuracy axis in graphs. It's a little hacky, but I don't think we need to find a general purpose solution to that problem just yet.

### DIFF
--- a/digits/templates/models/python_layer_explanation.html
+++ b/digits/templates/models/python_layer_explanation.html
@@ -1,0 +1,75 @@
+{# Copyright (c) 2015-2016, NVIDIA CORPORATION.  All rights reserved. #}
+
+<h1>Using Python layers</h1>
+
+<p>
+    Caffe lets you define custom layers with Python.
+    Python layers are obviously much slower than their C++ counterparts, but are easy to write and test quickly since using them doesn't require recompilation.
+    DIGITS moves some files around for you so that you can specify different Python layers for each network.
+</p>
+<p>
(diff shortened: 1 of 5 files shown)


digits/templates/models/python_layer_explanation.html
```diff
@@ -0,0 +1,75 @@
+{# Copyright (c) 2015-2016, NVIDIA CORPORATION.  All rights reserved. #}
+
+<h1>Using Python layers</h1>
+
+<p>
+    Caffe lets you can define custom layers with Python.
```
gheinrich: small typo here

gheinrich:
This is working for me (I tried both the "blank square" and "accuracy" layers examples on MNIST+LeNet).

digits/templates/models/python_layer_explanation.html
lukeyeager: Haha whoops - right at the beginning! Thanks.

lukeyeager:
Thanks for the review, Greg. Here's a screenshot of the feature for those who are interested:

python-layer-explanation

PR #655, opened 2016-03-24
Don't sanity check val phase if val set not present

### LINKED ISSUE #654: Cannot train GoogLenet without a validation dataset.
This issue seems to happen with the current master-branch.
I am doing some experiments where I do not have access to a validation dataset. 
On previous digits-version I Could easily train my network on DIGITS whenever it only had the training dataset.

The network is the GoogLenet that comes with digits.
Now, if i create a LMDB though digits (text file specification), that only has training dataset I receive the following.

`CaffeTrainSanityCheckError: Layer 'conv1/7x7_s2' references bottom 'data' at the TEST stage however this blob is not included at that stage. Please consider using an include directive to limit the scope of this layer.
`

This seems related to #601 . But I could not figure out a way to fix it.
Any suggestions ?

gheinrich:
close #654

### DIFF
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -382,8 +382,9 @@ class CaffeTrainTask(TrainTask):
         # network sanity checks
         self.logger.debug("Network sanity check - train")
         CaffeTrainTask.net_sanity_check(train_val_network, caffe_pb2.TRAIN)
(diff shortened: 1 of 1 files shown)


lukeyeager:
Oops, that's probably coming from https://github.com/NVIDIA/DIGITS/pull/622 - my bad!

PR #657, opened 2016-03-25
Allow autocomplete for testing image paths

lukeyeager:
Before this PR, you could manually enter a path to an image on the server for testing, but the image would be deleted afterwards!

Now you get autocomplete help for entering paths, and the image won't be deleted afterwards.


gheinrich:
Looks good to me! I checked the classification and generic paths.

PR #658, opened 2016-03-25
Device query updates

lukeyeager:
**Versioned DSOs**
We really shouldn't ever load unversioned DSOs because you shouldn't need dev files to run any application. I ran into this issue when using `nvidia-docker`, which doesn't include `libnvidia-ml.so` by default.

**Displaying GPU memory**
We could just display CUDA info only - there's no clear reason to call NVML here - but I have a vague recollection of CUDA being unable to query the memory available for newer, fancier GPUs sometimes. So I'm just adding it as a fallback path.

I can't test this on Windows, but I've attempted to preserve @crohkohl's changes from #199.

### DIFF
--- a/digits/device_query.py
+++ b/digits/device_query.py
@@ -109,27 +109,52 @@ def get_library(name):
     Returns a ctypes.CDLL or None
     """
     try:
-        if platform.system() == 'Linux':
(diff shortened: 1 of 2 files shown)


gheinrich:
Verified with/without this commit and the code looks OK to me.

Issue #667, opened 2016-04-04
'Related Jobs' window feature


TimZaman:
It's too bad #137 was abandoned. It was a huge amount of half-working features that i am trying to split out and work on one by one (the which i want to add). The related jobs view is very nice to me as i can see which datasets i can throw away when they are not used, and i can jump around easily between jobs using the name dataset.
Therefore the feature request to implement this, which i have already done. i will file a pull request soon. 
<img width="1200" alt="8e24a58e-1432-11e5-9808-6acd7e63bb80" src="https://cloud.githubusercontent.com/assets/7721540/14253313/517c3cf6-fa8b-11e5-8db7-89900c04c1d0.png">

lukeyeager:
Why close this? I agree with you - I'd like to see this feature added soon.

TimZaman:
Ah sorry Luke. I have this feature implemented in my own fork, but i worked very fast so I didnt put it in a nice branch to make a decent pull request. Actually, this feature is the one i miss most (from my own fork) when i use the vanilla Nvidia-master branch.

lukeyeager:
Hey @TimZaman are you working on this already? I see some commits referencing this issue. If not, @Lucaszw can do it.

TimZaman:
It's been in our fork for ages, and I have no clue how you guy can live without it. How about the Confusion Matrix? That's just so damn important to us (also in our fork, but only for torch models): sometimes your accuracy is 90%, and only the confusion matrix will show it's just 100% biased towards a specific label (due to dataset bias).

Example: 77% accuracy in graph.. but meanwhile the confusion matrix tells me:
![image](https://cloud.githubusercontent.com/assets/7721540/15441660/6a2bc9c2-1edc-11e6-8ac1-8b0b0c89535e.png)

lukeyeager:
> i will commit the Related Jobs next week

Cool, thanks!

> How about the Confusion Matrix?

Well we do have this: https://github.com/NVIDIA/DIGITS/pull/608.

But it looks like you're talking about something different - storing all the validation results and viewing *those* as a confusion matrix. Cool idea! I think that might be hard to do for Caffe...

TimZaman:
Torch generally outputs the confusion matrix during validation, i set up a
simple regex that i catch and that's saved in the pickle file an updated
and sent with socketio with validation pass and voila. And i dont output
the confmat when there are more than 100 labels or it gets hefty. I didnt
[...]

PR #137, opened 2015-06-08
Accuracy & confusion matrix

groar:
See #17 

Adds a new kind of job for performance evaluation of trained classifiers. It is now possible to visualize :
* accuracy / recall curve
* confusion matrix

Accuracy and the confusion matrix are computed against a chosen snapshot of a training task, and against both the **validation set and testing set** (if it exists). An "evaluate performance" button has been added on the training view. This is currently the only way to run an evaluation job. The results are stored in the job directory in the form of two pickle files. 

button

## Accuracy / recall curve

accuracy recall curve

## Confusion matrix

I chose a very simple representation of the confusion matrix (not in the form of a matrix !), because it is more adapted to datasets with lots of classes. For each class, the top 10 most represented classes are displayed, with their respective %.   

confusion matrix

## Related jobs

I added a "Related jobs" section on each job show view. It displays the jobs which depends on the current job. For example, models trained on a specific dataset, evaluations ran on a specific model.

Related jobs

Let me know what you think, critiques and comments are more than welcome.

PR #668, opened 2016-04-04
Implemented #667 related jobs (once part of #137)

TimZaman:
See title.
related_jobs

### DIFF
--- a/digits/job.py
+++ b/digits/job.py
@@ -266,3 +266,10 @@ class Job(StatusCls):
         Returns False if this job can be edited
         """
         return False
+
+    def parent_jobs(self):
+        """
+        Returns the list of the jobs on which the job
+        depends
+        """
+        raise NotImplementedError('Implement me!') #@TODO
--- a/digits/scheduler.py
+++ b/digits/scheduler.py
@@ -208,6 +208,20 @@ class Scheduler:
             return None
         return self.jobs.get(job_id, None)
 
+    def get_related_jobs(self, job):
+        """
+        Look through self.jobs to try to find the Jobs
+        whose parent contains job
+        """
+        related_jobs = []
+        for j in self.jobs:
+            try:
+                if job in self.jobs[j].parent_jobs():
+                    related_jobs.append(self.jobs[j])
+            except NotImplementedError: 
+                pass
+        return related_jobs
+
     def abort_job(self, job_id):
         """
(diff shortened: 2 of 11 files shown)


digits/job.py
```diff
@@ -266,3 +266,10 @@ def is_read_only(self):
         Returns False if this job can be edited
         """
         return False
+
+    def parent_jobs(self):
+        """
+        Returns the list of the jobs on which the job
+        depends
+        """
+        raise NotImplementedError('Implement me!') #@TODO
```
gheinrich: you could return an empty list to avoid having to catch the exception in the caller?

digits/scheduler.py
```diff
@@ -208,6 +208,20 @@ def get_job(self, job_id):
             return None
         return self.jobs.get(job_id, None)
 
+    def get_related_jobs(self, job):
+        """
+        Look through self.jobs to try to find the Jobs
+        whose parent contains job
+        """
+        related_jobs = []
```
gheinrich: I think this doesn't return anything for model jobs. Should you initialize as `related_jobs = job.parents_job()` to capture the dataset associated with model jobs?

gheinrich:
Thanks for submitting this PR! Can you define what is meant by "related" job?

TimZaman:
In the dataset view, it shows you the associated models that ran on the dataset..

gheinrich:
You made changes in the model view too though there doesn't seem to be a way to show any related jobs on those? I thought you would show models that use the same dataset but that isn't how the code implements this. Would you like to revert your changes in the model files?

Also, if you haven't done so already, would you mind signing the [CLA](https://github.com/gheinrich/DIGITS/blob/master/README.md#bugs-and-feature-requests). Thanks!

digits/job.py
TimZaman: I was just following the repository's convention. But i agree the exceptions are quite dirty, will change.
https://github.com/NVIDIA/DIGITS/blob/master/digits/job.py#L152-L157

digits/scheduler.py
TimZaman: Correct. Will implement that too.

TimZaman:
I have now made so many improvements upon digits that this one's already outdated. Gonna close it this pull request. Might open another one if i get to it.

PR #684, opened 2016-04-15
Add text classification example

gheinrich:


TimZaman:
Hi Greg. This is awesome. Read the paper earlier as well and thought it was awesome. Thanks man, this is very cool!
Tried the regular approach (now busy on the alternative one).
I double checked and did exactly as was outlined. I do get an error.
The error is

`/usr/share/lua/5.1/cudnn/TemporalConvolution.lua:92: bad argument #1 to 'size' (out of range)`

And the verbose output of torch is


```
Last output:
Running initial validation before first train epoch..
Validation (epoch 0): loss = 3.2712374114555, accuracy = 0.071214285714286
Training (epoch 0.0005): loss = 4.3866891860962, lr = 0.01
/usr/share/lua/5.1/cudnn/TemporalConvolution.lua:92: bad argument #1 to 'size' (out of range)

```

Weird. I might try recreating the dataset to make absolutely sure i made it correctly. However, did seem to pass the Validation epoch (7.12% accuracy epoch 0) which tells me that the model and the data are probably legit.

gheinrich:
Hi @TimZaman thanks for the feedback! I used `cudnn.TemporalConvolution` to train the network (I found it to be 17 times faster than the regular `nn.TemporalConvolution`! Just to check: did you remember to set mean subtraction to `none` when you created the classification model?

gheinrich:
hi @TimZaman are you still having an issue with this example?

TimZaman:
Hi @gheinrich, i did not use mean substraction as i was supposed to.
I have last week swapped the (bit) outdated torch-nv to the normal torch installation. This did the trick. Other than the easy install of torch-nv, are there any advantages in using torch-nv wrt the build from source?

lukeyeager:
I tested the dataset creation and import. When I tried to test the model, I got this error:

    [FAIL] ...5-02/install/share/lua/5.1/cudnn/TemporalConvolution.lua:92: bad argument #1 to 'size' (out of range)

Maybe it's because I don't have this commit: https://github.com/soumith/cudnn.torch/pull/152? I had to start with a non-standard fork ([16.04 issues](https://github.com/torch/torch7/issues/670#issuecomment-216336538)), so I'm having trouble updating just the `cudnn` module. Sorry for my noobery! I'll find some time to figure out the Torch build system and keep testing this, but it looks good so far.

gheinrich:
Oh so it's the same error @TimZaman got. You could update just `cudnn` by doing something like:
```sh
$ cd .../torch/extra/cudnn
$ git fetch origin && git checkout origin/master
$ luarocks make cudnn-scm-1.rockspec 
```

gheinrich:
> Other than the easy install of torch-nv, are there any advantages in using torch-nv wrt the build from source?

I don't know of other advantages...

TimZaman:
Yeah so the cudnn stuff is not DIGITS related. @gheinrich I would recommend just leaving out the cudnn, switch to nn, albeit 17x slower (worked fine with me, but maybe make a note about cudnn being faster but you'd have to update) and just merge.

lukeyeager:
Got it working. Cool tutorial!

text-classification

PR #686, opened 2016-04-16
Allocate LMDB keys in tcs.Vec object

### LINKED ISSUE #659: Torch + LMDB, not enough memory
Hello, I am trying to use lmdb dataset that I created with DIGITS and torch defined model. The dataset has around 300k objects. The loading is interrupted by the error: 

```
2016-03-26 09:30:27 [INFO ] creating data readers
2016-03-26 09:30:27 [INFO ] opening LMDB database: /place/home/dmitry/git/DIGITS/digits/jobs/20160325-003220-9caf/train_db
2016-03-26 09:30:27 [INFO ] opening LMDB database: /place/home/dmitry/git/DIGITS/digits/jobs/20160325-003220-9caf/train_db
2016-03-26 09:30:27 [INFO ] opening LMDB database: /place/home/dmitry/git/DIGITS/digits/jobs/20160325-003220-9caf/train_db
2016-03-26 09:30:27 [INFO ] opening LMDB database: /place/home/dmitry/git/DIGITS/digits/jobs/20160325-003220-9caf/train_db
2016-03-26 09:32:08 [FAIL] /home/dmitry/torch/install/share/lua/5.1/threads/threads.lua:264:
[thread 4 callback] /place/home/dmitry/git/DIGITS/tools/torch/data.lua:326: not enough memory
```

I have tried to install different versions of LMDB lib, but it does not help. Thank you in advance.

gheinrich:
Luajit heap size must be located in the lower 2GB of addressable memory.
There is therefore a limit of 2GB on heap size when using Luajit.
This change moves LMDB keys to a tds.Vec object, where memory does not rely on the Lua allocator.

close #659

### DIFF
--- a/docs/BuildTorch.md
+++ b/docs/BuildTorch.md
@@ -63,6 +63,7 @@ Install extra luarocks packages:
 ```sh
 % luarocks install image
 % luarocks install "https://raw.github.com/deepmind/torch-hdf5/master/hdf5-0-0.rockspec"
+% luarocks install tds
 ```
 
 ## LMDB support
--- a/scripts/travis/install-torch.sh
+++ b/scripts/travis/install-torch.sh
@@ -21,6 +21,7 @@ cd $INSTALL_DIR; ./install.sh -b
 # install custom packages
 ${INSTALL_DIR}/install/bin/luarocks install sys
 ${INSTALL_DIR}/install/bin/luarocks install image
+${INSTALL_DIR}/install/bin/luarocks install tds
 ${INSTALL_DIR}/install/bin/luarocks install "https://raw.github.com/deepmind/torch-hdf5/master/hdf5-0-0.rockspec"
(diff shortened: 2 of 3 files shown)


TimZaman:
Don't pull this in yet, see #687 ; for me luajit gobbles up all the cpu resources it can for no obvious reason.

gheinrich:
From https://github.com/NVIDIA/DIGITS/issues/687#issuecomment-211118777

> you have definitely solved the memory issue by using tds, that's for sure, and that one can be merged i think.

lukeyeager:
Hmm, this breaks compatibility with the latest packaged version of Torch. Is there any way to fall back to a normal array in case `tds` isn't loaded?

TimZaman:
On the other hand, it shouldn't be difficult, it's just a one-liner really:
`local Keys = tds.Vec()`
should be ye old
local Keys = {}
`when tds isnt loaded`

PR #690, opened 2016-04-18
Select framework on model cloning

### LINKED ISSUE #688: Cloning a vanilla Torch Model selects a vanilla Caffe Model but runs the Torch one.
See title. You can replicate by:
Click new model. Select Torch->Alexnet.
Make the job.
Press clone.
You will see that the model selector is hovering now on Caffe->Alexnet in the Standard Network tab.
This last fact is just a UI remnant. If you would press 'Create Model' now, it will create a torch Alexnet model, although Caffe was selected!

If one were to switch to the torch alexnet standard network, and then back to the caffe alexnet standard network, one would get indeed get the Caffe one.
Seems like a straightforward UI bug.

gheinrich:
close #688


lukeyeager:
Verified bug and fix.

PR #692, opened 2016-04-19
Default to Lua table if tds not installed

gheinrich:
If tds is not installed because e.g. Torch was installed with torch-nv deb package then default to using a regular Lua table to store LMDB keys.
There is a limit of 2GB for heap memory in Luajit so we might not be able to train on some large datasets without tds.


lukeyeager:
(Follow up from https://github.com/NVIDIA/DIGITS/pull/686)

lukeyeager:
I'm not set up to test this at the moment, but it looks pretty safe to me.

PR #695, opened 2016-04-20
Remove unnecessary installs

gheinrich:
`sys` and `image` modules are automatically installed as part of the default Torch7 install.


lukeyeager:
So should we remove `image` from this list, too?
https://github.com/NVIDIA/DIGITS/blob/v3.2.0/docs/BuildTorch.md#luarocks-dependencies

gheinrich:
Yes indeed. Thanks, I have updated the patch.

Issue #693, opened 2016-04-19
Non-Square image problems


TimZaman:
Working with non-square images. Did any of you try for this?
I cannot train because on the initial main.lua Validation() run it crashes with **torch error code -11**. Probably a fat memory or access issue?

Let's say my dataset is 30x1. Creation of the lmdb is fine.
Then torch output log tells me 
> Image channels are 1, Image width is 1 and Image height is 30

So that's not right, it said 30x1, and by universal and digits convention it should indeed be width*height.
Then in the data.lua source code my eye falls on https://github.com/NVIDIA/DIGITS/blob/master/tools/torch/data.lua#L281
`logmessage.display(0,'Image channels are ' .. self.ImageChannels .. ', Image width is ' .. self.ImageSizeY .. ' and Image height is ' .. self.ImageSizeX)`

Huh? width is Y and height is X? What's going on here? We need to revision that. 

This got so weird that I decided to recreate the db. I made the not-working one with PNG+squash. Then i decided to do this with None Image Encoding+crop (although i did not crop anything). That last fact made it work! Yay!
Now then i decided to go back to the default 'squash'+None Image Encoding. And that worked. So something's up with nonsquare PNG's creation or loading.
To make things even weirder, if i resized my 30x1 input to 1x30, the model could run! This means it could load PNG's this time. I think the squashing-recomputation fixed something there.

Also, in all cases the dataset browsing works fine. Can anyone help me debug this (on the data.lua end)?

_Sidenote: why are we allocating 2x the amount of data per image: https://github.com/NVIDIA/DIGITS/blob/master/tools/torch/data.lua#L351 something to do with jpegs?_

gheinrich:
Hi, thanks for the report.

> Working with non-square images. Did any of you try for this?

This is tested as part of the automatic Travis [tests](https://github.com/NVIDIA/DIGITS/blob/v3.2.0/digits/model/images/classification/test_views.py#L929-L933) on every commit. There is also an example [there](https://github.com/NVIDIA/DIGITS/pull/684). Of course we can't exclude the possibility of a bug lurking somewhere.

> by universal and digits convention it should indeed be width*height

[...]

PR #697, opened 2016-04-20
Image dimension hint

gheinrich:
The dataset creation form does not explicitly state which of the image dimension fields is the width and which is the height.
Though this can be inferred through `See example` it would probably help to make this explicit.
See #693


TimZaman:
Don't forget the log message display:
https://github.com/NVIDIA/DIGITS/blob/master/tools/torch/data.lua#L281

lukeyeager:
Sure, can't hurt.

PR #701, opened 2016-04-22
Dump full model when creating Torch model snapshot

### LINKED ISSUE #496: Torch SpatialBatchNormalization layer cannot be loaded from weights
During the training a snapshot of model is saved for later use. It consists of model script, weights snapshot, mean data, and labels file.

However, if model contains SpatialBatchNormalization layer that's not enough. The layer contains fields **running_mean** and **running_std** which are calculated from training data but later lost during save/load snapshot procedure.

Maybe saving a sanitized model (with cleared intermediate states) instead of weights would be a good idea?

gheinrich:
close #496

### DIFF
--- a/tools/torch/main.lua
+++ b/tools/torch/main.lua
@@ -351,6 +351,24 @@ logmessage.display(0,'Network definition: \n' .. model:__tostring())
 
 logmessage.display(0,'Network definition ends')
 
+local function saveModel(model, directory, prefix, epoch)
+    local filename
+    local modelObjectToSave
+    if model.clearState then
+        -- save the full model
+        filename = paths.concat(directory, prefix .. '_' .. epoch .. '_Model.t7')
+        modelObjectToSave = model:clearState()
+    else
+        -- this version of Torch doesn't support clearing the model state => save only the weights
+        local Weights,Gradients = model:getParameters()
+        filename = paths.concat(directory, prefix .. '_' .. epoch .. '_Weights.t7')
+        modelObjectToSave = Weights
+    end
+    logmessage.display(0,'Snapshotting to ' .. filename)
+    torch.save(filename, modelObjectToSave)
+    logmessage.display(0,'Snapshot saved - ' .. filename)
+end
+
 local function switchType(newType, model, loss)
     if newType == 'float' then
         logmessage.display(0,'switching to floats')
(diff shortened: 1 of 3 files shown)


gheinrich:
Waiting for feedback before merging

This commit changes the way we create model snapshots: instead of saving just the learnable parameters we are now saving the full model. This is required to preserve parameters such as those computed during batch normalization.

This change is backward compatible: old model snapshots will be loaded too.

Downside: this makes it difficult to deploy a CuDNN-trained model on a non-CuDNN system since the CuDNN layers need to be explicitly converted to NN in a fine-tuning hook.

Besides, snapshots are now slightly bigger than they used to be (up to a factor of 2).

pansk:
I confirm this solved my problem with BatchNormalization & SpatialBatchNormalization

gheinrich:
as @pansk mentioned, `clearState()` isn't yet part of the module API in `torch7-nv.deb`. I'll have to find a way to deal with this before we can merge this commit.

tools/torch/main.lua
```diff
@@ -560,46 +567,6 @@ while tmp_batchsize <= trainSize do
...
-    logmessage.display(0,'random number state saved - ' .. filename)
-
-    --save lrPolicy state
-    filename = paths.concat(opt.save, 'lrpolicy_' .. backup_epoch .. '.t7')
-    logmessage.display(0,'lrpolicy state saving to ' .. filename)
-    torch.save(filename, optimizer.lrPolicy)
-    logmessage.display(0,'lrpolicy state saved - ' .. filename)
-end
-
-
```
TimZaman: So no resuming-feature any time soon I recon?

TimZaman:
I'm afraid you forgot to change the part where there where intermediate epochs saved.
https://github.com/NVIDIA/DIGITS/blob/master/tools/torch/main.lua#L822-827, @gheinrich.
How about:
```
-- if required, save snapshot at the end
if opt.epoch > last_snapshot_save_epoch then
    local filename
    local modelObjectToSave
    if model.clearState then
        filename = paths.concat(opt.save, snapshot_prefix .. '_' .. opt.epoch .. '_Model.t7')
        modelObjectToSave = model:clearState()
    else
    --stepvalues-- this version of Torch doesn't support clearing the model state => save only the weights
        filename = paths.concat(opt.save, snapshot_prefix .. '_' .. opt.epoch .. '_Weights.t7')
        modelObjectToSave = Weights
    end
    logmessage.display(0,'Snapshotting to ' .. filename)
    torch.save(filename, modelObjectToSave)
    logmessage.display(0,'Snapshot saved - ' .. filename)
end
```

gheinrich:
> I'm afraid you forgot to change the part where there where intermediate epochs saved

Well spotted, thanks!

gheinrich:
I have updated the commit to take account of comments from @TimZaman

TimZaman:
Works.

tools/torch/main.lua
```diff
@@ -370,15 +388,22 @@ end
...
     if paths.filep(opt.weights) then
         logmessage.display(0,'Loading weights from pretrained model - ' .. opt.weights)
-        local w, grad = model:getParameters()
-        w:copy(torch.load(opt.weights))
+        if (string.find(opt.weights, '_Weights')) then
+            local w, grad = model:getParameters()
+            w:copy(torch.load(opt.weights))
+        else
+            -- the full model was saved
+            assert(string.find(opt.weights, '_Model'))
```
TimZaman: I have found problems with this approach when using external models that were not created in digits (we can thus finally use in digits now, since we dump the full model and not just the weights! Yay!
I would therefore just remove the assertion for _Model. Or maybe, if possible, load in the file and see if there's a model structure in there if you really want to be clean.
This approach as it is now requires someone to change the filename of his pretrained model.

TimZaman:
Bonjour Greg, I have inserted a comment about the usage of models that do not have the **_Model** string in the filename.

Issue #670, opened 2016-04-04
JavaScript errors with Safari


lukeyeager:
Several lab users today ran into issues using DIGITS with Safari.

They couldn't select different solver types or see advanced learning rate options. 

It's probably a quick fix.

TimZaman:
I can confirm this issue with the latest digits master version and Safari Version 9.0.3 (10601.4.4):

![image](https://cloud.githubusercontent.com/assets/7721540/14281783/593330a2-fb3b-11e5-9a36-12cef940ce4e.png)

peterneher:
I have the same issue with Firefox 38.5.2

jmancewicz:
#702 fixes this in Safari.

jmancewicz:
I'm using,"Show advanced learning rate options"

PR #702, opened 2016-04-22
fix bug in solver_type selector

jmancewicz:
This selector, although missing a closing bracket, works in chrome and firefox, but not in safari. Seemed like a good thing to fix just in case browsers become more picky in the future.


lukeyeager:
Partially fixes https://github.com/NVIDIA/DIGITS/issues/670.

Is there a similar fix for the "Advanced Learning Rate" checkbox?

lukeyeager:
LGTM!

PR #708, opened 2016-04-26
Add parameter sweep to learning rate and batch size

jmancewicz:
Adding the ability to sweep Batch Size and Learning Rate.  Their entry fields now accept multiple values.

image

image

Each permutation of the swept fields will generate one job. With Batch Size [1,2] and Learning Rate [0.01, 0.015, 0.02], six jobs will be created.  The job names are augmented by the swept field value.  When multiple jobs are launched, the page is redirected to the home page rather than a job page.

image

If you were to clone one of these jobs, it will only have the values of that particular job. The clone data is saved per job.

### DIFF
--- a/digits/model/forms.py
+++ b/digits/model/forms.py
@@ -120,10 +120,10 @@ class ModelForm(Form):
             tooltip = "If you provide a random seed, then back-to-back runs with the same model and dataset should give identical results."
             )
 
-    batch_size = utils.forms.IntegerField('Batch size',
+    batch_size = utils.forms.MultiIntegerField('Batch size',
             validators = [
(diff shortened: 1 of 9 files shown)


lukeyeager:
Looks great, thanks for the screenshots!

At first glance, the code works great and looks to be written well - no complaints!

Let's add a simple test so we don't break it in the future and I think we're good to go. @gheinrich do you want to take a look too?

jmancewicz:
Added tests

lukeyeager:
Tests look good (despite the fact that TravisCI can't complete before the time limit :angry:).

Thanks for the great work!

Issue #706, opened 2016-04-25
Show runtime instead of total time on home page


pansk:
When executing more jobs than the number of GPUs available, jobs get queued.
At the end, the "Runtime" field is filled with the difference between the task creation and the task end, which is not the actual time spent training the network.
Is it possible to add a new field containing the actual training time (task end time - task start time), or change the meaning of this field? (the same number appears in the main screen, under Status).

lukeyeager:
Doesn't the "Running" section tell you what you want to know?

![job-status-timings](https://cloud.githubusercontent.com/assets/687269/14792289/6c503c1e-0ace-11e6-9c9c-01d4e7701874.jpg)

pansk:
Yes, it is indeed the "running" value.
What I mean, it would be more handy to have that number in the main screen and/or in the detail view, rather than the total running time. It's useful to quickly compare efficiency of two different network architectures.

Since we're on that page, another interesting number to show would be the top memory allocation during training (but I see it could have little meaning: apparently when I had two DIGITS instance running, it reported the total amount of GPU memory occupied, not the one allocated by this process). On dedicated boxes it could be quite handy, anyway.

TimZaman:
Meh, I'm not sure those add a lot. Those are particular wishes you have and are nice to compare efficiency, but from my own experience i have never wanted to quanitfy that. I am more interested for example in accuracy that's why we have this for accuracy (which i care about a lot, so i can kill the one's that are doing bad).
![image](https://cloud.githubusercontent.com/assets/7721540/14826985/90ac8866-0be0-11e6-8c46-d9cb7e09dccd.png)
Then we also expanded the hardware view a bit:
![image](https://cloud.githubusercontent.com/assets/7721540/14827031/b92cb4e6-0be0-11e6-8946-680034f1219f.png)
I could do a pull request for this, but i am unsure if it's wanted in the first place, regarding UI things.

jmancewicz:
Showing the runtime rather than the total time makes a lot of sense. I'll make that change.
The accuracy graph might be worth adding to the running jobs as well.

gheinrich:
I think the CPU utilization widget is interesting too.

The colour-coded confusion matrix is nice but I'm afraid people will be asking for the raw numbers if we switch to this. How does the epoch slider work?

TimZaman:
[...]

PR #709, opened 2016-04-26
change the Done time to be the running rather than elapsed

jmancewicz:
This addresses the first item in #706.


mpbrigham:
Model job running time is not retrieved with commit e2b3514, since job.status_history (which doesn't contain wait time) is accessed rather than job.tasks[-1].status_history. Commit 5c20784 displays both running and elapsed times.

PR #713, opened 2016-04-28
Show dataset information on model clone

### LINKED ISSUE #711: Dataset information not shown when cloning a model
When creating a new model and selecting a dataset:
digits2

When cloning a model:
digits
If you only have one dataset, I don't see a way of having the information displayed.
If you have multiple datasets in the list, click on any other dataset, then back on the one you want.

gheinrich:
close #711


lukeyeager:
Verified both bug and fix. Thanks!

PR #714, opened 2016-04-28
Update job form data on name edit

### LINKED ISSUE #710: Cloning a renamed job uses the old name
1. Create a job (dataset or model) and call it "A"
2. Rename the job to "B"
3. Use "Clone Job"
4. The default name for the cloned job is the original "A", instead of B.

Annoying when I accidentally called a job "CIFAR Color" but instead it should be "CIFAR Grayscale", I have to fix all the cloned jobs afterwards.

gheinrich:
close #710

### DIFF
--- a/digits/views.py
+++ b/digits/views.py
@@ -178,6 +178,14 @@ def edit_job(job_id):
         if not name:
             raise werkzeug.exceptions.BadRequest('name cannot be blank')
         job._name = name
+        # update form data so updated name gets used when cloning job
+        if 'form.dataset_name.data' in job.form_data:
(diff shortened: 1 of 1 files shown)


gheinrich:
Didn't know how to write a more elegant fix without changing a lot of code

lukeyeager:
Seems like that should work. @jmancewicz will you take a quick look since you wrote the cloning code?

jmancewicz:
This is a fine fix, but I'll need to change the sweep a little. Currently, the sweep modifies the job name in a way that you would not want the form.name to change.  I'll make that change soon.

Very nice, @gheinrich.

PR #719, opened 2016-04-29
Use dictionary for dataset/model options

gheinrich:
This will make it easier to add dataset/model types

### DIFF
--- a/digits/templates/home.html
+++ b/digits/templates/home.html
@@ -44,13 +44,15 @@ function deleteJob(job_id) {
 <div class="pull-right">
     New Dataset
     <ul class="nav nav-pills">
-        {% for category in new_dataset_options %}
+        {% for category in new_dataset_options.keys()|sort %}
+        {% set options = new_dataset_options[category] %}
         <li class="dropdown active">
             <a href="#" class="dropdown-toggle" data-toggle="dropdown">
-                {{category[0]}} <span class="caret"></span>
+                {{category}} <span class="caret"></span>
             </a>
             <ul class="dropdown-menu">
-                {% for option in category[1] %}
+                {% for id in options.keys()|sort %}
+                {% set option = options[id] %}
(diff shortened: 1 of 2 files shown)


lukeyeager:
Fine with me. But you probably want to switch to using an [OrderedDict](https://docs.python.org/2/library/collections.html#collections.OrderedDict) to preserve the order when displayed on the page.

gheinrich:
Yes, possibly. In my preliminary work on data extensions [there](https://github.com/gheinrich/DIGITS/blob/dev/data-extensions/digits/views.py#L65-L76) I am adding items to the dictionary by iterating through the list of supported extensions. I am not sure what's best between the default (alphanumeric) ordering of a dictionary or the chronological order of an `OrderedDict`. Any preference? 

For consistency perhaps the values of `new_dataset_options` and `new_model_options` should also be (ordered) dictionaries? With the `id` field being the key.

lukeyeager:
> In my preliminary work on data extensions there I am adding items to the dictionary by iterating through the list of supported extensions.

Cool! I like it.

> I am not sure what's best between the default (alphanumeric) ordering of a dictionary or the chronological order of an OrderedDict. Any preference?

Well iterating through keys of a `dict` doesn't promise any particular order:
```py
>>> {'a':1,'b':2,'d':3,'c':4}.keys()
['a', 'c', 'b', 'd']
```
So either use an `OrderedDict` or iterate through `sorted(dict.keys())`. The latter might make more sense depending on how you implement `extensions.data.get_extensions()`.

> For consistency perhaps the values of new_dataset_options and new_model_options should also be (ordered) dictionaries? With the id field being the key.

Sure, yeah. I guess OrderedDicts are slightly easier to work with than lists of tuples.

gheinrich:
Thanks for the feedback. I've changed the code to use only dictionaries. Since the sorting is a visualization preference this is done from the Jinja template.

PR #722, opened 2016-05-02
Fix Torch visualization of color image layers

### LINKED ISSUE #721: Image visualization
With torch, images are fed into the network as BGR (I assume), but RGB convention when displaying intermediate results, so intermediate results show up in false colors.

To show the issue, I created a network with just one multiply(1) layer, here's the inference result:

lenabgrrgb

gheinrich:
close #721

### DIFF
--- a/digits/utils/image.py
+++ b/digits/utils/image.py
@@ -294,6 +294,7 @@ def get_layer_vis_square(data,
         normalize = True,
         min_img_dim = 100,
         max_width = 1200,
+        channel_order = 'RGB',
         ):
     """
     Returns a vis_square for the given layer data
@@ -306,6 +307,8 @@ def get_layer_vis_square(data,
     normalize -- whether to normalize the data when visualizing
     max_width -- maximum width for the vis_square
     """
+    if channel_order not in ['RGB', 'BGR']:
+        raise ValueError('Unsupported channel_order %s' % channel_order)
     if data.ndim == 1:
         # interpret as 1x1 grayscale images
         # (N, 1, 1)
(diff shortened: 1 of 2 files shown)


digits/utils/image.py
```diff
@@ -306,6 +307,7 @@ def get_layer_vis_square(data,
     normalize -- whether to normalize the data when visualizing
     max_width -- maximum width for the vis_square
     """
+    assert channel_order == 'RGB' or channel_order == 'BGR'
```
lukeyeager: How about:
```python
if channel_order not in ['RGB', 'BGR']:
  raise ValueError('Unsupported channel_order %s' % channel_order)
```
> Raised when a built-in operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.
https://docs.python.org/2/library/exceptions.html#exceptions.ValueError

I've got `asserts` throughout the code, but I'm trying to phase them out where appropriate.

gheinrich: OK no problem I'll make a mental note not to use asserts :-)

lukeyeager:
Since RGB is the more standard representation (OpenCV users excepted), can we make that the default for `get_layer_vis_square`? Caffe should need to provide the extra flag, not Torch.

This is obviously not a big deal, but will make integrating other frameworks slightly more straightforward.

pansk:
Wouldn't switching to a different representation require changing the way how the image is fed to the network, and thus break previously trained networks?
In this case a conversion tool (which swaps weights and biases for R she B layers) could be quite useful, but I really don't know how to determine automatically if a layer is receiving/producing a RGB image.

gheinrich:
> Wouldn't switching to a different representation require changing the way how the image is fed to the network

This patch doesn't change the way data are fed to the network. The network (for Torch) still receives data in RGB order.

gheinrich:
> Since RGB is the more standard representation (OpenCV users excepted), can we make that the default for get_layer_vis_square?

I thought so too but didn't dare to change the Caffe path :-)

I have updated the commit to make RGB the default channel order. Thanks for the review!

@pansk could you try 395dbb3 ? Thanks!

pansk:
Sorry for the misunderstanding, I thought Luke was also referring to how the network is fed.
The solution is perfect, thank you!

gheinrich:
Thanks for the feedback. I'll merge the change then.

PR #723, opened 2016-05-02
Expose common dataset interface

gheinrich:
This will allow to create new types of datasets without making changes in other parts of DIGITS.

Ultimately, class relationships may look like:
data-classes

### DIFF
--- a/digits/dataset/images/classification/job.py
+++ b/digits/dataset/images/classification/job.py
@@ -1,10 +1,12 @@
 # Copyright (c) 2014-2016, NVIDIA CORPORATION.  All rights reserved.
 from __future__ import absolute_import
 
+import os
+
 from ..job import ImageDatasetJob
 from digits.dataset import tasks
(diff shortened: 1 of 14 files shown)


lukeyeager:
I like the direction! Looks like this is a work-in-progress?

Looks like you removed `json_dict` from `DatasetJob` and didn't add it to `GenericImageDatasetJob`.

gheinrich:
I didn't base this commit on the latest version of master because of the parameter-sweep issue. Other than that, I feel this could be good to merge if it passes the test of your most accurate review filter :-)

I didn't add `json_dict` to `GenericImageDatasetJob` because the version I removed from `DatasetJob` was classification specific (i.e. it had no effect on the generic path).

gheinrich:
Rebased on tip of master branch.

lukeyeager:
I think this is working. I ran into some Torch errors, but I don't think they were related.

PR #726, opened 2016-05-03
fix optional multi-fields

### LINKED ISSUE #717: Bugs with parameter sweep
*See https://github.com/NVIDIA/DIGITS/pull/708*

Don't have time to track them down and fix them right now:

1. If you try to create a model without specifying a batch size, then you get an error:

    sweep-bug-error1

2. On the page with the error, the fields are filled in with weird values, causing problems later:

    sweep-bug-error2

jmancewicz:
*Close #717*


lukeyeager:
Works for me, thanks!

PR #727, opened 2016-05-04
Allow MultiNumberRange to be inclusive or exclusive

jmancewicz:
The Learning Rate range validator has a min of 0.0 inclusive, but a learning rate of 0 seems bad, (I may be wrong), so, I've added min_inclusive and max_inclusive booleans to that 0 is excluded.

### DIFF
--- a/digits/utils/forms.py
+++ b/digits/utils/forms.py
@@ -339,26 +339,39 @@ class MultiNumberRange(object):
         interpolated using `%(min)s` and `%(max)s` if desired. Useful defaults
         are provided depending on the existence of min and max.
     """
-    def __init__(self, min=None, max=None, message=None):
+    def __init__(self, min=None, max=None, min_inclusive=True, max_inclusive=True, message=None):
         self.min = min
         self.max = max
         self.message = message
+        self.min_inclusive = min_inclusive
+        self.max_inclusive = max_inclusive
 
     def __call__(self, form, field):
         fdata = field.data if isinstance(field.data, (list, tuple)) else [field.data]
         for data in fdata:
-            if data is None or (self.min is not None and data < self.min) or \
-                    (self.max is not None and data > self.max):
+            print 'checking', data, self.min
+            flags = 0
+            flags |= (data is None) << 0
+            flags |= (self.min is not None and self.min_inclusive and data < self.min) << 1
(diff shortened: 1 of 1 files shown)


lukeyeager:
Actually I've found it useful to have a learning rate of zero before. I was importing a pre-trained model and didn't want to change the weights at all (granted, we should just create an interface for doing that).

gheinrich:
Maybe this shouldn't apply to the learning rate as @lukeyeager has an ingenious use case there but exclusive fields sound like a useful thing to have. I think it's a good idea.

digits/utils/forms.py
```diff
@@ -339,26 +339,39 @@ class MultiNumberRange(object):
...
-            if data is None or (self.min is not None and data < self.min) or \
-                    (self.max is not None and data > self.max):
+            print 'checking', data, self.min
+            flags = 0
+            flags |= (data is None) << 0
+            flags |= (self.min is not None and self.min_inclusive and data < self.min) << 1
+            flags |= (self.max is not None and self.max_inclusive and data > self.max) << 2
+            flags |= (self.min is not None and not self.min_inclusive and data <= self.min) << 3
+            flags |= (self.max is not None and not self.max_inclusive and data >= self.max) << 4
+
```
gheinrich: Wondering why you are test the condition here and not directly below (i.e. without setting `flags` as an intermediate step)? I don't mind either way, I am just wondering if there is a trick to learn there :-)

jmancewicz: I had borrowed if from the wtforms case where there is only a min and max, and it has logic for cases min!=None, max!=None, and (min!=None && max!=None).  I had initially started working on a message such that there was one message that handled any permutation of settings, that's when I added the flags, and decided against it.  I don't find the flags offensive, but would be happy to change them if it's confusing or messy.

jmancewicz:
@lukeyeager, I just KNEW there would be a case for learning rate == 0!

So, I'll remove it from the learning rate field, but I'll keep it around and maybe make number range valuators with this option in the future.

lukeyeager:
> So, I'll remove it from the learning rate field

ping

jmancewicz:
@lukeyeager, I removed this from the learning rate.

lukeyeager:
> I removed this from the learning rate.

Thanks!

PR #728, opened 2016-05-04
Home page tables

jmancewicz:
I’ve been experimenting with incorporating the angularjs table (and jQuery/angularjs selection model) that I wrote for an internal app into DIGITS.  I know that having a selection model is not popular for web based interfaces and that jQuery actively discourages disabling text selection, but I find the benefits are worth it for the way I like interfaces to feel.

That said, I made a version of the front page with the Datasets and Models columns converted into tables.

screen shot 2016-05-04 at 2 15 30 pm

Space is tight for the headers and the Datasets starts to wrap. I can probably tweak that.

Moving to the selectable tables, we gain sorting, filtering, and selection.

Row sorting is indicated by the caret in the table header.

Filtering reduces the list to the matching items.

screen shot 2016-05-04 at 2 16 50 pm

And selection allows the user to delete all the selected jobs at one time.

The Refs column on the Datasets displays the number of Models that reference that Dataset. That allowed me to easily remove unused Datasets.

Aesthetically, I like the the look of the current DIGITS display, and functionally, I like the table view.

So, out of curiosity, I implemented an alternative view which looks somewhat more like the original (check the Toggle View checkbox).

screen shot 2016-05-04 at 2 17 25 pm

Which still has all the filter, sorting, and selection functionality, but a different view.  Without the Delete button on each item, there’s more white space, and perhaps moving Status to the right of Submitted would be better. When the window is smaller, though that will wrap.

The toggle is temporary and intended for anyone who wants to download my home-page-tables branch and see how it feels. I have a very short list of jobs so everything is pretty snappy. I’d be curious to hear how responsive it feels to others.

Also, I’ve removed pagination and just have a long list. I prefer this for data that is bounded and not slow to load. If it is desired to have pagination returned, I can add it back in.  I think the angular version is pretty nice.


jmancewicz:
Just added multi-column sorting with shift click.

screen shot 2016-05-04 at 4 26 05 pm

Here I clicked on Status header to see the jobs with Error status, then shift-clicked the Elapsed header to sub-sort by Elapsed time. This would be more useful if something like Loss, Accuracy, or some other metric was listed. However, the is the precursor to the work of grouping the jobs, so I put it out there.

TimZaman:
Lovely! In the current viewer, when i was at 'model page' 2, and clicked on a model, and pressed return, it would return me to the home page (default page 1 i guess). It would be nice to have it return to the same page you were browsing before. Otherwise it's near impossible to have quick looks without opening multiple tabs.

lukeyeager:
Oops I made a merge conflict when merging https://github.com/NVIDIA/DIGITS/pull/719.

digits/static/js/home_app.js
```diff
@@ -0,0 +1,190 @@
+"use strict";
```
lukeyeager: Did you write this? It's missing a copyright notice.

lukeyeager:
This looks awesome! I just used it to delete 38 jobs all at once - that used to take forever.

digits/static/js/home_app.js
jmancewicz: Ah, right.

PR #731, opened 2016-05-10
Data extension framework

gheinrich:
### Summary ###

This adds a data extension framework along the lines of:

data-classes

This also adds an image gradient extension (same as this [tutorial](https://github.com/NVIDIA/DIGITS/tree/digits-3.0/examples/regression) )

Samples extensions will be added in separate pull requests.

This pull requests depends on #723 

### Progress ###

- [x] add data extension framework
- [x] add templates
- [x] add views
- [x] rebase to tip of master branch / squash commits
- [x] add sample extension
- [x] add tests

Documentation may be added later when we want to advertise the feature.

### DIFF
--- a/digits/dataset/generic/job.py
+++ b/digits/dataset/generic/job.py
@@ -1,0 +1,110 @@
+# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
+from __future__ import absolute_import
+
+from ..job import DatasetJob
+from digits.dataset import tasks
+from digits.utils import subclass, override, constants
+
+# NOTE: Increment this everytime the pickled object changes
+PICKLE_VERSION = 1
+
+
+@subclass
+class GenericDatasetJob(DatasetJob):
+    """
+    A Job that creates a dataset using a user-defined extension
+    """
+    def __init__(self,
+                 backend,
+                 feature_encoding,
+                 label_encoding,
+                 batch_size,
+                 num_threads,
+                 extension_id,
+                 extension_userdata,
+                 **kwargs
+                 ):
+        self.backend = backend
+        self.feature_encoding = feature_encoding
+        self.label_encoding = label_encoding
+        self.num_threads = num_threads
+        self.batch_size = batch_size
+        self.extension_id = extension_id
+        self.extension_userdata = extension_userdata
+
+        super(GenericDatasetJob, self).__init__(**kwargs)
+        self.pickver_job_dataset_extension = PICKLE_VERSION
+
+        # create tasks
+        for stage in [constants.TRAIN_DB, constants.VAL_DB, constants.TEST_DB]:
+            self.tasks.append(tasks.CreateGenericDbTask(
+                job_dir=self.dir(),
+                job=self,
+                backend=self.backend,
+                stage=stage,
+                )
+            )
+
+    def __setstate__(self, state):
+        super(GenericDatasetJob, self).__setstate__(state)
+        self.pickver_job_dataset_extension = PICKLE_VERSION
+
+    def create_db_task(self, stage):
+        for t in self.tasks:
+            if t.stage == stage:
+                return t
+        return None
+
+    def create_db_tasks(self):
+        return self.tasks
+
+    @override
+    def get_backend(self):
+        """
+        Return the DB backend used to create this dataset
+        """
+        return self.backend
+
+    @override
+    def get_entry_count(self, stage):
+        """
+        Return the number of entries in the DB matching the specified stage
+        """
+        return self.create_db_task(stage).entry_count
+
+    @override
+    def get_feature_db_path(self, stage):
+        """
+        Return the absolute feature DB path for the specified stage
+        """
+        return self.path(self.create_db_task(stage).dbs['features'])
+
+    @override
+    def get_feature_dims(self):
+        """
+        Return the shape of the feature N-D array
+        """
+        shape = self.create_db_task(constants.TRAIN_DB).feature_shape
+        if len(shape) == 3:
+            # assume image and convert CHW => HWC (numpy default for images)
+            shape = [shape[1], shape[2], shape[0]]
+        return shape
+
+    @override
+    def get_label_db_path(self, stage):
+        """
+        Return the absolute label DB path for the specified stage
+        """
+        return self.path(self.create_db_task(stage).dbs['labels'])
+
+    @override
+    def get_mean_file(self):
+        """
+        Return the mean file (if it exists, or None)
+        """
+        mean_file = self.create_db_task(constants.TRAIN_DB).mean_file
+        return self.path(mean_file) if mean_file else None
+
+    @override
+    def job_type(self):
+        return 'Generic Dataset'
--- a/digits/dataset/generic/test_views.py
+++ b/digits/dataset/generic/test_views.py
@@ -1,0 +1,261 @@
+# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
+from __future__ import absolute_import
+
+import json
+
+from bs4 import BeautifulSoup
+
+import digits.test_views
+
+from digits import extensions
+
+# May be too short on a slow system
+TIMEOUT_DATASET = 45
+
+################################################################################
+# Base classes (they don't start with "Test" so nose won't run them)
+################################################################################
+
+
+class BaseViewsTest(digits.test_views.BaseViewsTest):
+    """
+    Provides some functions
+    """
+
+    @classmethod
+    def dataset_exists(cls, job_id):
+        return cls.job_exists(job_id, 'datasets')
+
+    @classmethod
+    def dataset_status(cls, job_id):
+        return cls.job_status(job_id, 'datasets')
+
+    @classmethod
+    def abort_dataset(cls, job_id):
+        return cls.abort_job(job_id, job_type='datasets')
+
+    @classmethod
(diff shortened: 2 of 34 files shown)


examples/image-reconstruction/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+Output:
+![Loss](output.jpeg)
+
+Ground truth:
+![Loss](ground-truth.bmp)
+
+
+
+
```
TimZaman: This PR looks awesome! Here though, it's not immediatelly obvious to me what we're trying to do. Also it's unclear what i should see in the example.

gheinrich:
Thanks for the feedback. Indeed the image reconstruction tutorial is work in progress. The idea is to train a network that learns how to reconstruct distorted images but the model isn't working well. Mostly the example is there to show that you can add a data extension in DIGITS to create a dataset for this task pretty easily. Getting the model to actually work will require more efforts.

The main idea behind this PR is to allow anybody to import any data in DIGITS so you don't have to write a script to create the LMDB yourself. You could imagine importing data from text, csv, numpy, or anything you can read from Python. Possibilities are endless.

A subsequent PR should come to introduce a visualization framework, allowing DIGITS users to easily add and select view extensions to visualize the output of a network.

TimZaman:
Sounds amazing man, and will be very useful! We were in dire need of this a few weeks back, working on upscaling and incoloring.

gheinrich:
Finally got coverage to increase by adding more tests. I think this is ready for review. Thanks!

lukeyeager:
Something funky happened in your rebase with `master`, and now I can't get to any of the new stuff from the homepage.

digits/dataset/generic/job.py
```diff
@@ -0,0 +1,107 @@
...
+
+    @override
+    def job_type(self):
+        return 'Generic Dataset'
+
+
+
+
+
+
```
lukeyeager: Bunch of extra space here

digits/dataset/generic/views.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    Returns a form for a new GenericDatasetJob
+    """
+
+    form = GenericDatasetForm()
+
+    ## Is there a request to clone a job with ?clone=<job_id>
+    utils.forms.fill_form_if_cloned(form)
+
+    extension = extensions.data.get_extension(extension_id)
```
lukeyeager: Need to check for `None` here

digits/dataset/generic/views.py
```diff
@@ -0,0 +1,123 @@
...
+    utils.forms.fill_form_if_cloned(extension_form)
+
+    template, context = extension.get_dataset_template(extension_form)
+    rendered_extension = flask.render_template_string(template, **context)
+
+    return flask.render_template('datasets/generic/new.html',
+        extension_title = extension.get_title(),
+        extension_id = extension_id,
+        extension_html = rendered_extension,
+        form=form
```
lukeyeager: Spacing here doesn't match other arguments. According to `flake8`, there shouldn't be any spaces around the `=` in keyword arguments. We haven't done that in the rest of DIGITS, but we might as well start following the convention when it's convenient.

lukeyeager: No need to fix all of them, I'm just bringing it up since I'm looking at it.

digits/dataset/generic/views.py
```diff
@@ -0,0 +1,123 @@
...
+        job = GenericDatasetJob(
+            username            = utils.auth.get_username(),
+            name                = form.dataset_name.data,
+            backend             = form.dsopts_backend.data,
+            feature_encoding    = form.dsopts_feature_encoding.data,
+            label_encoding      = form.dsopts_label_encoding.data,
+            batch_size          = int(form.dsopts_batch_size.data),
+            num_threads         = int(form.dsopts_num_threads.data),
+            extension_id        = extension_id,
+            extension_userdata  = extension.get_user_data(),
```
lukeyeager: Let's do fix the whitespace here, though. I fixed it in some other places with https://github.com/NVIDIA/DIGITS/pull/744.

digits/dataset/tasks/create_generic_db.py
```diff
@@ -0,0 +1,142 @@
...
+            self.entry_count = count
+            return True
+
+        # feature shape
+        match = re.match(r'Feature shape for stage %s: (.*)' % self.stage, message)
+        if match:
+            self.feature_shape = eval(match.group(1))
+            return True
+
+        # feature shape
```
lukeyeager: *label shape

digits/extensions/data/interface.py
```diff
@@ -0,0 +1,104 @@
...
+
+class DataIngestionInterface(object):
+    """
+    A data ingestion extension
+    """
+
+    def __init__(self, **kwargs):
+        """
+        Init
+        """
```
lukeyeager: Probably don't need this docstring

digits/extensions/data/interface.py
```diff
@@ -0,0 +1,104 @@
...
+        This function is called on the main thread
+        The returned list will be spread across all reader threads
+        Reader threads will call encode_entry() with IDs returned by this function in no particular order
+        """
+        raise NotImplementedError
+
+
+
+
+
```
lukeyeager: Bunch of extra white space

digits/extensions/data/imageGradients/data.py
```diff
@@ -0,0 +1,95 @@
...
+@subclass
+class DataIngestion(DataIngestionInterface):
+    """
+    A data ingestion extension for an image gradient dataset
+    """
+
+    def __init__(self, **kwargs):
+        """
+        Init
+        """
```
lukeyeager: Probably don't need this docstring

gheinrich:
> Something funky happened in your rebase with master, and now I can't get to any of the new stuff from the homepage.

Actually I've made it such that the gradients extension doesn't show on the home page with this [line](https://github.com/gheinrich/DIGITS/blob/41cdd094ebfe16fc155d150417877504f206d34e/digits/extensions/data/__init__.py#L8). I thought this was an interesting extension to play with but maybe not something we want all users to see by default. Sorry I should have mentioned this.

digits/dataset/generic/test_views.py
```diff
@@ -0,0 +1,251 @@
...
+        job_id = self.create_dataset()
+        assert self.dataset_wait_completion(job_id) == 'Done', 'create failed'
+        assert self.delete_dataset(job_id) == 200, 'delete failed'
+        assert not self.dataset_exists(job_id), 'dataset exists after delete'
+
+    def test_invalid_number_of_reader_threads(self):
+        try:
+            job_id = self.create_dataset(json=True, dsopts_num_threads=0)
+            assert self.dataset_wait_completion(job_id) != 'Done', 'should have failed'
+        except RuntimeError:
```
lukeyeager: Your checks seem weird. Do you mean to allow the job to fail in one of two ways (either status != 'Done' or `RuntimeError`)?

digits/dataset/generic/views.py
gheinrich: Sure, let me try to fix all the ones I can spot in the new code!

lukeyeager:
Yeah, I found it eventually (edited my post above)

lukeyeager:
Since you're hiding the only new feature you added, I'm not sure how to test the UI, exactly. This is just one of several "enabler" pull requests, right? Not much real added value yet?

digits/dataset/generic/test_views.py
gheinrich: actually I am expecting the call to `create_dataset` to fail I could remove the assert on line 142

lukeyeager: Or just add an `assert False` if you don't expect to ever get there.

gheinrich:
I have fixed a lot of PEP8 errors in the latest commits. Hopefully I didn't break anything. I'll do another round of testing tomorrow but I'm still interested in your feedback.

To test you can enable the gradient extension in `/digits/extensions/data/__init__.py`. I am thinking we don't want to show this by default.

Note that in the dataset creation form, one subset of the fields is coming from the data extension. The other subset is coming from `/digits/dataset/generic/forms.py`. The latter subset corresponds to the options that apply to all datasets (DB backend, encoding, number of encoder threads, ...).

gheinrich:
Rebased again...

lukeyeager:
It's kinda weird that there's a noop test_db task created for object detection datasets:

@gheinrich was this your intention?

object-detection-test-dataset

gheinrich:
For data extensions we always spawn a `test_db` task from which we create an instance of a `create_generic_db.py` process. What happens within this process depends on the data extension specifics: in the case of object detection there is no provision to create a test DB so this never actually does anything. The gradient extension, for example, may create an actual test DB if instructed to do so.

So to answer the question: we always need to create a `test_db` task because we don't know what the extension will choose to do about it. Are you saying we shouldn't show it in the UI if it didn't actually create a database? That would certainly make sense.

lukeyeager:
> Are you saying we shouldn't show it in the UI if it didn't actually create a database? That would certainly make sense.

I guess that would work, yeah. Either detect that we won't need the task and don't create it (sounds nice), or just hide the fact that we created a useless task (also works).

gheinrich:
We can't detect that we don't need the task because the itemization is running in the external process. But we can not show the task if it's done (without errors) and its entry count is zero. This is implemented in #813.

PR #734, opened 2016-05-10
Group jobs on the front page

jmancewicz:
There's been a request to provide the ability to group jobs on the front page so that related jobs can be grouped together.

This is a first pass at adding that sort of feature to the front page of DIGITS.

In this branch, there is a Group checkbox that toggles grouping in the view which is on by default.
screen shot 2016-05-10 at 4 10 40 pm

Without grouping
screen shot 2016-05-10 at 4 12 47 pm

The Toggle View checkbox toggles between the table style view and the list style, but the list style will be removed soon.  I had just wanted to see how I liked it, and it turns out that I don't.

screen shot 2016-05-10 at 4 14 46 pm

It's a real estate hog.

There's a Group button which assigns the group for the selected jobs.

Select jobs and click the Group button
screen shot 2016-05-10 at 4 15 52 pm

Enter a name
screen shot 2016-05-10 at 4 16 28 pm

Selected jobs are assigned to that group.
screen shot 2016-05-10 at 4 17 30 pm

The group is a property of the job, basically a specific tag. Currently, the only way to assign groups is through the above method. Groups could be assigned at creation time with an optional field near the name field. That way jobs that are generated with a parameter sweep could all be assigned the same group.

This is not ready for code review. I'm going to move the javascript for this and #728 into separate files, and tidy up a little.


gheinrich:
That looks great! The dataset and model lists are showing pretty much the same thing except for framework/backend. Could we merge these lists to show a single job list? That would save a lot of space on the main page and with the filters it would be easy to select between datasets and models.

lukeyeager:
https://github.com/NVIDIA/DIGITS/pull/728 is merged. Looks like it caused a conflict with this one?

lukeyeager:
How do I un-group jobs?

jmancewicz:
You can set the group to ungrouped, which is a group. I can make it so that empty string maps to ungrouped and I could add an ungroup button, but I don't love that because of UI bloat.

lukeyeager:
> #728 is merged. Looks like it caused a conflict with this one?

ping

jmancewicz:
This is a reimplementation of the group changes onto the recent master. It's pretty much the same. The Group Name modal explains that a blank name ungroups the jobs.

screen shot 2016-07-06 at 10 33 51 am

There are now Group Name forms added to new Job pages.

screen shot 2016-07-06 at 10 35 51 am

The Group Jobs toggle is saved to local storage, which means that it is shared with all the window/tabs viewing DIGITS. If there are multiple windows open viewing DIGITS front page, and the Group Jobs is toggled in one window, it will toggle in all windows. Likewise, the Models table column visibility settings are saved in local storage, and changes in one view affect the other views.

To allow potentially desirable differences in views, things like the column sorting state and group bar collapsed state are not saved in local storage and are therefore not persistent between sessions.

jmancewicz:
@gheinrich or @lukeyeager, any objection to me merging this?

digits/views.py
```diff
@@ -218,6 +222,56 @@ def get_job_list(cls, running):
...
+    not_found = 0
+    forbidden = 0
+    group_name = utils.routing.get_request_arg('group_name').strip()
+    job_ids = flask.request.form.getlist('job_ids[]')
+
+    for job_id in job_ids:
+        try:
+            job = scheduler.get_job(job_id)
+            if job is None:
+                print '%s not found' % job_id
```
gheinrich: maybe replace with log points (`logger.warning(...)`)

digits/job.py
```diff
@@ -56,6 +56,7 @@ def __init__(self, name, username, persistent = True):
         self._id = '%s-%s' % (time.strftime('%Y%m%d-%H%M%S'), os.urandom(2).encode('hex'))
         self._dir = os.path.join(config_value('jobs_dir'), self._id)
         self._name = name
+        self.group = group
```
gheinrich: question: shall we increment `PICKLE_VERSION` and set `group='' ` in `__setstate__()` for earlier versions? I see that in `json_dict()` your are checking whether the attribute `group` exists. I am wondering what is best between setting all fields when loading picked jobs (and using default values when loading older jobs that don't have the new field) or checking later on at run time whether the field exists... I have a tiny inclination towards the former method but I'm curious to know what you think.

gheinrich:
Hi Joe, that's a very useful PR. Is there a way in which those Javascripts can be updated without requiring a browser cache flush? When you switch back and forth between these PRs the home page UI acts erratically.

lukeyeager:
> Is there a way in which those Javascripts can be updated without requiring a browser cache flush? When you switch back and forth between these PRs the home page UI acts erratically.

I agree, this is annoying. Rails has a nice solution. I wonder if there's a parallel for Flask?

> In production, Rails inserts an MD5 fingerprint into each filename so that the file is cached by the web browser. You can invalidate the cache by altering this fingerprint, which happens automatically whenever you change the file contents.
http://edgeguides.rubyonrails.org/asset_pipeline.html#main-features

digits/views.py
jmancewicz: Done.

digits/job.py
jmancewicz: @gheinrich, I think moving the `group` to the `__setstate__` is the right thing to do. Having the group added at construction and load removes the need to checks in any other routes that might need the group. It is much cleaner. Thank you.

The `PICKLE_VERSION` bump doesn't facilitate any conditional code distinguishing jobs with and without `group` set, but as a matter of protocol, it seems like a fine thing to do.

jmancewicz:
I'm unable to reproduce the erratic behaviour. I've been swapping between this branch and master as well as staying on this branch and checking out master versions of the js files. I've been reloading the home page while DIGITS restarts with reloader, and still don't see anything out of the ordinary. @lukeyeager, perhaps you can help me reproduce this Monday.

digits/job.py
```diff
@@ -15,7 +15,7 @@
 from digits.utils import sizeof_fmt, filesystem as fs
 
 # NOTE: Increment this everytime the pickled object changes
-PICKLE_VERSION = 1
+PICKLE_VERSION = 2
```
jmancewicz: I'm not sure a version bump is warranted for this change.

PR #735, opened 2016-05-13
Object detection workflow

gheinrich:
Planning to split this into small pull requests. This is the complete object detection work flow.

This includes frameworks for data and view extensions.

The object detection data extension creates datasets from object detection datasets. On the main page, select `New dataset > Images > Object Detection`
select-object-detection

The dataset is configured with a form:
object-detection-form

To create the model: on the main page, select `New Model > Images > Object Detection`.

After the model is trained, on the model page, select the bounding box view extension:
select-bounding-box-view

Then proceed to inference as usual. If you do `Test Many`, you may see something like:
detector-inference

~~This includes as a bonus:~~
~~- an image gradient work flow (the data extension creates datasets as in the image regression example) and the view extension displays arrows that point to the "light" in the images,~~
~~- an image processing work flow (the data extension creates datasets with images as inputs/outputs - useful for FCNs)~~

Example of gradient visualization:
gradient-config

gradients-visualization

### DIFF
--- a/digits/extensions/data/objectDetection/utils.py
+++ b/digits/extensions/data/objectDetection/utils.py
@@ -1,0 +1,282 @@
+# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
+
+import csv
+import cv2 as cv
+import numpy as np
+import os
+
+class ObjectType:
+
+    Dontcare, Car, Van, Truck, Bus, Pickup, VehicleWithTrailer, SpecialVehicle,\
+    Person, Person_fa, Person_unsure, People, Cyclist, Tram, Person_Sitting,\
+    Misc = range(16)
+
+    def __init__(self):
+        pass
+
+class Bbox:
+
+    def __init__(self, x_left=0, y_top=0, x_right=0, y_bottom=0):
+        self.xl = x_left
+        self.yt = y_top
+        self.xr = x_right
+        self.yb = y_bottom
+
+    def area(self):
+        return (self.xr - self.xl) * (self.yb - self.yt)
+
+    def width(self):
+        return self.xr - self.xl
+
+    def height(self):
+        return self.yb - self.yt
+
+    def get_array(self):
+        return [self.xl, self.yt, self.xr, self.yb]
+
+class GroundTruthObj:
+
+    """ This class is the data ground-truth
+
+        #Values    Name      Description
+        ----------------------------------------------------------------------------
+        1    type         Describes the type of object: 'Car', 'Van', 'Truck',
+                          'Pedestrian', 'Person_sitting', 'Cyclist', 'Tram',
+                          'Misc' or 'DontCare'
+        1    truncated    Float from 0 (non-truncated) to 1 (truncated), where
+                          truncated refers to the object leaving image boundaries.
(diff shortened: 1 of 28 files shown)


lukeyeager:
```
  File "/home/lyeager/digits/digits/extensions/data/kittiObject/data.py", line 139, in encode_entry
    datum_bbox = caffe.io.bbox_to_datum(label,
AttributeError: 'module' object has no attribute 'bbox_to_datum'
```
I removed `bbox_to_datum` from `caffe.io` because I figured we could just include it in our code rather than checking it into Caffe.

lukeyeager:
Error handling for creating KITTI datasets is not good. Exceptions raised while the task is running should be caught and logged as errors, like so:
https://github.com/NVIDIA/DIGITS/blob/v3.3.0/tools/create_db.py#L779-L781
That will allow DIGITS to display the proper error message.

Also, I can't see any results on the `dataset#show` page, including the task log file. This is super helpful for debugging other failed tasks.

digits/extensions/data/kittiObject/data.py
```diff
@@ -0,0 +1,264 @@
...
+import operator
+import os
+import random
+
+from digits.utils import image, subclass, override, constants
+from ..interface import DataIngestionInterface
+from .forms import DatasetForm
+from .kitty import GroundTruth, GroundTruthObj
+from .kitty import cropImage_onlyForImage, cropImage_onlyForLabel
+from .kitty import resize_bbox_list, resizeImage_onlyForImage
```
lukeyeager: Did you mean to use "kitti" instead of "kitty"?

gheinrich:
> Exceptions raised while the task is running should be caught and logged as errors, like so:
https://github.com/NVIDIA/DIGITS/blob/v3.3.0/tools/create_db.py#L779-L781

I actually had that but the problem was I was missing something like this:
https://github.com/NVIDIA/DIGITS/blob/v3.3.0/digits/dataset/tasks/create_db.py#L218-L221

I have updated the "data extension" framework with better error handling (including those that occur in reader threads). So you would now get something like:
create-dataset-exceptions

gheinrich:
Updated to remove dependency on OpenCV bindings for Python.

digits/extensions/data/objectDetection/utils.py
```diff
@@ -0,0 +1,284 @@
+#!/usr/bin/env python2
```
lukeyeager: Is this script really executable?

gheinrich: Whoops. I'll remove that line.

lukeyeager:
Woohoo I just successfully completed the whole object detection workflow!

gheinrich:
> Woohoo I just successfully completed the whole object detection workflow!

Excellent! Which SHA1 did you use (there's been quite a few revisions so I'd like to make sure you've been using one of the last ones).

lukeyeager:
The latest for this branch.

I think we need to figure out how to add DetectNet as a "default/standard/zoo" network before this makes sense to merge, maybe? Also, maybe we should enable the gradient thing? It could be a cool walkthrough demo for DIGITS.

PR #739, opened 2016-05-15
Showing runtime instead of total time on home page and models overview page

mpbrigham:
Related to #706 and #709

### DIFF
--- a/digits/model/views.py
+++ b/digits/model/views.py
@@ -37,7 +37,7 @@ def index():
     for rjob in raw_jobs:
         train_outs = rjob.train_task().train_outputs
         val_outs = rjob.train_task().val_outputs
-        history = rjob.status_history
+        history = rjob.train_task().status_history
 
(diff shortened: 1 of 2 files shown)


lukeyeager:
The trouble here is that one job can have many tasks. For example, here's a job which consists of a graph of 5 tasks, some of which run concurrently:

5-running-tasks

Since they run concurrently, you can't just add the runtime of each task together. And you certainly can't just take the runtime of the last task.

I'm closing this PR. In order to fix this properly, we'll need a function which calculates the total runtime of a job more cleverly - probably add up all the time where at least one of the tasks is running.

mpbrigham:
Hi Luke, thank you for your feedback. Can model jobs also have more than one task?

lukeyeager:
> Can model jobs also have more than one task?

Not currently. I just want a solution that works for all types of jobs.

Issue #741, opened 2016-05-16
Issues with home page angular tables


lukeyeager:
Following up on https://github.com/NVIDIA/DIGITS/pull/728, here are some bug reports / feature requests:

lukeyeager:
BUG: When you click+drag over several entries, you can select them all at once. But if you drag too far, and your "mouse-up" action happens outside of the table, then Angular doesn't catch it. Then it stays stuck in "selection" mode.

lukeyeager:
REQUEST: When you select a dataset (or several), can you filter to only show the models that were trained on that dataset[s]?

lukeyeager:
REQUEST: When a job finishes, move it from the top section down into the appropriate table.

lukeyeager:
REQUEST: update count of jobs based on filters (next to "Datasets" or "Models").

TimZaman:
Oh man we really cannot pull this in in the current state.
There are only around 30 characters available in which one can describe his model, right? How does one distinguish between jobs that have run. At least the datasets's name should be in there, probably some more stuff regarding the settings. This is how it then looks in practice. Indeed i could shorten the dataset name, but then again, **there is too little space for naming and annotating**. My suggestion is to do away with the two column approach; maybe use tabs?

![image](https://cloud.githubusercontent.com/assets/7721540/15316641/2750b47a-1c1e-11e6-8488-590d12867dd9.png)

enoonIT:
I agree with @TimZaman , in this current state it's impossible to distinguish between jobs.
At the very least I'd make the column width resizable on the fly by mouse dragging

TimZaman:
Suggestion with using tabs:
![image](https://cloud.githubusercontent.com/assets/7721540/15323739/b46003d2-1c43-11e6-9d0a-9f116d4e53fa.png)

and 

![image](https://cloud.githubusercontent.com/assets/7721540/15323743/ba018748-1c43-11e6-9911-d8cda0cdd3eb.png)

In the wide way there is a lot of room to add more lovely column-metrics

jmancewicz:
The side by side tables are very tight especially with the width restrictions from bootstrap. The non-name fields are about as small as they can be while still being readable.

I had initially had a tabbed view, but moved to the side by side to attempt to keep close to the origianal look. I can put out a tabbed version to see how that works.

lukeyeager:
BTW, the tooltips added with https://github.com/NVIDIA/DIGITS/pull/748 help with these issues:

[...]

PR #742, opened 2016-05-16
Fix selection bug mentioned in #741

jmancewicz:
Fix selection bug mentioned in #741 where if drag selection finishes with a mouse up outside of the table, that it gets stuck in a mouse down mode.


lukeyeager:
Confirmed. Thanks!

PR #743, opened 2016-05-16
Calculate status and runtime for job from tasks

### LINKED ISSUE #706: Show runtime instead of total time on home page

### LINKED ISSUE #729: Digit's homepage job management reports 'Running' when it's 'Waiting'
Title says it all. Recreate by creating a queue of jobs that exceed the amount of available GPU's. They will all say that they are running, but when you click on them they say 'Waiting'.

lukeyeager:
*Close #706, close #729*
*Fix #709, replace #739*

@pansk @TimZaman @jmancewicz @mpbrigham - Thanks for the bug reports and proposed fixes!

Is everyone happy with this solution? Check the docstring for how `runtime_of_tasks()` is calculated:
```py
    def runtime_of_tasks(self):
        """
        Returns the time (in sec) between when the first task started and when the last task stopped
        NOTE: this may not be what you're expecting if there was some WAIT time in-between
        """
```

### DIFF
--- a/digits/job.py
+++ b/digits/job.py
@@ -158,6 +158,62 @@ class Job(StatusCls):
         """
         raise NotImplementedError('Implement me!')
 
+    def status_of_tasks(self):
+        """
+        Returns the status of the job's tasks
+        """
+        job_status = self.status
+
+        if job_status in [Status.ABORT, Status.ERROR]:
+            return job_status
+
+        task_statuses = [t.status for t in self.tasks]
+
+        important_statuses = [
+            Status(s) for s in [
+                # Sorted by importance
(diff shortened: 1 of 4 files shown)


jmancewicz:
Looks good.

pansk:
Yep, it also reports correct results for existing tasks. Thank you!

PR #744, opened 2016-05-16
Expose iter_size solver option

### LINKED ISSUE #660: Exposing iter_size parameter in DIGITS interface to enable larger training batches

lukeyeager:
*Close #660*
*Replace #675*

Expose `iter_size` solver option (https://github.com/BVLC/caffe/pull/1977) for NVcaffe >= 0.14.

/cc @drozdvadym


digits/templates/models/images/classification/new.html
```diff
@@ -437,6 +438,17 @@ <h4 style="display:inline-block;">Python Layers</h4>
...
+    $('#iter-size-option').hide();
+    $('#iter_size').prop('disabled', true);
+    {% for fw in frameworks %}
+    {% if fw.can_accumulate_gradients() %}
+    if (fwid == "{{ fw.get_id() }}") {
+        $('#iter_size').prop('disabled', false);
+        $('#iter-size-option').show();
+    }
+    {% endif %}
+    {% endfor %}
```
gheinrich: or maybe you can extend the structure on lines 414-416 with a `can_accumulate_gradients` field for a more direct access to framework properties?

lukeyeager: Oh yeah, good call. I also forgot to add this to the `Generic` path ...

digits/model/images/classification/views.py
```diff
@@ -253,23 +253,24 @@ def create():
...
+                        batch_size = form.batch_size.data[0],
+                        batch_accumulation = form.batch_accumulation.data,
+                        val_interval = form.val_interval.data,
+                        pretrained_model = pretrained_model,
+                        crop_size = form.crop_size.data,
+                        use_mean = form.use_mean.data,
+                        network = network,
+                        random_seed = form.random_seed.data,
+                        solver_type = form.solver_type.data,
+                        shuffle = form.shuffle.data,
```
lukeyeager: Sorry for the whitespace changes. It had to be done at some point.

lukeyeager:
Refactored to use the name `batch_accumulation` instead of `iter_size`.

gheinrich:
That looks good to me - I haven't tested let me know if you want me to.

lukeyeager:
Actually it would be great if you could do a quick smoketest. No need to validate convergence, just check that:

1. The option doesn't appear for Torch
2. It does appear for NVcaffe >= 0.14
3. When you set the value to 2, the generated `solver.prototxt` contains the new value

Assuming all is good, please merge. Thanks!

digits/model/tasks/caffe_train.py
```diff
@@ -426,6 +426,11 @@ def save_files_classification(self):
 
         solver.snapshot_prefix = self.snapshot_prefix
 
+        # Batch accumulation
+        from digits.frameworks import CaffeFramework
+        if CaffeFramework().can_accumulate_gradients():
+            solver.iter_size = self.batch_accumulation
+
```
gheinrich: I think you need to make this change in `save_files_generic()` too.

gheinrich:
The classification path is working for me - there is a little omission in the generic path.

digits/model/tasks/caffe_train.py
lukeyeager: d'oh

PR #745, opened 2016-05-16
Deselect with click outside the tables

jmancewicz:
Changing the behavior to deselect the table rows if the mouse is clicked outside of the table with the exception of `input`, `textarea`, `button`, and `a`


lukeyeager:
Works for me!

Unless I mousedown on one element and mouseup on another element - lol. Not really an issue, though.

PR #746, opened 2016-05-16
remove warning when cloning with new form elements

jmancewicz:
When a job is cloned and a newer form has been added to the job creation page, DIGITS warns that the new form element was not cloned.  This is unnecessary and distracting.  Removing.


lukeyeager:
Excellent, thanks!

(discovered while testing https://github.com/NVIDIA/DIGITS/pull/744)

PR #747, opened 2016-05-16
A few small layout changes

jmancewicz:
The tables on the front page were shifted to the right from being in a container.

### DIFF
--- a/digits/templates/home.html
+++ b/digits/templates/home.html
@@ -13,9 +13,9 @@
 {% endblock %}
 
 {% block content %}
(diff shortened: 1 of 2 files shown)


lukeyeager:
Can you add `style:margin-top:20px` to `div.gpus-available` to match the `h2` margin? Looks a little better to me.

jmancewicz:
Added the margin-top.  It does look much better.

jmancewicz:
Also, changed the parenthetical number of jobs to reflect the filtered length.

PR #748, opened 2016-05-16
Add tooltips and round the time values

jmancewicz:
Adding tooltips to the job name in case the overrun and ellipsis appear as well as to the elapsed and submitted times to get a little more information. If the elapsed time is 4m, the tooltip will unpack that a bit to 3 minutes, 43 seconds or 4 minutes, 13 seconds.  To that end, the time functions are changing from flooring to rounding of the least significant value.

### DIFF
--- a/digits/static/js/jquery.time_filters.js
+++ b/digits/static/js/jquery.time_filters.js
@@ -5,10 +5,10 @@ function print_time_diff(diff) {
         return 'Negative Time';
     }
     var total_seconds = Math.floor(diff);
-    var days = Math.floor(total_seconds/(24*3600));
-    var hours = Math.floor((total_seconds % (24*3600))/3600);
-    var minutes = Math.floor((total_seconds % 3600)/60);
-    var seconds = Math.floor(total_seconds % 60);
+    var days = total_seconds/(24*3600);
+    var hours = (total_seconds % (24*3600))/3600;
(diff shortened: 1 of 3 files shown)


lukeyeager:
Tooltips are working for me on Firefox. Can you make them show up for just the links and not the whole `td`?

jmancewicz:
I'd forgotten one. Fixed.

digits/static/js/jquery.time_filters.js
```diff
@@ -22,12 +22,12 @@ function print_time_diff(diff) {
...
-        return pair(days, 'day', hours, 'hour');
+        return pair(Math.floor(days), 'day', Math.round(hours), 'hour');
     else if (hours >= 1)
-        return pair(hours, 'hour', minutes, 'minute');
+        return pair(Math.floor(hours), 'hour', Math.round(minutes), 'minute');
     else if (minutes >= 1)
-        return pair(minutes, 'minute', seconds, 'second');
-    return plural(seconds, 'second');
+        return pair(Math.floor(minutes), 'minute', Math.round(seconds), 'second');
+    return plural(Math.round(seconds), 'second');
```
gheinrich: I should probably stare at this for a little while longer but I'm missing the point behind the change :cold_sweat:

gheinrich:
Works for me - that's very useful, thanks!

PR #749, opened 2016-05-16
Update batch sizes for standard torch networks

lukeyeager:
Make them match Caffe's from https://github.com/NVIDIA/DIGITS/pull/733.

I tested them all on a GTX 980 (4GB card) and none ran out of memory.


lukeyeager:
@gheinrich this is ready for review

gheinrich:
That looks good to me thanks!

PR #750, opened 2016-05-17
Fixed ctrl-click and job count

jmancewicz:
Changed the parenthetical job counts on the front pages to represent the filtered list.
ctrl-click had been opening a new tab AND redirecting the current page. Removed code that is no longer needed for selection.


lukeyeager:
LGTM

Issue #738, opened 2016-05-14
Model Creation UI Discussion & Tweaks


TimZaman:
Hi everyone. I have been working with Digits so much the past couple of months, I decided to take a step back and see how we use it, and was wondering how others actually use it.
First off, the Standard Networks are great to have a initial model up, but then we will always customize the model, and tweak the parameters a bit. This means that our focus here lies on the model.
Currently, the model with us is just messy we use it: we copy it to a (syntax highlighting) text editor, then change it, and paste it back in the textarea.
At first i changed the textarea to monospace. Then the box for the network was still very small to work in. I decided to make it bigger, and while at it, i found the 'Ace' editor with BSD licence: [https://ace.c9.io/ ](https://ace.c9.io/ )

**Is it allowed for you guys (NVIDIA Digits) to incorporate Ace?** It has a very small footprint, and has highlighing for python (caffe models) as well as torch (lua). And incorporating is around 20 LOC.

To answer my own question 'how do i use digits' i show here how i changed up the UI:

![ui_ace_caffe](https://cloud.githubusercontent.com/assets/7721540/15267339/86b28fc6-19bf-11e6-881b-fcd1f615af32.png)
![ui_with_ace](https://cloud.githubusercontent.com/assets/7721540/15267340/86b46e4a-19bf-11e6-9cdf-3bde8a3756cd.png)

lukeyeager:
Neat! Ace looks like a pretty cool product - and it even has protobuf support? As you pointed out, there shouldn't be a problem integrating a BSD-licensed tool.

Since you've already done the integration work, do you want to make a pull request? I like the rearrangement of the page layout too. If you want to include that, keep it in a separate commit so it's easier to review.

lukeyeager:
Looks like this is what @GitHub uses, too:
https://github.com/showcases/projects-that-power-github

TimZaman:
Sure Luke i will _try_ to make a neat PR _tomorrow_.

PR #751, opened 2016-05-17
Ace editor & Model Job UI Tweaks

TimZaman:
Implements #738.

- [x] Added only the minimal amount of minified Ace libraries
- [x] Ace editor syncs with the custom_network flask box upon interaction
- [x] Implement for classification job
- [x] Implement for generic job
- [x] Change the model creation UI to [3 rows; 1 row; 1 row]

Note that for purposes of backwards compatibility and maintaining flask/python integrity, the custom_network container is still in place, and only syncs with the ace editor when it has to. Although the file diffs seem to indicate i did a lot, the ace editor is only four .js files and a few LOC, while the majority of the code change is just moving around the `<div>`'s to tweak the model creation UI.


lukeyeager:
Very cool, great addition! I've tried it and it works great so far, let me start reviewing the code...

1. I need a signed CLA from you before I can accept this: https://github.com/NVIDIA/DIGITS/blob/v3.3.0/.github/CONTRIBUTING.md#pull-requests

1. After addressing any code nitpicks, would you mind implementing this pull request in two separate commits: (1) rearranging the `new.html` templates and (2) adding Ace? That will make `git blame` more useful later, as well as making the commit stream easy to read.

TimZaman:
Sure Luke. Note the CLA link is dead in https://github.com/NVIDIA/DIGITS/blob/master/.github/CONTRIBUTING.md#L33

lukeyeager:
> Note the CLA link is dead in https://github.com/NVIDIA/DIGITS/blob/master/.github/CONTRIBUTING.md#L33

Whoops, good call. https://github.com/NVIDIA/DIGITS/pull/753.

digits/templates/models/images/classification/new.html
```diff
@@ -376,14 +382,16 @@ <h4 style="display:inline-block;">Python Layers</h4>
               )
         .done(function(data) {
                 data = $.parseJSON(data);
-                $('textarea[name=custom_network]').val(data['network']);
+                editor.setValue(data['network']);
+                editor.setValue(editor.getValue(), 1); // Moves cursor to end of editor
```
lukeyeager: Can we move the cursor to the top instead of the end? Did you have a particular reason for wanting to see the end first?

TimZaman: Sorry Luke, I insist we keep it at the bottom. Do you have a particular reason for wanting to see the start first?

No, seriously; I couldn't care less. I did it because by default it selects _all the text_ by default when set text is Set(). Therefore i just put the cursor at the end: for torch this is often the most convenient as the top is often full of inclusions and stuff I rarely touch.

lukeyeager:
For reference, here are some alternate themes available for Ace:
https://ace.c9.io/build/kitchen-sink.html

I don't have a strong opinion on the theme - "Monokai" works for me. @jmancewicz: do you have an opinion?

digits/templates/models/images/classification/new.html
```diff
@@ -647,12 +659,24 @@ <h4 style="display:inline-block;">Python Layers</h4>
                                 <div class="form-group{{mark_errors([form.custom_network])}}">
                                     {{form.custom_network.label}}
                                     {{form.custom_network.explanation(file='/models/images/classification/custom_network_explanation.html')}}
-                                    <div class="pull-right">
-                                        <a href=# class="btn btn-info" onClick="return visualizeNetwork();">Visualize</a>
-                                    </div>
-                                    {{form.custom_network(class='form-control', rows=10)}}
+                                    <div id="divCheckbox" style="display: none;">{{form.custom_network(class='form-control', rows=10)}}</div>
```
jmancewicz: I was going to suggest that `divCheckbox` be `div_checkbox` or `div-checkbox` as long as it was new, but then saw that it wasn't being used. Is there a reason to id that div?

digits/templates/models/images/classification/new.html
jmancewicz: Opening a text document at the tail is nonstandard. Selecting all the text is really nonstandard. I suggest having it not select any text, and open at the head for this PR, and if you prefer, change it to open at the tail in your repo.

digits/templates/models/images/classification/new.html
```diff
@@ -647,12 +659,24 @@ <h4 style="display:inline-block;">Python Layers</h4>
...
+                                    <a href=# class="btn btn-info" onClick="return visualizeNetwork();">Visualize</a>
                                 </div>
                         </div>
+                        <div id="ace-editor"></div>
+                        <script>
+                            var editor = ace.edit("ace-editor");
+                            editor.setTheme("ace/theme/monokai");
```
jmancewicz: I prefer the dark theme personally, but think that Chrome might be more in keeping with the color theme in DIGITS. Just needed to put that out there.

TimZaman: I feel exactly the same: prefer the dark theme, but the Chrome theme is more in the color theme. Do we go for  Beauty or the Arts, your call? :)

digits/templates/models/images/classification/new.html
TimZaman: Ah that's a shameful copy-paste error. Well spotted.

digits/templates/models/images/classification/new.html
lukeyeager: If you don't mind, let's go with Chrome for now. It's easy to swap out in your fork, and it's pretty easy to change our mind later if we want to. Cool?

TimZaman: Sure, chief. Do you want me to include the .js of monokai, so someone can swap it out with a one-liner?

lukeyeager: Let's not. Even though they're super tiny files, I don't think it makes sense to check in files that don't get used.

lukeyeager:
Based on [their license](https://github.com/ajaxorg/ace/blob/master/LICENSE), can you add these lines to the top of those minimized JS files?
```js
// Copyright (c) 2010, Ajax.org B.V.
// All rights reserved.
```

TimZaman:
- [x] Cursor Position To Start
- [x] Theme to Chrome .js, remove monokai js
- [x] Append Ajax BV licence header to minified ace files 
- [x] div id removal
- [x] squash commits and rebase to two for (1) template rearranging and (2) Ace-related.

All set and ready for review.

lukeyeager:
> div id removal

Technically you didn't do this one, but who cares.

I still need your CLA before I can merge, but otherwise this looks great! I like the Chrome theme, it seems to fit in pretty well.

TimZaman:
CLA is on the other computer, keep forgetting to send it. Actually i did do the id removal but only for the generic and not the classification. In fact i encounter those minor 'mistakes' more often wrt the classification/generic files.

lukeyeager:
> In fact i encounter those minor 'mistakes' more often wrt the classification/generic files.

I know, it's a pain. Fortunately @gheinrich is helping us forge ahead towards a nice, general solution that will let us de-duplicate a lot of that code eventually.

TimZaman:
> > div id removal

> Technically you didn't do this one, but who cares.

Fixed.

lukeyeager:
CLA received. Looks great. Thanks for the contribution!

PR #754, opened 2016-05-17
Move Datasets and Models into tab view

jmancewicz:
This is an option for dealing with the length of job names on the front page. The Datasets and Models are in tabs. One upside of not putting them in tabs is the potential to show to dependencies between models and datasets.


gheinrich:
I don't understand all the ramifications of the code change but this is working for me.

jmancewicz:
It just changes the col-md-6 divs to full width divs which are only displayed when it's tab is set to display. It turns out to be a small change.  The controller was left over from when I was testing multple versions in the same page.

PR #756, opened 2016-05-18
View extension framework

gheinrich:
This implements a visualization framework along the lines of:

data-classes

Depends on #731 

A `rawData` view extension is added and selected by default to show the network output in the same way as in the original generic type of networks.

### DIFF
--- a/digits/extensions/view/__init__.py
+++ b/digits/extensions/view/__init__.py
@@ -1,0 +1,35 @@
+# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
+from __future__ import absolute_import
+
+from . import rawData
+
+view_extensions = [
+    # set show=True if extension should be listed in known extensions
+    {'class': rawData.Visualization, 'show': True},
+]
+
+
+def get_default_extension():
+    """
+    return the default view extension
+    """
+    return rawData.Visualization
+
+
+def get_extensions():
+    """
+    return set of data data extensions
+    """
+    return [extension['class'] for extension
+            in view_extensions if extension['show']]
+
+
+def get_extension(extension_id):
+    """
+    return extension associated with specified extension_id
+    """
+    for extension in view_extensions:
+        extension_class = extension['class']
+        if extension_class.get_id() == extension_id:
+            return extension_class
+    return None
--- a/digits/extensions/view/rawData/config_template.html
+++ b/digits/extensions/view/rawData/config_template.html
@@ -1,0 +1,7 @@
+{# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved. #}
+
+{% from "helper.html" import print_flashes %}
+{% from "helper.html" import print_errors %}
+{% from "helper.html" import mark_errors %}
+
+<small>This visualization has no configuration options</small>
--- a/digits/model/images/generic/test_views.py
+++ b/digits/model/images/generic/test_views.py
@@ -23,6 +23,7 @@ import mock
 import PIL.Image
 from urlparse import urlparse
 
+from digits import extensions
 from digits.config import config_value
 import digits.dataset.images.generic.test_views
 import digits.test_views
@@ -239,6 +240,11 @@ class BaseTestViews(BaseViewsTest):
 
     def test_nonexistent_model(self):
         assert not self.model_exists('foo'), "model shouldn't exist"
+
+    def test_view_config(self):
+        extension = extensions.view.get_default_extension()
+        rv = self.app.get('/models/view-config/%s' % extension.get_id())
+        assert rv.status_code == 200, 'page load failed with %s' % rv.status_code
 
     def test_visualize_network(self):
         rv = self.app.post('/models/visualize-network?framework='+self.FRAMEWORK,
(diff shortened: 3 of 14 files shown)


lukeyeager:
I tested the gradient workflow and it's working well for me. But isn't it supposed to draw a red arrow in the direction of the gradient when you test an image? I thought I saw a screenshot from you that did that. Maybe I'll find out what I'm doing wrong when I dig into the code review ...

digits/extensions/view/__init__.py
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
+from __future__ import absolute_import
+
+from . import rawData
+
+view_extensions = [
+    # set show=True if extension should be listed in known extensions
+    {'class': rawData.Visualization, 'show': True},
```
lukeyeager: Seems like we should also show the gradient datasets if we're showing the models

digits/extensions/view/rawData/config_template.html
```diff
@@ -0,0 +1,10 @@
+{# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved. #}
+
+{% from "helper.html" import print_flashes %}
+{% from "helper.html" import print_errors %}
+{% from "helper.html" import mark_errors %}
+
+<small>This visualization has not configuration options</small>
```
lukeyeager: how about "does not have" or "has no"

digits/model/images/generic/test_views.py
```diff
@@ -595,9 +601,15 @@ def test_infer_many_from_folder(self):
...
+        # try selecting the extension explicitly
+        extension = extensions.view.get_default_extension()
+        extension_id = extension.get_id()
+
         rv = self.app.post(
                 '/models/images/generic/infer_many?job_id=%s' % self.model_id,
-                data = {'image_list': file_upload, 'image_folder': os.path.dirname(self.test_image)}
+                data = {'image_list': file_upload,
+                        'image_folder': os.path.dirname(self.test_image),
+                        'view_extension_id': extension_id}
```
lukeyeager: Is this going to break the JSON api? What do I need to provide for a classification network? What happens if I forget to add the new key?

lukeyeager:
Our new URLs seem a little weird:
```sh
# classification (old)
/datasets/images/classification/new
/models/images/classification/new
# generic (old)
/datasets/images/generic/new
/models/images/generic/new
# gradients (new)
/datasets/generic/new/image-gradients
/models/images/generic/new?extension_id=image-gradients
```
Why the inconsistency? I like the move to `/[datasets|models]/generic/new`, why doesn't that work for models? And why does one specify the extension in the query string but the other doesn't?

digits/extensions/view/__init__.py
gheinrich: That's the rawData visualization - it's just showing the raw data. To see the little arrows you need to select the gradient visualization (which isn't in this PR - it's in https://github.com/NVIDIA/DIGITS/pull/735/commits/e8001f863c93ddceae6875362182399e60f8af75 if you're curious to try). The screenshot was on the PR description for #735 - sorry it's confusing!)

digits/model/images/generic/test_views.py
gheinrich: If you don't specify a view extension then the default one (raw data) is used (see [there](https://github.com/gheinrich/DIGITS/blob/747f462ffe08fd63d6a62375c50cca2ac70166e0/digits/extensions/view/__init__.py#L12) ).

gheinrich:
>  I like the move to /[datasets|models]/generic/new, why doesn't that work for models?

The generic datasets now support any 3D data (barring bugs). I initially wanted to add generic models too to handle any 3D data but realized that it would be a huge change. We'd have to change a lot of the training and inference paths. So I thought as a first step we could use the existing image/generic models with both the image/generic datasets and the generic datasets.

> And why does one specify the extension in the query string but the other doesn't?

No strong reason actually - I didn't know how to make the `inference_id` parameter optional when using URLs in the form of `/models/images/generic/new/<extension_id>`.

digits/extensions/view/rawData/config_template.html
gheinrich: Sure! Which of these sounds best to you?

digits/extensions/view/__init__.py
lukeyeager: Whoops, my bad. I got confused.

digits/extensions/view/rawData/config_template.html
lukeyeager: Probably just remove the "t"

digits/model/images/generic/test_views.py
lukeyeager: Cool.

lukeyeager:
Ok, I figured you had a good reason for not creating `/models/generic/`. I agree we don't need to do all that work now, but I look forward to being able to do it eventually and to convert all datasets to `/datasets/generic` and all models to `/models/generic`, or whatever we end up calling them.

> I didn't know how to make the inference_id parameter optional when using URLs in the form of /models/images/generic/new/<extension_id>.

This oughta work:
http://stackoverflow.com/questions/14032066/flask-optional-url-parameters

gheinrich:
> And why does one specify the extension in the query string but the other doesn't?

I have updated the "Generic datasets" commit to fix this. For image gradients, routes now look like:

```
# gradients (new)
/datasets/generic/new/image-gradients
/models/images/generic/new/image-gradients
```

If the extension ID is unspecified in the model creation URL, then this defaults to the "other" kind of datasets.

PR #758, opened 2016-05-19
Remove image file created during infer_one

gheinrich:
Somehow this was forgotten in #657


lukeyeager:
D'oh! This code duplication is killing us. Thanks for the fix.

PR #762, opened 2016-05-19
Return proper HTTP status codes after inference

lukeyeager:
Otherwise, the test suite can miss failed tests:
https://github.com/NVIDIA/DIGITS/blob/v3.3.0/digits/model/images/generic/test_views.py#L348


lukeyeager:
@gheinrich I think this was your code originally. Can you review and merge?

That's assuming it won't interfere with your object detection PRs. If so, this can wait.

gheinrich:
This looks good thanks.

PR #763, opened 2016-05-20
More support for generic all in one nets

lukeyeager:
Do a little work to enable all-in-one nets in the generic path. See the added test for the type of network this enables.


lukeyeager:
Oops, the caffe build on TravisCI doesn't know about `Input` layers yet.

PR #765, opened 2016-05-20
Retrieve caffe version from caffe.exe.

IsaacYangSLA:
Try to retrieve caffe version from command line option '-version' and if caffe.exe is not built with such information, fallback to default (0,11,0).

### DIFF
--- a/digits/config/caffe_option.py
+++ b/digits/config/caffe_option.py
@@ -192,8 +192,25 @@ class CaffeOption(config_option.FrameworkOption):
             # XXX: guess and let the user figure out errors later
             return parse_version(0,11,0)
         elif platform.system() == 'Windows':
-            # XXX: guess and let the user figure out errors later
-            return parse_version(0,11,0)
+            p = subprocess.Popen([executable,'-version'],
+                    stdout = subprocess.PIPE,
+                    stderr = subprocess.PIPE)
+            if p.wait():
+                raise config_option.BadValue(p.stderr.read().strip())
+            else:
+                pattern = 'version'
+                parsed_version = parse_version(0,11,0)
+                for line in p.stdout:
+                    if pattern in line:
+                        try:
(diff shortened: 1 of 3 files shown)


digits/templates/layout.html
```diff
@@ -51,7 +51,8 @@
                 {% if server_name %}
                 <li><span class="navbar-text">Server: {{server_name}}</span></li>
                 {% endif %}
-                <li><span class="navbar-text">Version: {{server_version}}</span></li>
+                <li><span class="navbar-text">Server ver.: {{server_version}}</span></li>
+				<li><span class="navbar-text">Caffe ver.: {{caffe_version}}</span><li>
```
lukeyeager: Spacing looks weird here

digits/config/caffe_option.py
```diff
@@ -192,8 +192,25 @@ def get_version(executable):
...
+                raise config_option.BadValue(p.stderr.read().strip())
+            else:
+                pattern = 'version'
+                parsed_version = parse_version(0,11,0)
+                for line in p.stdout:
+                    if pattern in line:
+                        try:
+                            # if CAFFE for Windows is built with properly formatted version string
+                            parsed_version = parse_version(line[line.find(pattern)+len(pattern):].rstrip())
+                        except:
```
lukeyeager: Please don't ever use a bare `except`/`pass`. If you're expecting a particular type of error, catch it like:
```py
except ValueError:
  pass
```

digits/templates/layout.html
lukeyeager: ^ I meant the whitespace in the code.

But also, it looks kinda cramped and unreadable right now. How about this instead?

caffe-version-ui

IsaacYangSLA:
The suggested Info menu layout is much better.  I changed that file accordingly.  Like the following screen shot.
image

PR #766, opened 2016-05-20
TravisCI overhaul

lukeyeager:
Brings the build+test time down from ~42 minutes to ~22 minutes. About 10 minutes of the difference comes from the improved Torch build time - they build a lot of dependencies from source.


gheinrich:
Great!

PR #767, opened 2016-05-21
Related Jobs in Dataset and Model view

TimZaman:
See #667 , revised version of #668.
See below for screenshots. Very straightforward.

Ready for review @lukeyeager (https://github.com/NVIDIA/DIGITS/issues/667#issuecomment-220705017)

### Dataset View:
image

### Model View:
image

### DIFF
--- a/digits/dataset/views.py
+++ b/digits/dataset/views.py
@@ -24,15 +24,17 @@ def show(job_id):
     if job is None:
         raise werkzeug.exceptions.NotFound('Job not found')
 
+    related_jobs = scheduler.get_related_jobs(job)
+
     if request_wants_json():
         return flask.jsonify(job.json_dict(True))
     else:
         if isinstance(job, dataset_images.ImageClassificationDatasetJob):
-            return dataset_images.classification.views.show(job)
+            return dataset_images.classification.views.show(job, related_jobs=related_jobs)
         elif isinstance(job, dataset_images.GenericImageDatasetJob):
-            return dataset_images.generic.views.show(job)
+            return dataset_images.generic.views.show(job, related_jobs=related_jobs)
         elif isinstance(job, generic.GenericDatasetJob):
-            return generic.views.show(job)
+            return generic.views.show(job, related_jobs=related_jobs)
         else:
             raise werkzeug.exceptions.BadRequest('Invalid job type')
 
--- a/digits/scheduler.py
+++ b/digits/scheduler.py
@@ -214,6 +214,29 @@ class Scheduler:
             return None
         return self.jobs.get(job_id, None)
 
+    def get_related_jobs(self, job):
+        """
+        Look through self.jobs to try to find the Jobs
+        whose parent contains job
+        """
+        related_jobs = []
+
+        if isinstance(job, ModelJob):
+            datajob = job.dataset
+            related_jobs.append(datajob)
+        elif isinstance(job, DatasetJob):
+            datajob = job
+        else:
+            raise ValueError("Unhandled job type %s" % job.job_type())
+
+        for j in self.jobs.values():          
+            # Any model that shares (this/the same) dataset should be added too:
+            if isinstance(j, ModelJob):
+                if datajob == j.train_task().dataset:
+                    related_jobs.append(j)
+
+        return related_jobs
+
     def abort_job(self, job_id):
         """
         Aborts a running Job
--- a/digits/templates/job.html
+++ b/digits/templates/job.html
@@ -235,6 +235,53 @@ $('#delete-job').on('click', function(event) {
         </div>
         <div class="gpu-utilization-info" style="display:none;"></div>
 
+
+        <!-- Related jobs -->
+        {% if related_jobs is not none and related_jobs|length > 0 %}
+        <div class="panel panel-default">
+            <div class="panel-heading">
+                <h4 class="panel-title">
+                    Related jobs
+                </h4>
+            </div>
+            <div class="panel-collapse">
+                <div class="panel-body">
+                    <div class="panel-group">
+                        <!-- Related model jobs -->
+                        {% for r_job in related_jobs %}
+                        {% if prev_job_model != r_job.job_type() %}
+                            <h4 class="panel-title" style="margin:10px">{{r_job.job_type()}}</h4>
+                            {% set prev_job_model = r_job.job_type() %}
+                        {% endif %}
+                        <div id="{{r_job.id()}}-status" class="panel panel-{{ r_job.status.css }}">
+                            <div class="panel-heading">
+                                <h4 class="panel-title">
+                                    <a href="{{ url_for('digits.views.show_job', job_id=r_job.id()) }}">
+                                        {{ r_job.name() }}
+                                        <small>{{r_job.status.name}}</small>
+                                    </a>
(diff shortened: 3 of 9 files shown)


digits/dataset/views.py
```diff
@@ -24,15 +24,17 @@ def show(job_id):
...
     else:
         if isinstance(job, dataset_images.ImageClassificationDatasetJob):
-            return dataset_images.classification.views.show(job)
+            return dataset_images.classification.views.show(job, related_jobs=related_jobs)
         elif isinstance(job, dataset_images.GenericImageDatasetJob):
-            return dataset_images.generic.views.show(job)
+            return dataset_images.generic.views.show(job, related_jobs=related_jobs)
         elif isinstance(job, generic.GenericDatasetJob):
-            return generic.views.show(job)
+            return generic.views.show(job, related_jobs=related_jobs)
```
gheinrich: I think you forgot to add this parameter to the `show()` function in `digits/dataset/generic/views.py`

digits/model/images/classification/job.py
```diff
@@ -57,3 +57,6 @@ def download_files(self, epoch=-1):
 
         return download_files
 
+    @override
+    def parent_job(self):
+        return [self.train_task().dataset]
```
gheinrich: why don't you move this up one level in the class hierarchy i.e. in `digits/model/images/job.py`? That would save the hassle of duplicating the implementation in `digits/model/generic/job.py`

digits/scheduler.py
```diff
@@ -214,6 +214,28 @@ def get_job(self, job_id):
...
         return self.jobs.get(job_id, None)
 
+    def get_related_jobs(self, job):
+        """
+        Look through self.jobs to try to find the Jobs
+        whose parent contains job
+        """
+        related_jobs = []
+
+        if 'Model' in job.job_type():
```
gheinrich: why don't you do:
`if isinstance(job, ModelJob)`

digits/scheduler.py
```diff
@@ -214,6 +214,28 @@ def get_job(self, job_id):
...
+            datajob = job.parent_job()[0]
```
gheinrich: I think this is the only reference to `parent_job()` and it's only useful if the job is a `ModelJob` in which case the `dataset` attribute is returned. Wouldn't it be simpler to remove this `parent_job()` method altogether and do something like:
```
if isinstance(job, ModelJob):
    datajob = job.dataset
elif isinstance(job, DatasetJob):
    datajob = job
else:
    raise ValueError("Unhandled job type %s" % job.job_type())
```

digits/model/images/classification/job.py
TimZaman: +. Dang you're very sharp for a sunday.

digits/dataset/views.py
TimZaman: Ah that's because the file `digits/dataset/generic/views.py` is brand new. Didn't catch that, well spotted.

TimZaman:
Excellent points, I reduced down the complexity to just one function. Looks good to me.
Ready for re-review, @gheinrich .

gheinrich:
> ... Digitizers.

Wow. Nice! :v:

digits/scheduler.py
```diff
@@ -214,6 +214,34 @@ def get_job(self, job_id):
...
 
+    def get_related_jobs(self, job):
+        """
+        Look through self.jobs to try to find the Jobs
+        whose parent contains job
+        """
+        related_jobs = []
+
+        if isinstance(job, ModelJob):
+            datajob = job.dataset
```
gheinrich: Thanks for the updates! If you don't mind a couple more comments... there you can do:
```
related_jobs.append(datajob)
```
so you can take lines 233-235 out.

digits/scheduler.py
```diff
@@ -214,6 +214,34 @@ def get_job(self, job_id):
...
+        elif isinstance(job, DatasetJob):
+            datajob = job
+        else:
+            raise ValueError("Unhandled job type %s" % job.job_type())
+
+        for j in self.jobs:
```
gheinrich: you can iterate over dictionary values:
```
for j in self.jobs.values():
```
That would be slightly more pythonic and you can replace `self.jobs[j]` with just `j` in the lines below

digits/scheduler.py
```diff
@@ -214,6 +214,34 @@ def get_job(self, job_id):
...
+            # If our request came from a model, just add the dataset:
+            if isinstance(job, ModelJob):
+                if datajob == self.jobs[j]:
+                    related_jobs.append(self.jobs[j]) # Add the dataset
+            
+            # Any model that shares (this/the same) dataset should be added too:
+            if isinstance(self.jobs[j], ModelJob):
+                if datajob == self.jobs[j].train_task().dataset:
+                    print(self.jobs[j].train_task().dataset)
```
gheinrich: also maybe you don't have to print this

TimZaman:
Keep going like this and we will run out of lines of code. Squashed yet again plz review @gheinrich .

digits/templates/job.html
```diff
@@ -235,6 +235,53 @@ <h4 class="panel-title">
...
+                <h4 class="panel-title">
+                    Related jobs
+                </h4>
+            </div>
+            <div class="panel-collapse">
+                <div class="panel-body">
+                    <div class="panel-group">
+                        <!-- Related model jobs -->
+                        {% for r_job in related_jobs %}
+                        {% if prev_job_model != r_job.job_type() %}
```
gheinrich: it might have been more explicit to call this `prev_job_type` but I don't think it's a big deal

digits/templates/job.html
```diff
@@ -235,6 +235,53 @@ <h4 class="panel-title">
...
+                            <h4 class="panel-title" style="margin:10px">{{r_job.job_type()}}</h4>
+                            {% set prev_job_model = r_job.job_type() %}
+                        {% endif %}
+                        <div id="{{r_job.id()}}-status" class="panel panel-{{ r_job.status.css }}">
+                            <div class="panel-heading">
+                                <h4 class="panel-title">
+                                    <a target="_blank" href="/jobs/{{ r_job.id() }}">
```
gheinrich: Something like this could have worked too:
```
<a href="{{ url_for('digits.views.show_job', job_id=r_job.id()) }}">
```
I am not sure which is better... I tend to use this form. I feel it's more resilient to future changes.

gheinrich:
Great PR, very useful! I tested it (also on some "generic" data jobs) and it works well. Thanks!

TimZaman:
Perfect, thanks for the review.

digits/templates/job.html
lukeyeager: I agree, I like that form better. I think/hope we're pretty consistent with that since https://github.com/NVIDIA/DIGITS/pull/507.

TimZaman:
Okay, updated the line. Want me to squash?

lukeyeager:
> Okay, updated the line. Want me to squash?

Yep, looks good to me!

TimZaman:
Squashed.
Maybe for another time, and depending on how you guys like this in practice (we use this a _lot_); it might even be better to put in a little @jmancewicz -style jquery table.

lukeyeager:
It's kinda weird that the current model shows up below as a "related" job, no? I can't see why I would ever click on that.

PR #768, opened 2016-05-21
Digits 3.0

helio-silva:
Device_query_py is not detecting my Nvidia hardware.
I had to replace it for old version from Digits 2.0 branch to be able to use GPU processing.

platform OSX 10.10.5.


gheinrich:
Hi @helio-silva did you submit this PR by accident? You are trying to merge commits from `digits-3.0` to `master` yet those commits are already present on `master`.

We'll have to debug why the newer `device_query.py` isn't working on MacOS but please bear in mind that this isn't an officially supported platform. Are you willing to help us identify the root cause for the regression? We can't fix this by merely reverting back to the older version of `device_query.py`.

PR #769, opened 2016-05-22
Allow BVLC/caffe and caffe-on-windows

IsaacYangSLA:
1. parse_version may return ('00000001','*','00000003','*final'), which is hard to to convert back to '1.0.0-rc3.'  Therefore, storing the original version string as a backup for Info menu.

2. Flavor will be NVIDIA if libcaffe.so is built with NVIDIA enhancement (Linux only).  In other cases, flavor is BVLC.

3. Store information in CaffeOption class so no need to query caffe for linked libraries twice.

### DIFF
--- a/digits/config/caffe_option.py
+++ b/digits/config/caffe_option.py
@@ -12,8 +12,10 @@ from . import config_option
 from . import prompt
 from digits import device_query
 from digits.utils import parse_version
+from digits.utils.errors import UnsupportedPlatformError
 
 class CaffeOption(config_option.FrameworkOption):
+
     @staticmethod
     def config_file_key():
         return 'caffe_root'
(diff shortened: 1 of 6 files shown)


digits/utils/errors.py
```diff
@@ -18,3 +18,8 @@ class LoadImageError(DigitsError):
     """
     pass
 
+class UnsupportedPlatformError(DigitsError):
+    """
+    Errors that occur while performing tasks in unsupported platforms
+    """
+    pass
```
TimZaman: :trollface:

digits/config/caffe_option.py
```diff
@@ -12,8 +12,12 @@
 from . import prompt
 from digits import device_query
 from digits.utils import parse_version
+from digits.utils.errors import UnsupportedPlatformError
 
 class CaffeOption(config_option.FrameworkOption):
+    version_string = None
+    flavor = None
```
lukeyeager: Let's not mix instance attributes and class attributes. `wtforms` does that and it's super confusing.

digits/config/caffe_option.py
```diff
@@ -141,78 +145,143 @@ def validate_version(cls, executable):
...
-                        except ValueError:
-                            # what follows "version" is "CAFFE_VERSION" or some unrecognized string
-                            pass
-                        finally:
-                            break
-                return parsed_version
+                    version_string = match.group(2)
+            return version_string
+        else:
+            raise UnsupportedPlatformError('Linked library in Linux only')
```
lukeyeager: That's not exactly true, is it? We could use `otool` on OSX, at least. How about: "Parsing of shared library version not supported on this platform"?

lukeyeager:
I've tested this and it lets me use either BVLC/caffe or NVIDIA/caffe in DIGITS - cool!

IsaacYangSLA:
I still have to keep Microsoft Caffe as one 'special case' because caffe.exe does not return version info currently.  And linked library approach can't apply to Windows platform.  Without such version info, validate_version raises exception.

This PR (for Windows Caffe to report version, [https://github.com/BVLC/caffe/pull/4200](url)) was just created on BVLC/Windows branch.  The same one was created on Microsoft's Caffe repository, but it looked like the PR should not go there.  Therefore, that one was closed and I created the new one on BVLC.

bygreencn:
about parse_version may return ('00000001','','00000003','final'), i had it return with the same sting as on Linux, you may check https://github.com/bygreencn/caffe/commit/38ff23844af9b38bcff3c3800e83bd8a96cea1ff

digits/config/caffe_option.py
```diff
@@ -126,94 +128,137 @@ def validate_version(cls, executable):
         executable -- path to a caffe executable
         """
         minimum_version = parse_version(0,11,0)
-        version = cls.get_version(executable)
+        info_dict = cls.get_info(executable)
 
-        if version is None:
+        if info_dict['ver_str'] is None:
             raise config_option.BadValue('Could not get version information from caffe at "%s". Are you using the NVIDIA fork?'
```
lukeyeager: Remove this comment about the NVIDIA fork - it's not required anymore

digits/config/caffe_option.py
```diff
@@ -126,94 +128,137 @@ def validate_version(cls, executable):
         executable -- path to a caffe executable
         """
         minimum_version = parse_version(0,11,0)
```
lukeyeager: I thought we agreed we would need different minimum versions for different flavors?

digits/config/caffe_option.py
IsaacYangSLA: No problem.

IsaacYangSLA:
Modified the validate_version function.  When no ver_str, raising an exception.  Minimum version also depends on flavor of Caffe.

lukeyeager:
Even though these checks currently work, they're misleading. Can you update them to only make the version check if the flavor is NVIDIA? BVLC/caffe > 1.0.0-rc3 (the minimum) supports all these features. 

https://github.com/NVIDIA/DIGITS/blob/c0862e1611/digits/frameworks/caffe_framework.py#L37
https://github.com/NVIDIA/DIGITS/blob/c0862e1611/digits/frameworks/caffe_framework.py#L127
https://github.com/NVIDIA/DIGITS/blob/c0862e1611/digits/model/tasks/caffe_train.py#L800

lukeyeager:
Tested and everything looks right:

bvlc@rc2: fails
bvlc@rc3: succeeds
nvidia@0.10.0: fails
nvidia@0.11.1: succeeds
nvidia@0.12.2: succeeds
nvidia@0.13.2: succeeds
nvidia@0.14.5: succeeds
nvidia@0.15.1: succeeds

lukeyeager:
Neat - I tried a parallel build on Travis for NVIDIA/caffe and BVLC/caffe and they both work great:
https://travis-ci.org/lukeyeager/DIGITS/builds/134811320

@IsaacYangSLA please address https://github.com/NVIDIA/DIGITS/pull/769#issuecomment-223157063 before merge

IsaacYangSLA:
Thanks.  There is a commit immediately below that comment.  It changed the behavior of 'validate_version' so it only checks minimum version on NVIDIA flavor.  Sorry that I piggybacked it with removal of Windows version workaround.

lukeyeager:
Now that https://github.com/BVLC/caffe/pull/4200 is merged, this should work for the `BVLC/caffe/windows` branch, too.

lukeyeager:
Code looks good and seems to be working well - thanks for the good work!

Please squash and I'll merge.

digits/utils/errors.py
lukeyeager: Fixed

PR #773, opened 2016-05-24
Only allow entering absolute path to textfields with autocomplete

Lucaszw:
Close #586

fixed white-space in view

fixed if statement problem

### DIFF
--- a/digits/views.py
+++ b/digits/views.py
@@ -5,6 +5,7 @@ import glob
 import json
 import platform
 import traceback
+import os
 
 import flask
 from flask.ext.socketio import join_room, leave_room
@@ -467,6 +468,11 @@ def path_autocomplete():
 
     """
     path = flask.request.args.get('query','')
+
+    if not os.path.isabs(path) :
+        # Only allow absolute paths by prepending forward slash
+        path = os.path.sep + path
+
     suggestions = glob.glob(path+"*")
     if platform.system() == 'Windows':
         # on windows, convert backslashes with forward slashes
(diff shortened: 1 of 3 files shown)


digits/views.py
```diff
@@ -435,6 +436,11 @@ def path_autocomplete():
 
     """
     path = flask.request.args.get('query','')
+
+    if not os.path.isabs(path) :
+        # Only allow absolute paths by appending forward slash
+        path = '/' + path
```
lukeyeager: @IsaacYangSLA can you try this on Windows? Maybe something like this would be better:
```py
path = os.path.sep + path
```

lukeyeager:
It's a little surprising to have a slash appended to my path for me, but it's also pretty clear why it's being done.
@jmancewicz what do you think? Too astonishing? Useful?

Lucaszw:
Close #586

Lucaszw:
@lukeyeager I found that appending the backslash helps avoids confusion as to whether the component is active or not. Ie someone may start writing a relative path, and just assume the the textfield isn't intelligent enough to validate the path (instead of thinking that relative paths are not allowed).

digits/views.py
IsaacYangSLA: Sure.

digits/views.py
```diff
@@ -435,6 +436,11 @@ def path_autocomplete():
 
     """
     path = flask.request.args.get('query','')
+
+    if not os.path.isabs(path) :
+        # Only allow absolute paths by appending forward slash
```
gheinrich: technically, isn't it more correct to say "prepending"?

Lucaszw:
I updated prepend statement and comment

lukeyeager:
Looks good and works great! Squash and I'll merge.

PR #776, opened 2016-05-24
Fix crop length assertion during Torch inference

gheinrich:
Issue reported by Rombout Karelse on DIGITS users list.


gheinrich:
Tested by Rombout Karelse

PR #781, opened 2016-05-24
Tail function may not exist in certain platforms.  Implementing an al…

IsaacYangSLA:
…ternative in python for those platforms.

Tested on a log file (~60K lines, caffe_output.log).
Tail took about 3ms to return.
Deque took about 17ms.
The performance hit seemed acceptable.

### DIFF
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -21,6 +21,7 @@ from digits import utils
 from digits.config import config_value
 from digits.status import Status
 from digits.utils import subclass, override, constants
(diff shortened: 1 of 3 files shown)


lukeyeager:
1. Please limit the first line of your commit messages to 50 characters: https://github.com/blog/926-shiny-new-commit-styles

2. I like the idea, but can we simply name the function `tail` and put it in `utils.filesystem`?

IsaacYangSLA:
Sure.  Thanks for the suggestion.

lukeyeager:
Thanks for the platform-specific fix!

PR #782, opened 2016-05-25
Don't show self in related jobs

lukeyeager:
Follow-up from https://github.com/NVIDIA/DIGITS/pull/767.

See https://github.com/NVIDIA/DIGITS/pull/767#issuecomment-221353355


TimZaman:
My bad.

gheinrich:
Works for me thanks.

PR #785, opened 2016-05-25
Config option for extensions

gheinrich:
Add config options to selectively enable extensions.
Default value comes from code (extension init module).

### DIFF
--- a/digits/config/current_config.py
+++ b/digits/config/current_config.py
@@ -2,6 +2,7 @@
 from __future__ import absolute_import
 
 from .caffe_option import CaffeOption
+from .extension_list import DataExtensionListOption, ViewExtensionListOption
 from .gpu_list import GpuListOption
 from .jobs_dir import JobsDirOption
 from .log_file import LogFileOption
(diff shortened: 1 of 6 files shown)


gheinrich:
Now that #756 is merged (:tada:) let me just update this patch to cover the view extensions too.

lukeyeager:
Very nice.

lukeyeager:
@gheinrich I'm not super pumped about how this works. If I've made any configuration choices and a `digits.cfg` has been generated, then my choice is saved and I don't get any new extensions that are added later (like with https://github.com/NVIDIA/DIGITS/pull/830, for example). So I have to go and reconfigure DIGITS each time to pick up the new extensions.

Should we just disable this configuration option and allow all extensions?

gheinrich:
The config option was mainly intended at making the "gradients" view and data extensions invisible by default. So what we can do is:
- review & merge #927 (support for plug-ins)
- in a single PR:
 - remove the config options for extensions (make them all visible)
 - move the "gradients" data & view extensions to the plug-in area (like the example on #921)
 - update the image regression tutorial to let users know how to install the "gradients" plug-ins

Would that work?

lukeyeager:
So users would have to manually `pip install` each new plugin?

gheinrich:
Not really, they would have to `pip install` only those that are distributed as plug-ins. The "built-in" extensions from `$DIGITS_HOME\digits\extensions` would be there by default.

PR #788, opened 2016-05-25
Cache Caffe and Torch builds in Travis for 1 week

lukeyeager:
By removing the build time entirely, this brings the total TravisCI build+test time down from ~20 minutes to ~12 minutes when a cached build is present.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,13 @@ env:
     global:
         - NUM_THREADS=4
 
+cache:
+    timeout: 604800  # 1 week
+    apt: true
+    directories:
+        - ~/caffe
+        - ~/torch
+
 before_install:
     # undo all of TravisCI's "helpful" Python shenanigans on 14.04
     - deactivate
--- a/scripts/travis/install-caffe.sh
+++ b/scripts/travis/install-caffe.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # Copyright (c) 2015-2016, NVIDIA CORPORATION.  All rights reserved.
 
 set -e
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -11,6 +11,13 @@ env:
     global:
         - NUM_THREADS=4
 
+cache:
+    timeout: 10080 # 1 week
```
gheinrich: according to https://docs.travis-ci.com/user/caching/ the `timeout` is expressed in seconds and `10080s=2.8h`. Am I not looking at the right doc?

lukeyeager: Haha whoops, sloppy math. Thanks!

gheinrich:
How do we handle the case where one build is started just before the timeout? And the case when the cache is empty and two jobs are started within minutes of each other? Maybe we don't want to handle these cases. Just curious.

lukeyeager:
> How do we handle the case where one build is started just before the timeout? And the case when the cache is empty and two jobs are started within minutes of each other? Maybe we don't want to handle these cases. Just curious.

From reading through the documentation, the cache is checked at the start of the build and uploaded at the end of the build (makes sense). So, if the cache is empty and two jobs start within a few minutes of each other, they'll both build and both upload a new cache (probably identical caches). It's a little wasteful, but I think it'll shake out in our favor overall.

I'm merging this now because I want to see how it behaves for a while. Then I can update BVLC's travis setup based on what I learn: https://github.com/BVLC/caffe/pull/4207#discussion_r64486749

lukeyeager:
Woohoo it's working great! Build+test times are consistently about 12 minutes now.
https://travis-ci.org/NVIDIA/DIGITS/pull_requests

PR #789, opened 2016-05-26
Add gradient view extension and update regression tutorial

gheinrich:
Depends on #785


lukeyeager:
I'm getting an error when trying to test an LMDB on a gradients model:
```
  File "/home/lyeager/digits/digits/model/images/generic/views.py", line 435, in infer_db
    summary_html=summary_html,
UnboundLocalError: local variable 'summary_html' referenced before assignment
```

gheinrich:
> I'm getting an error when trying to test an LMDB

Only with Torch, right? I'm preparing a patch for this, thanks!

lukeyeager:
> Only with Torch, right?

Nope, that was with Caffe

lukeyeager:
Thanks for fixing that (unrelated) bug in https://github.com/NVIDIA/DIGITS/pull/791. This works great now!

PR #790, opened 2016-05-26
Home page - update UI and sort by network outputs

jmancewicz:
I've made a few changes to the front page tables.  I've moved the running jobs to the angularjs table, and ported the selection into angular from jquery. 

This also has the settings menu which allows the user to modify the view of the model table. These settings are persistent.

screen shot 2016-05-25 at 5 35 46 pm


lukeyeager:
Can we use a minimized version of ngStorage and add a copyright notice to the top?

digits/static/js/ngStorage.js
```diff
@@ -0,0 +1,233 @@
+(function (root, factory) {
```
lukeyeager: Can we use a minimized version of this file and add a copyright notice to the top? You didn't edit it, did you?

lukeyeager:
There's a bug in the sorting by network outputs. When I try to sort based on a network output that only exists for one model, all the models that don't have that network output show up at the top, regardless of whether I sort ascending or descending. I'd expect anything with a non-null value at the top, sorted by the requested method.

home-pages-bug1

lukeyeager:
When sorting by "Loss last", is it using the training output or the validation output? Both combined?

lukeyeager:
I love that finished jobs drop down into the appropriate panel (solving https://github.com/NVIDIA/DIGITS/issues/741#issuecomment-219502413).

However, it's missing most of the job metadata when it drops down like elapsed time and network outputs:

home-pages-bug2

lukeyeager:
Can we not rename the network outputs? So, leave them like "loss_bbox" instead of "Loss Bbox".

jmancewicz:
Initially, I had total control of the column titles and the data keys. The keys were all lowercase with underscores and the titles were title case with spaces. So they could be converted back and forth. Now with data keys such as accuracyA, that assumption was no longer valid.  The column sorting converted the titles to data keys, and the attempt to sort 'accuracyA' was treated as 'accuracya' and not sorting. That's why the sort had the blanks at the top; it didn't have any field named accuracya.

And I've changed the method to not capitalize a word if the word has any capital letters in it. 'xyz_last' => 'Xyz Last', but 'xyZ_last' => 'xyZ last'.

I'm made it so that field names are only modified for display. That fixed the problem for my test. @lukeyeager, please verify that that fix sorts your accuracyA column correctly.

jmancewicz:
Again mAP_last was displaying as Map Last and converted to map_last as a key. Now, it use mAP_last as the key and display as 'mAP Last'.

jmancewicz:
Does the conversion of  "loss_bbox" to "Loss Bbox" make it confusing?

jmancewicz:
The loss would be the validation loss. 

In the /models page, the loss would have been the training loss.

I've added a commit in which all outputs are there. I don't like that the names need to be so long, and there's data that I would think we could skip like 'min epoch'.

screen shot 2016-05-29 at 1 25 24 pm

I've removed the conversion of the keys (from 'leaning_rate' to 'Learning Rate'). I dislike that it adds inconsistency in the case of ui elements, but do like that the names are consistent with how they are represented elsewhere. So, I'm good either way. 

I've added (train) and (val) to the labels which how they are represented on the graphs.

digits/static/js/ngStorage.js
lukeyeager: Thanks!

lukeyeager:
I like the flexibility of the new list - now I have access to everything. However, I have 203 things I can sort by (calculated with `$("ul.dropdown-menu > li").length`), which makes the dropdown list longer than my list of 170 jobs. That seems excessive.

Also, the sorting still isn't working. See screenshots above.

jmancewicz:
Just committed a change that fixes the sorting issue caused by spaces in the field name.

lukeyeager:
Confirmed that sorting works now

jmancewicz:
Added a filter to move the empty values to the end of the list during sort.

jmancewicz:
Changed the page so that when the status change causes the job to move from the Running Jobs table, it pulls date from the server for the job including all the network output data that isn't needed for the Running Jobs view.

lukeyeager:
Looks good to me! Thanks for the fix - this is really slick now.

Please squash and I'll merge.

PR #791, opened 2016-05-26
Fix use of uninitialized summary_html

gheinrich:
`summary_html` may be used unitialized if an error occurs during inference.


lukeyeager:
*Reported at https://github.com/NVIDIA/DIGITS/pull/789#issuecomment-221912150*

PR #795, opened 2016-05-27
Fix use of non-common dataset interface in generic models

gheinrich:
The common dataset interface exposes a get_mean_file() method to retrieve the mean file.
Using the dataset.mean_file attribute only works for "images/generic" datasets.
This should have been changed in #723


lukeyeager:
How do I test this fix?

lukeyeager:
> How do I test this fix?

Oh ... #797

PR #796, opened 2016-05-27
Remove unused resize_mode in Torch path

gheinrich:
The resize mode is no longer used by the Torch wrapper in `test.lua`.
Besides, this field isn't part of the common dataset interface.
Using it is causing problems during multiple image inference in Torch when using generic datasets.
This should have been changed in #723


lukeyeager:
I just did multiple inference in Torch on an old siamese model and it worked fine. How do I verify the bug and fix?

lukeyeager:
> How do I verify the bug and fix?

Oh ... #797

PR #797, opened 2016-05-27
Add tests for models trained on generic datasets

gheinrich:
Depends on #795 and #796


lukeyeager:
Thanks for the tests and fixes!

lukeyeager:
Aw rats, git added two copies of your commit that reads "Remove unused resize_mode in Torch path" because it had a different checksum. Oh well, no harm done.

PR #798, opened 2016-05-27
Update inference.py (bug fix)

patricio-astudillo:
The error message contains a reference to self, which is not present since it is outside a class. I propose to remove the self from the error message.

### DIFF
--- a/tools/inference.py
+++ b/tools/inference.py
@@ -93,7 +93,7 @@ def infer(input_list, output_dir, jobs_dir, model_id, epoch, batch_size, layers,
                 snapshot_filename = f
                 break
     if not snapshot_filename:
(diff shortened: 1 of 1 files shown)


gheinrich:
That looks good to me thanks for fixing this!

gheinrich:
@patricio-astudillo do you mind filling the CLA (see [instructions](https://github.com/NVIDIA/DIGITS/blob/61427811a2b7b23cc2004efe8ebc6b7ded200681/.github/CONTRIBUTING.md#pull-requests)). Thanks!

patricio-astudillo:
Ok, Ive sent the CLA to digits@nvidia.com
Is there anything else I need to do from my side?

lukeyeager:
Thanks for the fix!

PR #803, opened 2016-05-31
Object Detection walk-through

gheinrich:


examples/object-detection/README.md
```diff
@@ -0,0 +1,134 @@
...
+- An image folder contains supported images (.png, .jpg, .jpeg, .bmp, .ppm).
+- A label folder contains .txt files in KITTI format that define the ground truth.
+Note that for each image in the image folder there must be a corresponding text file in the label folder.
+For example if the image folder includes an image named `foo.png` then the label folder needs to include a file named `foo.txt`.
+
+There needs to be one of the above set of directories for each of the training and validation sets.
+
+### Loading the data into DIGITS
+
+To create an Object Detection dataset in DIGITS, [this version of `3.4-dev`](https://github.com/NVIDIA/DIGITS/tree/2613082eec6c207548c99bb69fa5defc9473b1e8), or any more recent version, is required.
```
lukeyeager: Out-of-date reference. Let's just say DIGITS 4.0?

examples/object-detection/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+On the DIGITS home page, select the `Datasets` tab then click `Image>Object Detection`:
```
lukeyeager: The UI changed a little bit: `New Dataset > Images > Object Detection`.

examples/object-detection/README.md
```diff
@@ -0,0 +1,134 @@
...
+![dataset form](dataset-review.png)
+
+## Model creation
+
+### DetectNet
+
+In this example we will use **DetectNet**.
+DetectNet is a GoogLeNet-derived network that is specifically tuned for Object Detection.
+
+In order to train DetectNet, [NV-Caffe](https://github.com/NVIDIA/caffe) version [0.15.1](https://github.com/NVIDIA/caffe/tree/00f4c1db1dd579d31b1a4f953af61022f6b0da7a) or later is required.
```
lukeyeager: Let's use NVcaffe to be consistent with other documentation.
That URL can be simplified to https://github.com/NVIDIA/caffe/tree/v0.15.1

examples/object-detection/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Training DetectNet in DIGITS
+
+On the DIGITS home page, select the `Models` tab then click `Image>Object Detection`:
+
+![select dataset](select-object-detection-model.png)
+
+On the model creation page:
+- Select the dataset that was created in the previous section.
+- Set the base learning rate to 0.001.
```
lukeyeager: `0.0001` works best for me. @ssarathy any problem with that? Your suggested value was twice as big: `0.0002`.

examples/object-detection/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+You may click `Visualize` to review the network topology:
+
+![click](click-visualize.png)
+
+![detectnet](detectnet.png)
+
+> NOTE: this instance of DetectNet requires at least 12GB of GPU memory. Consider reducing the training batch size if you have less memory:
+```
+layer {
```
lukeyeager: Instead of suggesting that they change the network description, let's suggest they use the Solver UI to set the batch size to 2 and the batch accumulation to 5. That fits in memory on a 4GB board and is mathematically equivalent to a batch size of 10.

lukeyeager:
The screenshots for the home page and for the "New Object Detection Dataset" page need updating with the latest UI changes.

examples/object-detection/README.md
ssarathy: I got pretty good results with 0.0001 and 0.0002 . So either value is fine.

examples/object-detection/README.md
jmancewicz: The route here takes me to the installation instructions for BVLC http://caffe.berkeleyvision.org/installation.html. A clear link to the digits instructions for building caffe would be good.

jmancewicz: The current top of tree NVcaffe does not compile with cudnn 4.0.
Based on the error, I'd guess it's from this commit to NVcaffe

lukeyeager: > The current top of tree NVcaffe does not compile with cudnn 4.0.
> Based on the error, I'd guess it's from this commit to NVcaffe
> 4248b1f - (2 weeks ago) Use cuDNN routine FindEx to find best algorithm. - Pooya Davoodi

Please file a bug here: https://github.com/NVIDIA/caffe/issues/new

examples/object-detection/README.md
```diff
@@ -0,0 +1,134 @@
...
+  include {
+    phase: TRAIN
+  }
+  data_param {
+    batch_size: 8
+    backend: LMDB
+  }
+}
+```
+
```
jmancewicz: Maybe mention adding the Pretrained model if downloaded.
screen shot 2016-06-16 at 5 12 19 pm

examples/object-detection/README.md
jjmata: FWIW, the dataset instructions are for advanced users only.  The naive user (like me) will probably have a hard time picking the right KITTI files/organizing them.

examples/object-detection/README.md
gheinrich: I think that's mentioned on line 83

IsaacYangSLA:
@gheinrich I will try it on Windows and get the results.

lukeyeager:
Isaac, if you do want to try this on Windows, check the latest updates to this pull request. I updated it to change the suggested learning rate, etc.

examples/object-detection/README.md
douzsh: I tried to train detectNet with Digits. Everything is OK after I applied the latest changes but the result is always 0...  50 zero boxes after forward_net. I don't know why.
I used the prototxt in NVIDIA/caffe/examples/kitti. Do you have any suggestion?

RSly:
Hi, I was checking this walk-through to reproduce the results.
I couldn't follow the part regarding the Kitti database and .txt label format. I could download some txt files (data_object_label_2.zip, are these the good ones?), but It was not clear what is in those .txt files (looks like 3D bounding boxes?)

=> could you please provide the links to the images and labels to be used for this walk-through? 
=>Also, in order to train with other databases, an example of such files with a little explanation of what should go to the label .txt can be of great help (can it be like [x,y, width, height, category] of a bounding box? what should be it?).

Thanks

Greendogo:
It would definitely be helpful to more clearly define the datasets we're to download.
From here: http://www.cvlibs.net/datasets/kitti/eval_object.php
I can see that they are quite large, and I'd rather not download one only to find out it was the wrong one later:

> Download left color images of object data set (12 GB)
> Download right color images, if you want to use stereo information (12 GB)
> Download the 3 temporally preceding frames (left color) (36 GB)
> Download the 3 temporally preceding frames (right color) (36 GB)
> Download Velodyne point clouds, if you want to use laser information (29 GB)
> Download camera calibration matrices of object data set (16 MB)
> Download training labels of object data set (5 MB)
> Download object development kit (1 MB)

douzsh:
Download left color images of object data set (12 GB)
Download training labels of object data set (5 MB)

I think these two sets are what you need.

PR #804, opened 2016-05-31
Workaround for python layer in Windows

IsaacYangSLA:
When Python layer is included in network, an exception occurred in caffe.exe (see [this](https://github.com/Microsoft/caffe/issues/87), related to LayerParamter having no to_python converter.  However, both pycaffe in Windows, and caffe/pycaffe in Linux worked without issue.  This workaround tries to run pycaffe in Windows when both of the following conditions meet.
1.  At least one layer in the network has Type == 'Python'
2.  platform.system() is 'Windows'
In other conditions, the usual caffe.exe will be used.

### DIFF
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -13,6 +13,7 @@ import time
 
 from google.protobuf import text_format
 import numpy as np
+import platform
 import scipy
 
 from .train import TrainTask
@@ -787,6 +788,32 @@ class CaffeTrainTask(TrainTask):
 
     @override
     def task_arguments(self, resources, env):
+        """
+        Generate Caffe command line options or, in certain cases, pycaffe Python script
+        Returns a list of strings
+
+        Arguments:
+        resources -- dict of available task resources
+        env -- dict of environment variables
+        """
+        if platform.system() == 'Windows':
+            if any([layer.type == 'Python' for layer in self.network.layer]):
+                # Arriving here because the network includes Python Layer and we are running inside Windows.
+                # We can not invoke caffe.exe and need to fallback to pycaffe
+                # https://github.com/Microsoft/caffe/issues/87
+                # TODO: Remove this once caffe.exe works fine with Python Layer
+                win_python_layer_gpu_id = None
+                if 'gpus' in resources:
+                    n_gpus = len(resources['gpus'])
+                    if n_gpus > 1:
+                        raise Exception('Please select single GPU when running in Windows with Python layer.')
+                    elif n_gpus == 1:
+                        win_python_layer_gpu_id = resources['gpus'][0][0]
+                # We know which GPU to use, call helper to create the script
+                return self._pycaffe_args(win_python_layer_gpu_id)
+
+        # Not in Windows, or in Windows but no Python Layer
+        # This is the normal path
         args = [config_value('caffe_root')['executable'],
                 'train',
                 '--solver=%s' % self.path(self.solver_file),
@@ -806,7 +833,33 @@ class CaffeTrainTask(TrainTask):
                     args.append('--gpu=%s' % ','.join(identifiers))
         if self.pretrained_model:
             args.append('--weights=%s' % ','.join(map(lambda x: self.path(x), self.pretrained_model.split(':'))))
+        return args
 
+
+    def _pycaffe_args(self, gpu_id):
+        """
+        Helper to generate pycaffe Python script
+        Returns a list of strings
+
+        Arguments:
+        gpu_id -- the GPU device id to use
+        """
+        # TODO: Remove this once caffe.exe works fine with Python Layer
+        gpu_script = "caffe.set_device({id});".format(id=gpu_id) if gpu_id else ""
+        if self.pretrained_model:
+            weight_files = ','.join(map(lambda x: self.path(x), self.pretrained_model.split(':')))
+            loading_script = "solv.net.copy_from('{weight}');".format(weight=weight_files)
+        else:
+            loading_script = ""
+        command_script =\
+            "import caffe;" \
+            "{gpu_script}" \
+            "solv=caffe.{solver_type}Solver('{solver_file}');" \
+            "{loading_script}" \
+            "solv.solve()" \
+            .format(gpu_script=gpu_script, solver_type=self.solver_type,
+                    solver_file = self.solver_file, loading_script=loading_script)
+        args = ['python -c '+ '\"' + command_script + '\"']
         return args
 
     @override
--- a/digits/task.py
+++ b/digits/task.py
@@ -184,7 +184,7 @@ class Task(StatusCls):
         self.before_run()
 
         env = os.environ.copy()
-        args = self.task_arguments(resources, env )
+        args = self.task_arguments(resources, env)
         if not args:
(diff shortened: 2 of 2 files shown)


digits/model/tasks/caffe_train.py
```diff
@@ -26,6 +26,7 @@
 # Must import after importing digit.config
 import caffe
 import caffe_pb2
+import platform
```
lukeyeager: Can you move this up just before `import scipy`?

I'd like to stick with this style: https://github.com/NVIDIA/DIGITS/pull/501

digits/model/tasks/caffe_train.py
```diff
@@ -807,6 +813,26 @@ def task_arguments(self, resources, env):
         if self.pretrained_model:
             args.append('--weights=%s' % ','.join(map(lambda x: self.path(x), self.pretrained_model.split(':'))))
 
+        if platform.system() == 'Windows' and including_python_layer:
+            if len(identifiers) > 1:
```
lukeyeager: `identifiers` is not in scope if `gpus` is not in resources

digits/model/tasks/caffe_train.py
```diff
@@ -807,6 +813,26 @@ def task_arguments(self, resources, env):
...
+                    self.logger.warning('Python layer can not work in multiple GPU.  Will use the first GPU only')
```
lukeyeager: I think you're getting confused about which issue you're trying to solve.

1. multi-GPU + Python layer fails on all systems. We don't need to check for this - let the framework deal with it.
2. single-GPU + Python layer fails on Windows unless you run it with pycaffe.

Don't try to solve (1) in this pull request.

digits/model/tasks/caffe_train.py
IsaacYangSLA: Sure.  Thanks for the link on the coding style.

digits/model/tasks/caffe_train.py
IsaacYangSLA: Yes.  The framework is the best place to handle issue 1.  I will modify the codes not to check multi-GPU case.

IsaacYangSLA:
The codes were modified to follow suggestions.
1.  Raise an exception when users requesting multiple-GPU, but the environment is Windows and python layer is included in the network.  Currently pycaffe does not have set_solver_count, and set_device accept single gpu_id.
2.  When no gpu_id is specified, pycaffe tries not to set_device.
3.  Re-arrange import platform.

digits/model/tasks/caffe_train.py
```diff
@@ -798,15 +806,42 @@ def task_arguments(self, resources, env):
                 identifiers.append(identifier)
             if len(identifiers) == 1:
                 args.append('--gpu=%s' % identifiers[0])
+                if win_python_layer:
+                    win_python_layer_gpu_id = identifiers[0]
```
lukeyeager: I thought we agreed to throw an error here instead of silently ignoring the other GPUs?

digits/task.py
```diff
@@ -199,6 +199,11 @@ def run(self, resources):
 
         import sys
         env['PYTHONPATH'] = os.pathsep.join(['.', self.job_dir, env.get('PYTHONPATH', '')] + sys.path)
+
+        #https://docs.python.org/2/library/subprocess.html#converting-argument-sequence
+        if platform.system() == 'Windows':
+            args = ' '.join(args)
+
```
lukeyeager: Can you add a debug message here while you're at it? Something like this:
```py
self.logger.debug(args)
```

digits/model/tasks/caffe_train.py
IsaacYangSLA: This is one GPU case.  Line 817~818 throws exception and the browser will show it.  Like this:
image

digits/task.py
IsaacYangSLA: Sure.  Sorry that I forgot that.

digits/model/tasks/caffe_train.py
lukeyeager: Oh you're right, I see it now. Nevermind.

IsaacYangSLA:
Another code clean-up and re-arrange.  Thanks to Luke for the comments and better ideas to format the codes.

digits/model/tasks/caffe_train.py
```diff
@@ -787,10 +788,24 @@ def iteration_to_epoch(self, it):
...
+            if any([layer.type == 'Python' for layer in network.layer]):
+                win_python_layer_gpu_id = None
+                if 'gpus' in resources:
+                    n_gpus = len(resources['gpus'])
+                    if n_gpus > 1:
+                        raise Exception('Please select single GPU when running in Windows with Python layer.')
+                    elif n_gpus == 1:
+                        win_python_layer_gpu_id = resources['gpus'][0][0]
+                win_python_layer_args = self.pycaffe_args(win_python_layer_gpu_id)
+                return win_python_layer_args
```
lukeyeager: A simple `return self.pycaffe_args(...)` should suffice

digits/model/tasks/caffe_train.py
```diff
@@ -787,10 +788,24 @@ def iteration_to_epoch(self, it):
...
+
         args = [config_value('caffe_root')['executable'],
-                'train',
-                '--solver=%s' % self.path(self.solver_file),
-                ]
+            'train',
+            '--solver=%s' % self.path(self.solver_file),
+            ]
```
lukeyeager: No need to change this whitespace

digits/model/tasks/caffe_train.py
IsaacYangSLA: Yes.  That variable is totally unnecessary.

digits/model/tasks/caffe_train.py
IsaacYangSLA: Seems my IDE changed them.  I will change its settings, and revert those spaces back.

lukeyeager:
Tested on Ubuntu - doesn't cause any issues as far as I can tell.

digits/model/tasks/caffe_train.py
```diff
@@ -806,7 +833,33 @@ def task_arguments(self, resources, env):
...
+    def _pycaffe_args(self, gpu_id):
+        """
+        Helper to generate pycaffe Python script
+        Returns a list of strings
+
+        Arguments:
+        gpu_id -- the GPU device id to use
+        """
+        # TODO: Remove this once caffe.exe works fine with Python Layer
+        gpu_script = "caffe.set_device({id});".format(id=gpu_id) if gpu_id else ""
```
lukeyeager: Hey @IsaacYangSLA are you sure this actually runs in GPU mode? Don't you need a `set_mode()` too?

yanwenCi:
An error occurred when I use python layer in windows caffe, the error information is as follows, 
` No to_python (by-value) converter found for C++ type: class caffe::LayerParameter `

and I test the python layer in python console, the error again as follows

`Boost.Python.ArgumentError: Python argument types in
    Layer.__init__(DataLayer)
did not match C++ signature:
    __init__(struct _object * __ptr64, class caffe::LayerParameter)`

Why did this error happen? How can I figure it? 
Appreciate for your help.

PR #806, opened 2016-06-01
Image processing workflow

gheinrich:
This adds data and view extensions to train image processing networks in DIGITS. This may be used for de-noising, super-resolution, segmentation, etc.

The image processing data extension creates datasets for which both the input and the label are images.

The image view extension displays the network output as an image.

Only works with Caffe for now. Torch wrappers need to be updated to deal with image labels (currently only scalar or vector labels are supported).

### DIFF
--- a/digits/dataset/generic/test_views.py
+++ b/digits/dataset/generic/test_views.py
@@ -3,8 +3,11 @@ from __future__ import absolute_import
 
 import json
 import os
+import tempfile
 
 from bs4 import BeautifulSoup
+import numpy as np
+import PIL.Image
 
 import digits.test_views
 from digits import extensions
@@ -52,6 +55,10 @@ class BaseViewsTestWithDataset(BaseViewsTest):
     Provides some functions
     """
 
+    IMAGE_WIDTH = 32
+    IMAGE_HEIGHT = 32
+    CHANNELS = 1
+
     @classmethod
     def create_dataset(cls, **kwargs):
         """
@@ -77,7 +84,7 @@ class BaseViewsTestWithDataset(BaseViewsTest):
         if request_json:
             if rv.status_code != 200:
                 raise RuntimeError(
-                    'Model creation failed with %s' % rv.status_code)
+                    'Dataset creation failed with %s' % rv.status_code)
             return json.loads(rv.data)['id']
 
         # expect a redirect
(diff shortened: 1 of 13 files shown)


pansk:
:+1: for image labels in torch.
Is it possible also to have multiple image labels, or multiple image sources?

E.G. providing a normal map and a diffuse map and produce a shaded image (multiple input), or provide a single image and produce a super-resolution and a segmentation map (multiple output)

As far as you know, can this be done by creating the DB manually, and adding multi-channel images (for labels, sources, or both)?

gheinrich:
Hi @pansk we are using Caffe `Datum` objects to store data in LMDB. We may have one LMDB for features (inputs) and one LMDB for labels. That `Datum` format is used for both Caffe and Torch. We support only one `Datum` for the features and only one `Datum` for the labels. `Datum` objects require your data to be either actual ".png/.jpg/..." images ("encoded" case) or any 3D (`Channels x Height x Width`) tensor ("unencoded" case). 

You can store multiple images in your unencoded `Datum` if you put them side by side across the channel dimension. For example you can store 2 RGB images by constructing a 6-channel tensor.

Given this `Datum` limitation what I'm planning to do for the moment in Torch is to accept anything that can be stored in `Datum` objects. Later on we might consider using HDF5 for those generic datasets and extend support to N-dimensional tensors though that is a longer-term project.

Would that work for you?

digits/dataset/generic/test_views.py
```diff
@@ -1,13 +1,16 @@
 # Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
 from __future__ import absolute_import
 
-import json
-
 from bs4 import BeautifulSoup
+import json
+import numpy as np
```
lukeyeager: Major nitpick here, but PEP8 likes to separate standard library imports and 3rd party imports:

> Imports should be grouped in the following order:
1. standard library imports
2. related third party imports
3. local application/library specific imports
You should put a blank line between each group of imports. 

I've been trying to follow that format in our code since https://github.com/NVIDIA/DIGITS/pull/501.

digits/dataset/generic/test_views.py
```diff
@@ -231,6 +258,39 @@ def test_edit_notes(self):
...
+        assert dims == [self.IMAGE_HEIGHT, self.IMAGE_WIDTH, self.CHANNELS]
+
+
+class RandomImageSet(object):
+    """
+    Create a folder of random grayscale images
+    """
+
+    @classmethod
+    def create_imageset(cls, **kwargs):
```
lukeyeager: Seems simpler to make this a function instead of a method to avoid unnecessary class mixins.

digits/extensions/data/imageProcessing/data.py
```diff
@@ -0,0 +1,154 @@
...
+class DataIngestion(DataIngestionInterface):
+    """
+    A data ingestion extension for an image processing dataset
+    """
+
+    def __init__(self, **kwargs):
+        """
+        Init
+        """
+        # the parent __init__ method automatically populates this
```
lukeyeager: This docstring is pretty useless. Could you roll the comment up into the docstring?

digits/extensions/view/imageOutput/view.py
```diff
@@ -0,0 +1,106 @@
...
+
+CONFIG_TEMPLATE = "config_template.html"
+SUMMARY_TEMPLATE = "summary_template.html"
+VIEW_TEMPLATE = "view_template.html"
+
+
+@subclass
+class Visualization(VisualizationInterface):
+    """
+    A visualization extension to display bounding boxes
```
lukeyeager: Wrong docstring

digits/dataset/generic/test_views.py
gheinrich: actually I am not sure I get the difference between a standard library import and a 3rd party import. Is it correct to say that `numpy` and `PIL.Image` are 3rd party imports and `json`, `os`, `tempfile` are standard library imports?

lukeyeager: I think what they're distinguishing between are packages that come with a standard Python install (i.e. `apt-get install python`) vs. add-on packages (i.e. `pip install Flask`). At least, that's been my interpretation. I'm open to push-back if you think it's dumb.
```py
import json
import os
import tempfile

from bs4 import BeautifulSoup
import numpy as np
import PIL.Image

from digits import extensions
import digits.test_views
from digits.utils import constants
```

digits/dataset/generic/test_views.py
gheinrich: Sure!

digits/dataset/generic/test_views.py
gheinrich: I get it, thanks!

gheinrich:
Rebased and updated according to comments:
https://github.com/NVIDIA/DIGITS/pull/830#discussion_r72294307
https://github.com/NVIDIA/DIGITS/pull/830#discussion_r72294402

m5061125:
hi, gheinrich.
I am using digits 5 for medical segmentation, so far it is works fine for 2D images, I am wondering that whether if the digits support 3D image classiffication or segmentation. I know that caffe support N-D convolution and pooling by feeding with the hdf5 data format. Pls let me know if digits already support it, thanks.

PR #808, opened 2016-06-01
Add box size threshold for object detection validation set

gheinrich:


lukeyeager:
I'm going to try to run some jobs overnight with and without this change (both with the weight_decay change) to see how this affects performance.

lukeyeager:
My overnight tests indicate that this is doing pretty much what we expect.

#### Including all boxes
max mAP - 40.5
max precision - 76.8
max recall - 52.4

#### Excluding small boxes
max mAP - 50.0
max precision - 77.0
max recall - 65.3

PR #810, opened 2016-06-02
Separate, Expandable Plots

### LINKED ISSUE #786: Group network outputs
Currently, DIGITS displays all network outputs in one big graph, and another separate one for the learning rate:

outputs-all-together

As you can see, that's pretty useless for more complicated networks with many outputs.

In another internal NVIDIA tool, we group network outputs by type (or by name?) and show them in separate graphs, with the learning rate super-imposed over each graph:

outputs-separated

I find that much nicer. I'd like to add this to DIGITS, but in such a way that it doesn't take up the whole page.

Lucaszw:
Close #786 

screenshot from 2016-06-02 12 05 28

- Separate plots based on keywords including "val", and "test" 
- Learning rate shown on separate axis in same plot. 
- Plots are expandable and collapsible.


lukeyeager:
There's a bug when trying to view GoogLeNet. Maybe it can't handle slashes in the name? I'm getting `SyntaxErrors` in my console.

segmented-plots-googlenet

lukeyeager:
Looks like live updates aren't working very well. When I start a job and let it run to completion, the page should stay pretty much unchanged whether I refresh the page or not.

Before page refresh (looks bad):
live-updates-no-refresh

After page refresh (looks right):
live-updates-after-refresh

Lucaszw:
These problems should be fixed now. I also fixed a few things in terms of the ui as well. 

before
after

Lucaszw:
I migrated away from c3 to Chart.js, and things are running allot smoother. I also added some additional UI tweaks such as having the x-axis linked between the charts. 

screenshot from 2016-06-06 20 08 33

lukeyeager:
New chart library looks cool!

* "All Network Outputs" instead of "All Plots"
* Use straight lines for interpolation
* Turn off fancy transitions for sake of performance
* Can we grey out disabled outputs instead of striking out?
* There's a bug with the x-axis selector when job is still in 1st epoch

Lucaszw:
Should be good to go!

lukeyeager:
@jmancewicz will you take a look and see what you think?

jmancewicz:
In the View Large chart, when I move the slider, I see flickering of x axis labels, which have some bug, resulting (in the image) the number 28 being represented as 28.0000000000004 at about 45º angle.

screen shot 2016-06-15 at 8 02 26 pm

jmancewicz:
The value tips are clipped in the small charts.
screen shot 2016-06-15 at 8 07 45 pm

jmancewicz:
Chart.js seems more buggy than c3.

jmancewicz:
When a chart has data with different sample rates, the mouse over that displays the value tip, highlights the points on all datasets by index value.  c3 handles this correctly. I'm thinking this needs to go back to c3.

screen shot 2016-06-15 at 8 15 53 pm

jmancewicz:
As the domain slider is adjusted, I would expect the range scale would scale to reflect visible extrema.

jmancewicz:
When I mouse over a point in the chart, the cursor changes to pointer, but there is seems to be no action available.

Also, the mouse-over value tip is gone.

jmancewicz:
Then there is the question of what if any of the changes to the charts should be saved in local or session storage. It's easy to argue for any position, and I'll start; I think having the domain slider and the collapsed state saved in local storage might give a nice experience, especially when the range is scaled by the visible values.

Lucaszw:
Thanks Joe! I added a Range Slider for the Y-Axis , and changed the cursor to auto.

Lucaszw:
@lukeyeager Can you check if things are working for you? Thanks!

lukeyeager:
Closing for now (there are merge conflicts anyway). I'd like to revisit this.

I've backed up this work at https://github.com/lukeyeager/DIGITS/tree/backup/separate-output-graphs.

TimZaman:
Mourning this PR :(. I'd like to revisit this in plain old native D3 when I'm all out of stories.

Lucaszw:
@TimZaman It would definitely be cool to see this done in native D3 :)

TimZaman:
@Lucaszw I just had the best idea. I will do your homework, if you resurrect this PR? ;)

Lucaszw:
@TimZaman Haha alrighty ;) . I currently don't have a good dev machine to work off of (plus school being a little crazy); but I will try resurrecting this PR as soon as I get a chance :)

PR #813, opened 2016-06-06
Don't show create_generic_db dummy tasks

gheinrich:
Show only those tasks that are not 'D' (done without error) and those that have more than zero entries.


lukeyeager:
Seems like a clear improvement to me. Thanks!

PR #816, opened 2016-06-07
Fix encoded image channel order

gheinrich:
We shouldn't change the RGB channel order to BGR when using encoded images.


gheinrich:
Can be tested with #817

lukeyeager:
Fix verified. Thanks!

lukeyeager:
As I understand it, this shouldn't affect DetectNet convergence, but should end up creating a model which DIGITS will test properly.

PR #823, opened 2016-06-08
Delete a snapshot

cagbal:
I added deleting a specific snapshot of a model on the model page, since my harddisk sometimes runs out of space when so many model are saved. So, I put a Delete Model button under Download Model Button.
The deletion process does not allow: 

- deleting all snapshots (in that case, it becomes deletion of all model)
- deleting the last snapshot 

screenshot from 2016-06-08 22 13 42

### DIFF
--- a/digits/model/views.py
+++ b/digits/model/views.py
@@ -7,8 +7,10 @@ import json
 import math
 import tarfile
 import zipfile
+import os
 
 import flask
+from flask import redirect, url_for
 import werkzeug.exceptions
 
 from . import images as model_images
@@ -30,6 +32,12 @@ def show(job_id):
         {id, name, directory, status, snapshots: [epoch,epoch,...]}
     """
     job = scheduler.get_job(job_id)
+
+    task = job.train_task()
+
+    # Update the snapshots
+    task.detect_snapshots()
+
     if job is None:
         raise werkzeug.exceptions.NotFound('Job not found')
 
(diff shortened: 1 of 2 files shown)


digits/model/views.py
```diff
@@ -7,8 +7,10 @@
 import math
 import tarfile
 import zipfile
+import os
 
 import flask
+from flask import redirect, url_for
```
lukeyeager: Can you use the above import like `flask.redirect` and `flask.url_for`? That lines up with the coding style elsewhere in DIGITS.

digits/model/views.py
```diff
@@ -225,6 +233,51 @@ def download(job_id, extension):
     response.headers['Content-Disposition'] = 'attachment; filename=%s_epoch_%s.%s' % (job.id(), epoch, extension)
     return response
 
+@blueprint.route('/<job_id>/delete',
+        methods=['POST'])
+def delete(job_id):
```
lukeyeager: This route name is misleading. Can you format the route like `/<job_id>/delete_snapshot[.json]?epoch=<epoch>` instead?

cagbal: Thanks! I was not sure what the name should be. I will fix them and since I'm not so familiar with travis-ci,  could not understand the problem yet. There is no problem on my machine.

gheinrich:
Hi @cagbal I am wondering why you closed this PR?

cagbal:
Hi @gheinrich, actually I don't really want to close this PR, but I could not understand why my code could not pass the Travis' test (I have not used Travis before). Now, I'm working on it. In case of taking too much time to solve it, I closed it. When I, hopefully, solve the problem, I will open a new PR.

digits/model/views.py
```diff
@@ -30,6 +32,12 @@ def show(job_id):
         {id, name, directory, status, snapshots: [epoch,epoch,...]}
     """
     job = scheduler.get_job(job_id)
+
+    task = job.train_task()
```
gheinrich: this needs to be moved a couple of lines down (after the test `if job is None:`)

gheinrich:
hi @cagbal you don't need to create a new Pull Request, you can update this Pull Request using `--force`:

```
git push --force <your fork> <your commit>:heads/refs/deletemodel
```

I've added a comment on why Travis tests failed with your patch.

Also, can you make the change for generic image models too (`digits/templates/models/images/generic/show.html
`)? Thanks!

cagbal:
Hi @gheinrich.  I saw your comment. Thank you for your helps!  I will change the generic image models and will update the PR as you write.

digits/model/views.py
cagbal: After this, it passed the test.

cagbal:
@gheinrich Reopen button has become inactive saying that The deletemodel branch was force-pushed or recreated! I think I need to open a new PR.

PR #824, opened 2016-06-08
Fix Flask requirement to 0.10.1

lukeyeager:
Otherwise, you get a cryptic error:
```
    exc_class = default_exceptions[exc_class_or_code]
KeyError: 300
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ protobuf>=2.5.0
 six>=1.5.2
 requests>=2.2.1
 gevent>=1.0,<=1.0.2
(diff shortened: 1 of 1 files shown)


lukeyeager:
This is the version provided by Ubuntu 14.04 and 16.04 packages, so it's well tested.
http://packages.ubuntu.com/trusty/python-flask
http://packages.ubuntu.com/xenial/python-flask

lukeyeager:
TravisCI won't actually test this change since Flask is installed with a deb package.

Merging now because I need it for internal testing.

PR #827, opened 2016-06-09
Fix incorrect solver type name mapping

IsaacYangSLA:
The letter cases of solver types are not always upper case in PyCaffe solvers.  Only SGD can be mapped directly to SGDSolver.  Others have to be mapped via manually-defined dictionary.
In case the solver type is not recognized, throw ValueError.


lukeyeager:
I don't love the hard-coded mapping, but I don't see any other good solution. It'll have to do.

Thanks for the follow-up to #804!

PR #828, opened 2016-06-09
Rework Torch multi-GPU training

gheinrich:
This change enables automatic encapsulation in an `nn.DataParallelTable` container for multi-GPU training. This comes from @thatguymike's realization that this would simplify the programming model. A `disableAutoDataParallelism` flag is added to the internal parameters to disable this mechanism. This flag is intended to be used by users who wish to retain control over what exactly gets parallelized. As an example, the standard Alexnet model enables this flag to parallelize only the feature part of the network.

Note: this changes does not enable threading in `nn.DataParallelTable` when doing the automatic encapsulation. Some of the numbers below are showing training time with/without threading.

For the record, some training time numbers:

LeNet:
1 GPU => 1m3s
2 GPUs, all parallel, single-thread => 1m33s;1m34s
2 GPUs, all parallel, multi-thread => 3m41s

Alexnet:
1 GPU => 11m6s
2 GPUs, feature parallel, single-thread => 10m29s
2 GPUs, feature parallel, multi-thread => 9m34s;10m6s
2 GPUs, all parallel, threaded => 13m47s

GoogLeNet:
1 GPU => 16m28s
2 GPUs, all parallel, single-thread => 12m28s
2 GPUs, all parallel, multi-thread => 11m6s;11m7s

Text classification (1D convolutions followed by FC layers):
1 GPU => 32m35s;32m14s
2 GPUs, all parallel, single-thread => 19m41s
2 GPUs, all parallel, multi-thread => 18m46s

Very Simple segmentation FCN (only 2 conv layers followed by a deconv):
1 GPU => 7m12s;7m11s
2 GPUs, all parallel, single-thread => 19m48s
2 GPUs, all parallel, multi-thread => 21m41s

GPUs are Titan X.

### DIFF
--- a/digits/standard-networks/torch/ImageNet-Training/alexnet.lua
+++ b/digits/standard-networks/torch/ImageNet-Training/alexnet.lua
@@ -82,6 +82,7 @@ return function(params)
     end
     return {
         model = createModel(params.ngpus, channels, nclasses),
+        disableAutoDataParallelism = true,
         croplen = 224,
         trainBatchSize = 128,
         validationBatchSize = 32,
--- a/digits/standard-networks/torch/ImageNet-Training/googlenet.lua
+++ b/digits/standard-networks/torch/ImageNet-Training/googlenet.lua
@@ -55,7 +55,7 @@ local function inception(input_size, config)
    return concat
 end
 
-function createModel(nGPU, nChannels, nClasses)
+function createModel(nChannels, nClasses)
    -- batch normalization added on top of convolutional layers in feature branch
    -- in order to help the network learn faster
    local features = nn.Sequential()
@@ -99,28 +99,10 @@ function createModel(nGPU, nChannels, nClasses)
    local splitter = nn.Concat(2)
    splitter:add(main_branch):add(aux_classifier)
    --local googlenet = nn.Sequential():add(features):add(splitter)
+
    local googlenet = nn.Sequential():add(features):add(main_branch)
 
-   local model
-   if nGPU>1 then
-      local gpus = torch.range(1, nGPU):totable()
-      local fastest, benchmark
-      local use_cudnn = cudnn ~= nil
-      if use_cudnn then
-        fastest, benchmark = cudnn.fastest, cudnn.benchmark
-      end
-      model = nn.DataParallelTable(1, true, true):add(googlenet,gpus):threads(function()
-            if use_cudnn then
-              local cudnn = require 'cudnn'
-              cudnn.fastest, cudnn.benchmark = fastest, benchmark
-            end
-      end)
-      model.gradInput = nil
-   else
-      model = googlenet
-   end
-
-   return model
+   return googlenet
 end
 
 -- return function that returns network definition
@@ -135,7 +117,7 @@ return function(params)
         assert(params.inputShape[2]==256 and params.inputShape[3]==256, 'Network expects 256x256 images')
     end
     return {
-        model = createModel(params.ngpus, channels, nclasses),
+        model = createModel(channels, nclasses),
         croplen = 224,
         trainBatchSize = 32,
         validationBatchSize = 16,
--- a/docs/GettingStartedTorch.md
+++ b/docs/GettingStartedTorch.md
@@ -106,6 +106,7 @@ labelHook             | function     | No        | A function(input,dblabel) tha
 trainBatchSize        | number       | No        | If specified, sets train batch size. May be overridden by user in DIGITS UI.
 validationBatchSize   | number       | No        | If specified, sets validation batch size. May be overridden by user in DIGITS UI.
 fineTuneHook          | function     | No        | A function(net) that returns the model to be used for fine-tuning. The untuned model is passed as a function parameter.
+disableAutoDataParallelism | boolean | No        | By default models are encapsulated in a nn.DataParallelTable container to enable multi-GPU training when more than 1 GPUs are selected. Setting this flag to `true` disables this mechanism.
 
 ### Tensors
 
--- a/digits/standard-networks/torch/lenet.lua
+++ b/digits/standard-networks/torch/lenet.lua
@@ -44,27 +44,8 @@ return function(params)
     lenet:add(nn.Linear(500, nclasses))  -- 500 -> nclasses
     lenet:add(nn.LogSoftMax())
 
-    local model
-    if params.ngpus > 1 then
-      local gpus = torch.range(1, params.ngpus):totable()
-      local fastest, benchmark
-      local use_cudnn = cudnn ~= nil
-      if use_cudnn then
-        fastest, benchmark = cudnn.fastest, cudnn.benchmark
-      end
-      model = nn.DataParallelTable(1, true, true):add(lenet,gpus):threads(function()
-            if use_cudnn then
-              local cudnn = require 'cudnn'
-              cudnn.fastest, cudnn.benchmark = fastest, benchmark
-            end
-      end)
-      model.gradInput = nil
-    else
-      model = lenet
-    end
-
     return {
-        model = model,
+        model = lenet,
         loss = nn.ClassNLLCriterion(),
         trainBatchSize = 64,
         validationBatchSize = 32,
--- a/tools/torch/main.lua
+++ b/tools/torch/main.lua
@@ -280,6 +280,12 @@ local parameters = {
 network = network_func(parameters)
 local model = network.model
 
+-- embed model in parallel table unless explicitly disallowed in user-defined description
+if nGpus > 1 and not network.disableAutoDataParallelism then
+    local gpus = torch.range(1, nGpus):totable()
+    model = nn.DataParallelTable(1, true, true):add(model, gpus)
+end
+
 -- if the loss criterion was not defined in the network
 -- use nn.ClassNLLCriterion() by default
 local loss = network.loss or nn.ClassNLLCriterion()


gheinrich:
@pansk @TimZaman any feedback on this? Thanks!

TimZaman:
Sure. Any reason for not utilizing 3 and 4 GPUs?

docs/GettingStartedTorch.md
```diff
@@ -106,6 +106,7 @@ labelHook             | function     | No        | A function(input,dblabel) tha
 trainBatchSize        | number       | No        | If specified, sets train batch size. May be overridden by user in DIGITS UI.
 validationBatchSize   | number       | No        | If specified, sets validation batch size. May be overridden by user in DIGITS UI.
 fineTuneHook          | function     | No        | A function(net) that returns the model to be used for fine-tuning. The untuned model is passed as a function parameter.
+disableAutoDataParallelism | boolean | No        | By default models are encapsulated in a nn.DataParallelTable container to enable multi-GPU training when more than 1 GPUs are selected. Setting this flag to `true` disables this mechanism.
```
TimZaman: How about enableAutoDataParalellism to avoid double negatives ie `(..) not network.disableAutoDataParallelism`

TimZaman:
Looks good to me. 
**Off topic:**
I don't want to be a 1 cent PITA but how about making all the standard networks in the same function-template as well? so instead of `lenet=` and `googlenet=` make it `model=` and pass the param structure the same way, etc. Maybe even in the UI when someone presses 'Custom Network' fill it with a template/skeleton.

**On topic:**

- Am I correct in thinking that by default the UI selects '1 gpu'  and so the MNIST will by default run on 1 GPU right? We don't want to parallelize the smaller networks i recon.
- How/is your training time affected by batch size, since you are splitting along the batch dimension.
- Any reason for not testing for 3 and 4 GPUs?

docs/GettingStartedTorch.md
gheinrich: I would like to turn on automatic encapsulation by default. I am not sure how to avoid the double negative in that case?

gheinrich:
> Am I correct in thinking that by default the UI selects '1 gpu' and so the MNIST will by default run on 1 GPU right? We don't want to parallelize the smaller networks i recon.

Exactly. As you can see `LeNet` and the tiny segmentation FCN both show worse results when doing multi-GPU training. By default, only 1 GPU is used in DIGITS.

> How/is your training time affected by batch size, since you are splitting along the batch dimension. Because the speed-up is significant but we might be able to tune it.

We are doing strong scaling (keeping batch size constant and splitting it across the GPUs). I think we'd get a greater speed-up if we increased the batch size as the number of GPUs grows but then the optimization would be different, you possibly would have to change learning rate, etc. But it's a good point I'll try that.

> Any reason for not testing for 3 and 4 GPUs?

No good reason. I have only 2 GPUs on my workstation :-(

docs/GettingStartedTorch.md
TimZaman: `disableAutoDataParallelism | boolean | Non  `
to
`enableAutoDataParallelism | boolean | Ouais  `

TimZaman:
> > Any reason for not testing for 3 and 4 GPUs?
> No good reason. I have only 2 GPUs on my workstation :-(

https://developer.nvidia.com/academic_gpu_seeding

docs/GettingStartedTorch.md
gheinrich: Ouais :) actually this column specifies whether the field is mandatory but I see what you mean, I could simply make it optional and default to `true`. Let me do that, thanks!

digits/standard-networks/torch/ImageNet-Training/alexnet.lua
```diff
@@ -82,6 +82,7 @@ return function(params)
     end
     return {
         model = createModel(params.ngpus, channels, nclasses),
+        disableAutoDataParallelism = true,
```
TimZaman: Why this and not a UI button?

docs/GettingStartedTorch.md
gheinrich: Actually I'm a bit unsure... with `disableAutoDataParallelism` the user can disable encapsulation with `disableAutoDataParallelism=true` and the Lua code looks like:

```lua
if not network.disableAutoDataParallelism then
```

If I change the flag to `enableAutoDataParallelism` then the user can disable automatic encapsulation with  `enableAutoDataParallelism=false` and the Lua code looks like:

```lua
if network.enableAutoDataParallelism ~= false then
```

The former method sounds more like positive logic to me given that we want to give an option to disable encapsulation (as we want encapsulation by default).

digits/standard-networks/torch/ImageNet-Training/alexnet.lua
gheinrich: Good point! `croplen`, `trainBatchSize` and `validationBatchSize` are there for feature parity with Caffe, which supports defining these in .prototxt. Besides, it's convenient to have a suitable default in the model description (e.g. GoogLeNet requires a much smaller batch size than LeNet) and it would be cumbersome to propagate these defaults to the UI. By extension I am inclined to keep the auto parallelism option in the model description as it's very specific to Torch and the appropriate value is model dependent.

TimZaman:
NooOoOooOo! You forgot to remove the ngpu stuff in `alexnet.lua`!
https://github.com/gheinrich/DIGITS/blob/2da0b05cfb3c70a29029c429bc5d813df458ff25/digits/standard-networks/torch/ImageNet-Training/alexnet.lua#L48-67
Now going to redo all my test so far XD

digits/standard-networks/torch/ImageNet-Training/googlenet.lua
```diff
@@ -99,28 +99,10 @@ function createModel(nGPU, nChannels, nClasses)
    local splitter = nn.Concat(2)
    splitter:add(main_branch):add(aux_classifier)
    --local googlenet = nn.Sequential():add(features):add(splitter)
```
TimZaman: What's the idea of L99-L101

digits/standard-networks/torch/ImageNet-Training/alexnet.lua
TimZaman: > (..), it's convenient to have a suitable default in the model description (e.g. GoogLeNet requires a much smaller batch size than LeNet) and it would be cumbersome to propagate these defaults to the UI

In my opinion all these settings should be propagated _from_ the UI. In an ideal scenario one would not pre-select a standard 'Network' but a standard 'Classification Job'; that also fills in the correct (meta and) hyper parameters. But that'd be another PR altogether.

docs/GettingStartedTorch.md
TimZaman: I see what you mean and I get it now.

TimZaman:
This looks good to me. Quite an elegant solution.
I am still running benchmarks for 40 hours on 3 and 4 GPUs. Preliminary results in table below.
Running GoogleNet on ImageNet for 5 epochs.

On 2 GPU's the autoDataParallelism through this PR is increased by around 10% (18h27m to 16h43m). Increasing the batch size from 32 to 128 decreases the time to 14h19m, a decrease from the original speed of 30%. (But this requires 2GPU's with 8Gb memory).

#GPU's | disableAutoDataParallelism | batchSize | time (HH:MM)
:---: | :---: | :---: | :---:
1 |  false | 32 | 18:25
1 |  true | 32 | 17:14
2 |  false | 32 | 16:43
2 |  true | 32 | 18:27
2 |  false | 128 | 14:19
2 |  true | 128 | ERROR*
3 | false | 32 | TBD
3 | true | 32 | TBD
4 | false | 32 | TBD
4 | true | 32 | TBD

* out of memory error

gheinrich:
Thank you very much for taking the time to make those measurements. Hopefully all the `disableAutoDataParallelism=true` will replicate the timings of the single-GPU case (otherwise something really odd is going on!). 

With "auto data parallelism" I have had to disable threading in `DataParallelTable` because Torch threads annoyingly require that you `require` all the modules you need in all sub-threads. For example a model of mine is using `dpnn`. If I want to enable threading in `DataParallelTable` then I need to `require('dpnn')` from within each sub-thread. I didn't see an elegant way to let users do this without some cumbersome callbacks, etc. I have also measured that this accounts for roughly 10% hit on performance. Users who really want those 10% back can set `disableAutoDataParallelism=true` and do the parallelization in the model description.

Validation takes a long time indeed. You probably know this already but just in case you can validate less often by setting `Validation interval` to more than `1` in the UI.

TimZaman:
Hey @gheinrich look at those measurements ^.  Look at the impact of the batch size. Also, for a 4 GPU run on a batch size of 128, it only takes 4Gb per GPU. Should be maybe defined 'batch_size' when enabledataparallelism is turned on as 'batch size per gpu'?
Just to repeat the above table for 4 GPU's:

batchSize | time (HH:MM) 
:---: | :---:
32 | 14:06 
128 | 8:06

gheinrich:
Thanks! I don't understand why training time differs between 1,2,3,4 GPUs when `disableAutoDataParallelism=true`. Did you manually encapsulate your model in a `DataParallelTable` in those cases?

In DIGITS we are doing "strong scaling" ([ref](https://en.wikipedia.org/wiki/Scalability#Weak_versus_strong_scaling)) to keep the optimization problem exactly the same. If you increase the overall batch size then you're training something slightly different.

TimZaman:
I don't understand that either. I used vanilla GoogleNet from this PR. I might try running it again. But seriously i just make this queue in one go and didn't touch the server in the meantime.. Must be that planetary alignment thing.

gheinrich:
@TimZaman's data points from https://github.com/NVIDIA/DIGITS/pull/828#issuecomment-225548141 are very interesting. I ran a few more tests to double check some that seemed somewhat unexpected to me. Using my poor man's 2-TitanX workstation, the Vanilla GoogleNet and VOC dataset and doing 2 training epochs + 2 validations:

#GPU's | disableAutoDataParallelism | batchSize | time (MM:SS)
:---: | :---: | :---: | :---: | :---
1 |  false | 32 | 10:50
1 |  true | 32 | 10:56 | 
2 |  false | 32 | 8:22
2 |  true | 32 | 11:01
2 |  true | 64 | 7:10
2 |  true | 128 | 6:34

This looks more like I expected i.e.
- it doesn't matter what you set `disableAutoDataParallelism` to when you have only one GPU,
- when `disableAutoDataParallelism=true` it doesn't matter how many GPUs you have,
- with more than 1 GPU and `disableAutoDataParallelism=false` you start seeing a scaling benefit (on GoogLeNet),
- increasing batch size help speed things up though it changes the optimization dynamics and I did notice that the loss was going down more slowly when using a larger batch size (mostly because I didn't change the learning rate I suppose).

TimZaman:
That looks a lot more sensible indeed, i wish mine did too but i cant help it. changing the parameters dependent on ngpu can still be considered strong scaling if you define the 'total problem' as 'an epoch' instead of defining the problem as 'one batch'.
And as you have just experienced, with the current implementation you are still training something slightly different; are the gpu-specific gradients averaged?

gheinrich:
With strong scaling (keeping the total number of processed samples the same per iteration) we are solving exactly the same problem as all gradients are added together so it doesn't matter which GPU they are calculated on. This means we can keep the learning rate the same and if we initialize the random seed in the same way we will converge towards the same solution. That is the theory though in practice I did notice a tiny difference between runs (even on single GPU). I am going to try and see where it's coming from. The reproducibility of results is very important to some people. 

If you change the batch size then you increase the number of samples between each update of the parameters and even if you update the learning rate you will converge towards a pretty different solution.

TimZaman:
Hey Greg do you know if indeed the GPUs have thermal throttling? That'd explain some inconsistent results I had, given it was running on the office devbox with warm weather, etc.

gheinrich:
I am no expert in that area but in my understanding the clock frequency can be increased up to 1075MHz with GPU Boost if the temperature is low however the base frequency (1GHz) is guaranteed. This shouldn't affect performance tremendously unless the CPU does get throttled.

gheinrich:
I will merge this since it has been reviewed and tested and we're a long way from the next release.

PR #832, opened 2016-06-10
Remove cropping from validation phase in standard networks

gheinrich:
Random cropping during validation phase makes it impossible to reproduce results on the validation set during testing.

### DIFF
--- a/digits/standard-networks/caffe/alexnet.prototxt
+++ b/digits/standard-networks/caffe/alexnet.prototxt
@@ -19,9 +19,6 @@ layer {
   type: "Data"
   top: "data"
   top: "label"
-  transform_param {
-    crop_size: 227
-  }
   data_param {
     batch_size: 32
   }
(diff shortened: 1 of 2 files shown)


lukeyeager:
This doesn't actually change anything because of these lines:
https://github.com/NVIDIA/DIGITS/blob/v4.0.0-rc.1/digits/model/tasks/caffe_train.py#L280
https://github.com/NVIDIA/DIGITS/blob/v4.0.0-rc.1/digits/model/tasks/caffe_train.py#L293-L295

The val data layer uses the same crop size as the train data layer.

And this solution won't work in general, anyway. The train data blob needs to be the same shape as the val data blob. If you remove the cropping, then the data layer will use the same blob size as exists in the LMDB. Right?

gheinrich:
Ha, I tested it by removing `crop_size` to both the training and validation sets! Let me see if I can revisit this one but I think I'll just need to close this PR. Thanks!

PR #833, opened 2016-06-10
Add option to override default object detection classes

gheinrich:


lukeyeager:
Looks fine and should make DetectNet more generally useful. Have you tested it?

gheinrich:
I tested using the usual dataset and overriding the class list to get a different index for "car" (i.e. not `1`). Then I trained Detectnet with the updated class mapping and this looked OK.

douzsh:
Hi, I used 0.15 version NVcaffe on k80 for detectNet training with KITTI data.
I checked the LMDB generated from DIGITS with no problem.
With the example/kitti/train_val.prototxt I trained the detectNet. mAP, recall & precision are always 0. and the result is also all 0 boxes. Can you Help me?

PR #834, opened 2016-06-13
Improve generic DB creation error reporting

gheinrich:
This moves the reporting of errors that occurred in a reader thread to just after the thread completes.
Doing this avoids dropping an error when trying to add together arrays of different shapes if a reader thread errored because the feature/label shape was not constant.

Assertions are replaced with more meaningful exceptions (helps display the real error in DIGITS UI i.e. instead of showing `AssertionError` we are now showing `ValueError: shape mismatch ...`).

### DIFF
--- a/tools/create_generic_db.py
+++ b/tools/create_generic_db.py
@@ -147,9 +147,11 @@ class LmdbWriter(DbWriter):
         datums = []
         for (feature, label) in batch:
             # restrict features to 3D data (Caffe Datum objects)
-            assert feature.ndim == 3, "LMDB/Caffe expect 3D data"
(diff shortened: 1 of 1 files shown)


lukeyeager:
Good stuff. How do I test it? Can you give me an example input that will throw an error into `error_queue`?

gheinrich:
You can test with #835: clear all the padding and resize fields and try to create an object detection dataset. If images are not all the same size, this drops an obscure error because it's trying to add together tensors of different sizes when computing the mean.

lukeyeager:
Got it, thanks for the tip. LGTM.

generic-dataset-no-padding-error

PR #835, opened 2016-06-13
Make padding optional in object detection data extension

gheinrich:
This change makes it optional to pad images when creating an object detection dataset.

A form validator is added to verify that either both or none of width/height are specified when validating the form.
Doing this in the form validator instead of (like formerly) the extension instantiation helps drop a more user-friendly error.

Template updated to show width/height on same row.


gheinrich:
Depends on #833

lukeyeager:
Looks fine and should make DetectNet more generally useful.

PR #839, opened 2016-06-14
[Reopen Pull Request #823] Added Delete Snapshot Button

cagbal:
I closed the pull request #823 before, since it could not pass the test. I reopen it by changing some parts according to comments of @gheinrich, @lukeyeager.

I added deleting a specific snapshot of a model on the model page, since my harddisk sometimes runs out of space when so many model are saved. So, I put a Delete Snapshot button under Download Model Button.
The deletion process does not allow: 

- deleting all snapshots (in that case, it becomes deletion of all model)
- deleting the last snapshot 

Now, the model page seems like:

screenshot from 2016-06-14 17 21 37

### DIFF
--- a/digits/model/views.py
+++ b/digits/model/views.py
@@ -7,6 +7,7 @@ import json
 import math
 import tarfile
 import zipfile
+import os
 
 import flask
 import werkzeug.exceptions
@@ -30,8 +31,14 @@ def show(job_id):
         {id, name, directory, status, snapshots: [epoch,epoch,...]}
     """
     job = scheduler.get_job(job_id)
+
     if job is None:
         raise werkzeug.exceptions.NotFound('Job not found')
+
+    task = job.train_task()
+
+    # Update the snapshots
+    task.detect_snapshots()
 
     related_jobs = scheduler.get_related_jobs(job)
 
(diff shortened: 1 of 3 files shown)


TimZaman:
Hmm. It would be really great if you can add something that allows for deletion of all snapshots., or better, all but the last one. Sometimes i run something awesome and slow for 100 epochs, only to find out i have left the default '1 snapshot/epoch' on. And especially in that case you've got a big problem with disk usage. Also, how about listing the amount of storage used per job in the front page UI table? I've got hundreds of jobs there and it's hard figuring out which one to clean wrt disk size.

cagbal:
Deleting all snapshots is a great idea! Thanks! This button may be replaced by it. I am actually deleting all snapshots except the last one for the same reasons as you said. However, I wanted to give some flexibility to user(s/he may not be in same situation).  But, adding an additional delete button probably will push the current UI to down. I'll try it and put a screenshot here.

digits/model/views.py
```diff
@@ -30,9 +31,15 @@ def show(job_id):
...
     """
     job = scheduler.get_job(job_id)
+
     if job is None:
         raise werkzeug.exceptions.NotFound('Job not found')
 
+    task = job.train_task()
+
+    # Update the snapshots
+    task.detect_snapshots()
```
gheinrich: could you do this in `delete_snapshot()` instead?

gheinrich:
Alternatively there could be a `Delete snaphots` button which spawns a similar dialog to when you "delete model" from which you could select the snapshots to delete, with an option to "delete all but last one". That could be done using an AJAX query, without reloading the main page.

digits/model/views.py
cagbal: Sure, I've done it.

cagbal:
Honestly, I don't actually like refreshing the page, but I could not do it with Ajax query. I'll try.

TimZaman:
Greg's idea is great, make a simple button to the right of 'Download Model' on the same row. That spawns a small window where you can choose from three options: '[all but current; currently selected; all]'.
This also makes sure we have a clean interface and the button only says 'Delete'. I don't think this should take up a whole other row.

gheinrich:
I'm sure you'll find AJAX fun to use :-). There is an example [there](https://github.com/NVIDIA/DIGITS/blob/v4.0.0-rc.1/digits/templates/job.html#L297-L317)

cagbal:
:) Thanks for example! After I complete the changes, I'll again put a screenshot here.

lukeyeager:
Closing due to inactivity. Please re-open this if you want to continue working on it. Otherwise hopefully somebody else will pick up the ball and move ahead with this.

PR #840, opened 2016-06-14
Set GPU mode explicitly

IsaacYangSLA:
The solver_mode value in solver.prototxt seems not parsed in pycaffe.  Hence, even that parameter is set to GPU and we also call set_device to one of available GPU's, pycaffe still uses CPU for most of its computation.
To avoid that, in addition to set_device, caffe.set_mode_gpu() has to be called explicitly right after set_device() to enable GPU mode.
Additionally, using sys.executable instead 'python' to make the arguments more portable.

### DIFF
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -860,7 +860,10 @@ class CaffeTrainTask(TrainTask):
             solver_type = solver_type_mapping[self.solver_type]
         except KeyError:
             raise ValueError("Unknown solver type {}.".format(self.solver_type))
-        gpu_script = "caffe.set_device({id});".format(id=gpu_id) if gpu_id else ""
+        if gpu_id is not None:
+            gpu_script = "caffe.set_device({id});caffe.set_mode_gpu();".format(id=gpu_id)
+        else:
+            gpu_script = "caffe.set_mode_cpu();"
         if self.pretrained_model:
             weight_files = ','.join(map(lambda x: self.path(x), self.pretrained_model.split(':')))
             loading_script = "solv.net.copy_from('{weight}');".format(weight=weight_files)
@@ -875,7 +878,7 @@ class CaffeTrainTask(TrainTask):
             .format(gpu_script=gpu_script,
                     solver=solver_type,
                     solver_file = self.solver_file, loading_script=loading_script)
-        args = ['python -c '+ '\"' + command_script + '\"']
+        args = [sys.executable + ' -c '+ '\"' + command_script + '\"']
         return args
 
     @override


lukeyeager:
Doesn't `set_mode_gpu` belong in `gpu_script`?

Looks good otherwise.

IsaacYangSLA:
Yes, that looks better.

digits/model/tasks/caffe_train.py
```diff
@@ -860,7 +860,10 @@ def _pycaffe_args(self, gpu_id):
             solver_type = solver_type_mapping[self.solver_type]
         except KeyError:
             raise ValueError("Unknown solver type {}.".format(self.solver_type))
-        gpu_script = "caffe.set_device({id});".format(id=gpu_id) if gpu_id else ""
+        if gpu_id:
```
lukeyeager: This looks a little suspect. If `gpu_id == 0`, then this will return False, even though GPU `0` is a valid choice.

How about `if gpu_id is not None:`? Even if it does the same thing, it's a little more clear what types of values you're expecting.

digits/model/tasks/caffe_train.py
```diff
@@ -860,7 +860,10 @@ def _pycaffe_args(self, gpu_id):
...
+            gpu_script = "caffe.set_device({id});caffe.set_mode_gpu();".format(id=gpu_id)
+        else:
+            gpu_script = "caffe.set_mode_gpu();"
```
lukeyeager: Would it make more sense to set the mode to CPU here?

digits/model/tasks/caffe_train.py
IsaacYangSLA: Seems that gpu_id is a string, not an int.  But you are right, testing it against None is a better way to understand the program logic.

digits/model/tasks/caffe_train.py
IsaacYangSLA: My bad.  I didn't explain that portion clearly.  When gpu_id == None, it means either the resources['gpu'] is empty or 'gpu' is not in resources.  Those two conditions do not set mode to CPU in regular path of task_arguments.  I tried to match that behavior in _pycaffe_args.

lukeyeager: > Those two conditions do not set mode to CPU in regular path of task_arguments. I tried to match that behavior in _pycaffe_args.

Technically this is how we do it for Caffe training:
https://github.com/NVIDIA/DIGITS/blob/v4.0.0-rc.1/digits/model/tasks/caffe_train.py#L422-L427

But that's pretty sloppy. In this case, I think it makes sense to set the mode to CPU if there is no GPU assigned.

IsaacYangSLA: @lukeyeager , thanks for pointing that to me.  Setting the mode to CPU when gpu_id is None makes sense to me, too.

PR #841, opened 2016-06-15
removed forced slash when using autocomplete

### LINKED ISSUE #838: Windows autocomplete broken by #773
Reported on https://github.com/NVIDIA/DIGITS/issues/47#issuecomment-225852787

@Lucaszw do you mind having a look? Thanks!

Lucaszw:
*Fix #838*
*Comes from #773*

@IsaacYangSLA Would you be able to try this out for me on windows? Thanks!

### DIFF
--- a/digits/templates/layout.html
+++ b/digits/templates/layout.html
@@ -77,16 +77,10 @@ window.onload = function () {
                return base;
             }
             return baseName(suggestion.value);
-        }
+        },
+        minChars: 0
     });
     $(".autocomplete_path").removeAttr("autocomplete");
-    $(".autocomplete_path").on('keyup',function(e){
-      // If path is relative, enforce absolute
-      var isRelative = e.target.value[0] != '/' && e.target.value[0] != '\\';
-      if (e.target.value.length > 0 && isRelative) {
-        e.target.value = "/" + e.target.value;
-      }
-    });
 };
 
 </script>
--- a/digits/views.py
+++ b/digits/views.py
@@ -564,7 +564,7 @@ def path_autocomplete():
         # Only allow absolute paths by prepending forward slash
         path = os.path.sep + path
 
-    suggestions = glob.glob(path+"*")
+    suggestions = [os.path.abspath(p) for p in glob.glob(path+"*")]
     if platform.system() == 'Windows':
         # on windows, convert backslashes with forward slashes
         suggestions = [p.replace('\\', '/') for p in suggestions]


digits/templates/layout.html
```diff
@@ -77,15 +77,8 @@
...
-    });
-    $(".autocomplete_path").removeAttr("autocomplete");
-    $(".autocomplete_path").on('keyup',function(e){
-      // If path is relative, enforce absolute
-      var isRelative = e.target.value[0] != '/' && e.target.value[0] != '\\';
-      if (e.target.value.length > 0 && isRelative) {
-        e.target.value = "/" + e.target.value;
-      }
+        },
+        minChars: 0
```
gheinrich: why did you remove `$(".autocomplete_path").removeAttr("autocomplete");`?

Lucaszw: My apologies! That was a slip-up on my part (the line is re-added now).

IsaacYangSLA: The fix seems not working well.  If I type /working/, I could see a list of subfolders and files under c:/working/  (my digits is installed on drive c:).
However, if I type c:/working/, there is no list shown.  At the same time, browser showed two requests, path?query=c and path?query=c%3A.  Nothing else.

Lucaszw: Interesting... I also changed the minChars count in addition to removing the problematic line for @JVR32. @IsaacYangSLA  Would you be able to check for me if removing `minChars: 0`  fixes the problem for you? Thanks!

IsaacYangSLA:
The autocomplete feature seems different in Windows.
If I type E:\, I get a list of folders/files under E drive.  If I type e:\, C:\ or c:/, nothing.
Replacing \ with /, i.e. E:/, e:/, C:/, c:/  all show nothing.

If I type E:\ and select one folder (data_e), the string changes to E:/data_e, and no autocomplete.

Lucaszw:
All suggested paths should return absolute now, so this should hopefully fix the issue. @IsaacYangSLA Do you mind checking this once more and see if you have any better luck? Thanks

IsaacYangSLA:
@Lucaszw , this is what I saw on Windows after applying e9f907e.
image

C:/, C:\, c:/, c:\ does not return autocomplete list.

The following is shown when I do not type anything.
image

However, if I clicked one folder, the autocomplete didn't show anything inside that folder.

Lucaszw:
Thanks @IsaacYangSLA . Could you try for me selecting an option with the arrow keys and entering a forward slash and seeing if it returns anything?

IsaacYangSLA:
With 0e39e4f, after I clicked one folder, nothing happened.  However, after I type '/' at the end of the folder name, the console immediately showed the list, and the html page showed the same list.

image


The following is the initial state.  Sorry the console showed the list two times.  It's my fault.

image

digits/templates/layout.html
```diff
@@ -71,22 +71,17 @@
        serviceUrl: '/autocomplete/path',
        formatResult: function (suggestion, currentValue)
        {
+         console.log(suggestion);
```
lukeyeager: Please remove debug statement

lukeyeager:
> With 0e39e4f, after I clicked one folder, nothing happened. However, after I type '/' at the end of the folder name, the console immediately showed the list, and the html page showed the same list.

Yeah, that's the same old behavior we've always had.

Lucaszw:
Alright I think things are good to go. The debug statement has been removed.

lukeyeager:
LGTM!

PR #843, opened 2016-06-15
Fix the table to fit in the container

jmancewicz:
The wells containing the Models and Datasets tables were not strictly in the container div so that as the window width was decreased, the wells could creep off screen before the divs were resized.


jmancewicz:
@lukeyeager, this should do it.

lukeyeager:
Thanks!

PR #846, opened 2016-06-16
Keep the config button and popup within the container width

jmancewicz:
The config menu and button could creep out of the container. The menu could get so far out that it wasn't readable.


lukeyeager:
Seems like it's working. Thanks!

PR #847, opened 2016-06-16
Double sort with blanks fix

jmancewicz:
When sorting with two columns (click on a column header and shift-click a second) the blank entries in the second column could be at the top or the bottom of that column depending on the ascending / descending state.

I've replaced the predicte functions and empty_to_the_end filter with a custom sort filter.


jmancewicz:
@lukeyeager can you verify with your model list?

lukeyeager:
Works. Thanks!

PR #850, opened 2016-06-16
Always show view extension header

gheinrich:
Always show header, including when doing single image inference.
Rename "summary_template" to "header_template".


lukeyeager:
LGTM

Issue #849, opened 2016-06-16
Confusion between Windows paths and multi pre-trained networks


gheinrich:
Reported on https://github.com/NVIDIA/DIGITS/issues/49#issuecomment-226422731

PR #851, opened 2016-06-16
Fixes for multiple pretrained model files

lukeyeager:
Close https://github.com/NVIDIA/DIGITS/issues/849.

See explanation at https://github.com/NVIDIA/DIGITS/pull/498#issuecomment-226550671.

### DIFF
--- a/digits/model/tasks/torch_train.py
+++ b/digits/model/tasks/torch_train.py
@@ -250,7 +250,10 @@ class TorchTrainTask(TrainTask):
             args.append('--type=float')
 
         if self.pretrained_model:
-            args.append('--weights=%s' % self.path(self.pretrained_model))
+            filenames = self.pretrained_model.split(os.path.pathsep)
+            if len(filenames) > 1:
+                raise ValueError('Torch does not support multiple pretrained model files')
+            args.append('--weights=%s' % self.path(filenames[0]))
 
         return args
 
--- a/digits/model/forms.py
+++ b/digits/model/forms.py
@@ -273,10 +273,9 @@ class ModelForm(Form):
 
     def validate_custom_network_snapshot(form, field):
         if form.method.data == 'custom':
-            snapshot = field.data.strip()
-            if snapshot:
-                if not os.path.exists(snapshot):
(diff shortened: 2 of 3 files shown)


lukeyeager:
@gheinrich can you take a look?

digits/model/tasks/torch_train.py
```diff
@@ -250,7 +250,10 @@ def task_arguments(self, resources, env):
             args.append('--type=float')
 
         if self.pretrained_model:
-            args.append('--weights=%s' % self.path(self.pretrained_model))
+            filenames = self.pretrained_model.split(os.path.pathsep)
+            if len(filenames) > 1:
+                raise ValueError('Torch does not support multiple pretrained model files')
+            args.append('--weights=%s' % self.path(filenames[0]))
```
gheinrich: that looks good though (probably not due to this) it seems the error came too quickly to show in the UI, as if we hadn't yet created the socketIO interface. The error message only showed after I reloaded the page.
I tested that it works when only one pretrained model is provided.

gheinrich:
This looks good to me. On Linux this is working for me. It's difficult for me to test on Windows as I don't have the set-up. I just verified that `os.pathsep` returns `;` on Windows.

IsaacYangSLA:
I will test it here on Windows.

IsaacYangSLA:
Did the new image classification model page remove pre-trained weights field?  I couldn't find it on 4.0.0-rc.2.

lukeyeager:
> Did the new image classification model page remove pre-trained weights field?

Look under the "Custom Network" tab

IsaacYangSLA:
Thanks.  My bad.  Still a lot to learn about DIGITS.
I tested on Windows.  Separating multiple weight files with ';' and path starts with /, both caffe.exe and pycaffe can load those weights.  However, there is a bug in _pycaffe_args, so I need to send a PR on loading_script portion of _pycaffe_args.

lukeyeager:
Great, so we're good? Please merge if so.

IsaacYangSLA:
Looks great.  I believe it works good in Windows now.

PR #854, opened 2016-06-17
Improved separation with multiple network paths

IgorX2:
Now using `os.path.pathsep` instead of ':' as per comment in #498.


lukeyeager:
Duplicate of #851.

PR #855, opened 2016-06-17
Fix loading multi weight files in _pycaffe_args

IsaacYangSLA:
We have to loop over each pre-trained weight file when loading those files.
This fix depends on PR#851.  I will rebase this after PR#851 is merged.

### DIFF
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -864,11 +864,11 @@ class CaffeTrainTask(TrainTask):
             gpu_script = "caffe.set_device({id});caffe.set_mode_gpu();".format(id=gpu_id)
         else:
             gpu_script = "caffe.set_mode_cpu();"
+        loading_script = ""
         if self.pretrained_model:
-            weight_files = ','.join(map(lambda x: self.path(x), self.pretrained_model.split(os.path.pathsep)))
-            loading_script = "solv.net.copy_from('{weight}');".format(weight=weight_files)
-        else:
-            loading_script = ""
+            weight_files = map(lambda x: self.path(x), self.pretrained_model.split(os.path.pathsep))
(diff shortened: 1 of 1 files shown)


lukeyeager:
> I will rebase this after PR#851 is merged.

It's merged

IsaacYangSLA:
Thanks.  Just rebased it and force pushed to my fork.

digits/model/tasks/caffe_train.py
```diff
@@ -864,11 +864,11 @@ def _pycaffe_args(self, gpu_id):
             gpu_script = "caffe.set_device({id});caffe.set_mode_gpu();".format(id=gpu_id)
         else:
             gpu_script = "caffe.set_mode_cpu();"
+        loading_script = ""
         if self.pretrained_model:
-            weight_files = ','.join(map(lambda x: self.path(x), self.pretrained_model.split(os.path.pathsep)))
-            loading_script = "solv.net.copy_from('{weight}');".format(weight=weight_files)
-        else:
-            loading_script = ""
+            weight_files = map(lambda x: self.path(x), self.pretrained_model.split(os.path.pathsep))
```
lukeyeager: Should we leave the `if self.pretrained_model` statement in? What if it's `None`?

IsaacYangSLA: I think we need that if statement.  If it's None and we try to perform map on None, it throws exception.
The loading_script is init'ed as an empty string, before the if condition.  Hence, when self.pretrained_model is None, loading_script is still "", and the final script does not include statements to load pre-trained weights.

lukeyeager: Oh sorry, I thought you had deleted it. Nevermind!

lukeyeager:
Looks good. Thanks!

PR #856, opened 2016-06-17
fix table width and reformat the filter and buttons

jmancewicz:
When the Model and Dataset table widths were corrected, the Running Jobs table was not. This makes similar changes to the Running Jobs.


lukeyeager:
The 100 lines looks scary, but when you ignore whitespace changes it's only 6 LOC.

LGTM.

PR #857, opened 2016-06-17
Don't sanity check val phase if val set not present (bis)

gheinrich:
The generic path was left out in #655.
This fixes it and adds tests for the generic path.


lukeyeager:
Thanks for the fix and the tests!

PR #859, opened 2016-06-17
fix colspan and filter width on chrome

jmancewicz:


lukeyeager:
This is two UI fixes:
* Fix the width of the filter button on Chrome
* Cleanup UI for when no models exist

PR #861, opened 2016-06-18
Fixed the sorting function.

### LINKED ISSUE #853: Job sorting is bugged
Using: 99e2c46fa4babd2e278f6551c9f779e127119b77

Whenever I sort the Digits homepage, some jobs always stay at a given position.
See screenshots for example.
2016-06-16 23_25_30-digits
2016-06-16 23_25_11-digits

jmancewicz:
*Close #853*

The sorting function is fixed. It needed to return -1, 0, -1, but had been returning true, false. The Boolean return appeared to work in FireFox, but that may have been a coincidence.

Also changed the default descending choice when a column is first selected based on whether the columns data type is a string on not. This has strings sort a->z and numbers and dates (large -> small). This is similar to OSX file browser. This gives the "logical" choice and feels natural.

### DIFF
--- a/digits/static/js/home_app.js
+++ b/digits/static/js/home_app.js
@@ -313,9 +313,9 @@
         $scope.search_text = '';     // set the default search/filter term
         $scope.sort = {
             active1: 'submitted',
-            descending1: true,
+            descending1: 1,
             active2: 'submitted',
-            descending2: true,
+            descending2: 1,
         }
         $scope.jobs = [];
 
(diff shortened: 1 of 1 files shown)


jmancewicz:
@lukeyeager can you have a look when you have time?

lukeyeager:
I've verified that this PR fixes sorting behavior on Chrome. And I like the string vs number sorting change.

Thanks for the fix @jmancewicz and for the report @enoonIT!

PR #870, opened 2016-06-25
Fix handling of long multi-field documents

skje:
This fixes cases when a document has multiple, long fields, and reaches the FEATURE_LEN limit before the last field. This causes it to break out of the inner loop, but not the outer loop and raises an IndexError.

### DIFF
--- a/examples/text-classification/create_dataset.py
+++ b/examples/text-classification/create_dataset.py
@@ -64,6 +64,8 @@ def create_dataset(folder, input_file_name, db_batch_size=None, create_images=Fa
             sample = np.ones(FEATURE_LEN) # one by default (i.e. 'other' character)
             count = 0
             for field in row['fields']:
+                if count > FEATURE_LEN-1:
+                    break
                 for char in field.lower():
                     if char in cdict:
                         sample[count] = cdict[char]


lukeyeager:
Why do you prefer crashing to truncating?

examples/text-classification/create_dataset.py
```diff
@@ -64,6 +64,8 @@ def create_dataset(folder, input_file_name, db_batch_size=None, create_images=Fa
             sample = np.ones(FEATURE_LEN) # one by default (i.e. 'other' character)
             count = 0
             for field in row['fields']:
+                if count > FEATURE_LEN-1:
+                    break
```
gheinrich: this looks OK to me. Alternatively we could do:

```python
for char in field.lower():
    if count < FEATURE_LEN:
        if char in cdict:
            sample[count] = cdict[char]
    else:
        break
    count += 1
```

so we'd have only one condition on the length of the `sample` vector

```

lukeyeager:
@gheinrich can you either close this or merge it? As I understand it, it won't matter too much either way.

gheinrich:
Hi @skje sorry for the delay, do you mind sending a [CLA](https://github.com/NVIDIA/DIGITS/tree/digits-5.0#bugs-and-feature-requests)?

gheinrich:
Closing due to inactivity.

PR #871, opened 2016-06-27
Update doc to point to cudnn 5.1

IsaacYangSLA:
In Windows environment, document originally points to old dependency (cudnn 4).  Update it to recommend using the latest library (cudnn 5.1).


IsaacYangSLA:
I built Windows Caffe with cudnn 5.1 RC and tested it on one training session.  No build error and no runtime error.  Training results also looked OK.  I will merge this PR soon it since it's a document update, and the library worked OK.

PR #877, opened 2016-06-30
Addition of mean_path selection

djoubert:
Currently, when applying mean subtraction during the training of a model, the user is only able to use the mean image or pixels from the dataset on which the model is being trained. This pull request adds the option of specifying a path to a protoblob file containing a different mean image. 

This feature is useful when doing transfer learning: An already trained model is adapted for use with a new target dataset, but it is desired to maintain interoperability with the original source dataset. If the mean of the target dataset is significantly different from the source dataset, the performance of the new model on the source dataset will be needlessly negatively impacted by this discrepancy. By specifying the mean_path to be used for mean subtraction as the protoblob of the source dataset, the new model will still be applicable to the original source dataset.

### DIFF
--- a/digits/model/images/classification/views.py
+++ b/digits/model/images/classification/views.py
@@ -267,6 +267,7 @@ def create():
                         pretrained_model = pretrained_model,
                         crop_size = form.crop_size.data,
                         use_mean = form.use_mean.data,
+                        mean_path = form.mean_path.data,
                         network = network,
                         random_seed = form.random_seed.data,
                         solver_type = form.solver_type.data,
--- a/digits/model/images/classification/views.py
+++ b/digits/model/images/classification/views.py
@@ -267,7 +267,7 @@ def create():
                         pretrained_model = pretrained_model,
                         crop_size = form.crop_size.data,
                         use_mean = form.use_mean.data,
-                        mean_path = form.mean_path.data,
+                        custom_mean_path = form.custom_mean_path.data,
                         network = network,
                         random_seed = form.random_seed.data,
                         solver_type = form.solver_type.data,
--- a/digits/model/job.py
+++ b/digits/model/job.py
@@ -74,9 +74,12 @@ class ModelJob(Job):
         should be assumed that the mean image will be that of the dataset.
         returns: True if the custom_mean_path was successfully set, False otherwise
         """
-        if custom_mean_path != '' and path.isfile(custom_mean_path) is True:
-            self.custom_mean_path = custom_mean_path
-            return True
+        if custom_mean_path != '' and custom_mean_path is not None:
+            if path.isfile(custom_mean_path) is True:
+                self.custom_mean_path = custom_mean_path
+                return True
+            else:
+                return False
         else:
             return False
 
@@ -86,12 +89,14 @@ class ModelJob(Job):
         user specified path
         """
 
-        if self.dataset is None:
-            self.load_dataset()
-
-        if self.custom_mean_path is None:
-            # The user never set the mean path, so the dataset's mean path should be used
-            return self.dataset.path(self.dataset.get_mean_file())
-        else:
+        if self.custom_mean_path is not None:
             # The user has specified the file path to use
-            return self.dataset.path(self.custom_mean_path)
+            return self.custom_mean_path
+
+        if self.dataset is None:
+            # Cannot retrieve the mean path, as it isn't loaded anymore
+            return None
+        else:
+            # The user never set the mean path, so the dataset's mean path should be used
+            if self.custom_mean_path is None:
+                return self.dataset.path(self.dataset.get_mean_file())
--- a/digits/model/job.py
+++ b/digits/model/job.py
@@ -1,6 +1,8 @@
 # Copyright (c) 2014-2016, NVIDIA CORPORATION.  All rights reserved.
 from __future__ import absolute_import
 
+from os import path
+
 from . import tasks
 from digits.job import Job
 from digits.utils import override
@@ -20,6 +22,8 @@ class ModelJob(Job):
         """
         super(ModelJob, self).__init__(**kwargs)
         self.pickver_job_dataset = PICKLE_VERSION
+
+        self.custom_mean_path = None
 
         self.dataset_id = dataset_id
         self.load_dataset()
@@ -63,3 +67,28 @@ class ModelJob(Job):
         These files get added to an archive when this job is downloaded
         """
         return NotImplementedError()
+
+    def set_custom_mean_path(self, custom_mean_path):
+        """
+        Set the path to a custom mean image protoblob to be used by the model job. If the custom mean path is None, it
+        should be assumed that the mean image will be that of the dataset.
+        returns: True if the custom_mean_path was successfully set, False otherwise
+        """
+        if custom_mean_path != '' and path.isfile(custom_mean_path) is True:
+            self.custom_mean_path = custom_mean_path
+            return True
+        else:
+            return False
+
+    def get_mean_path(self):
+        """
+        returns the file path to the mean image protoblob\jpg\file, which is either the dataset's (default) or some
+        user specified path
+        """
+
+        if self.custom_mean_path is None:
+            # The user never set the mean path, so the dataset's mean path should be used
+            return self.dataset.path(self.dataset.get_mean_file())
+        else:
+            # The user has specified the file path to use
+            return self.dataset.path(self.custom_mean_path)
--- a/digits/model/job.py
+++ b/digits/model/job.py
@@ -86,6 +86,9 @@ class ModelJob(Job):
         user specified path
         """
 
+        if self.dataset is None:
+            self.load_dataset()
+
         if self.custom_mean_path is None:
             # The user never set the mean path, so the dataset's mean path should be used
             return self.dataset.path(self.dataset.get_mean_file())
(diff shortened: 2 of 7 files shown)


lukeyeager:
Hi @djoubert, thanks for the contribution! We met at ICML, right?

1. At first glance, it doesn't look like this works for Torch
2. You'll need to [send us a CLA](https://github.com/NVIDIA/DIGITS/blob/master/.github/CONTRIBUTING.md#pull-requests) before I can accept this

djoubert:
Hi Luke! No, I wasn't at ICML. You met my boss, Hannes Naude, he's asked me to handle the pull request. 

1. I'll check how the mean_path should be used when working with Theano, and commit when I have a solution.
2. I've sent the CLA to digits@nvidia.com, let me know if you haven't received it, I'll send it again.

TimZaman:
Very interesting observation on the usage of the mean wrt the transfer learning. I think it also makes sense just to have an option to not use mean images at all but just use our fixed [0:255] scale to [-1:1]. 

**2. A bit of help for Torch.**

Torch uses the file that's always called `mean.jpg`, a constant defined in [contants.py#L10](https://github.com/NVIDIA/DIGITS/blob/b06dfaac0eda04ce93168b3ef2a1fed5ec31bc46/digits/utils/constants.py#L10)
Then if you'd be so kind, change a minor typo here: [torch_train.py#L518](https://github.com/djoubert/DIGITS/blob/d02475a94081b32e7268c21a67d937821348e477/digits/model/tasks/torch_train.py#L518) and [torch_train.py#L523](https://github.com/djoubert/DIGITS/blob/d02475a94081b32e7268c21a67d937821348e477/digits/model/tasks/torch_train.py#L523) to make the three usage of the mean similar again within the document.
```py
filename = self.create_mean_file()
args.append('--mean=%s' % os.path.join(self.job_dir, constants.MEAN_FILE_IMAGE))
```
to 
```py
filename = self.create_mean_file()
args.append('--mean=%s' % filename)
```
This is because `os.path.join(self.job_dir, constants.MEAN_FILE_IMAGE)` is already done in the `create_mean_file()` function.
Now finally the only thing you have to do now is change the mean file that is used when needed at [torch_train.py#L112](https://github.com/djoubert/DIGITS/blob/d02475a94081b32e7268c21a67d937821348e477/digits/model/tasks/torch_train.py#L112), and maybe use the same function you have used in `caffe_train.py`. So therefore move up your function `def get_mean_path(self):` to the parent class `train.py`.

digits/model/tasks/caffe_train.py
```diff
@@ -223,6 +223,16 @@ def set_mean_file(self, layer, mean_file):
 
         layer.transform_param.mean_file = mean_file
 
+    def get_mean_path(self):
```
TimZaman: Maybe move this up to the parent class `train.py` to use this in Torch as well.

digits/model/tasks/torch_train.py
```diff
@@ -112,7 +112,7 @@ def create_mean_file(self):
         filename = os.path.join(self.job_dir, constants.MEAN_FILE_IMAGE)
         # don't recreate file if it already exists
         if not os.path.exists(filename):
```
TimZaman: Notice here that if `filename` exists, it will in this case always use `constants.MEAN_FILE_IMAGE` and thus the original `mean.jpg` will override anything you might manually supply yourselves. I am now wondering why the constant is used here in the first place and not just `.get_mean_file()`. I think it's because torch uses the jpg and caffe the proto

digits/model/tasks/torch_train.py
```diff
@@ -110,8 +110,8 @@ def before_run(self):
 
     def create_mean_file(self):
         filename = os.path.join(self.job_dir, constants.MEAN_FILE_IMAGE)
-        # don't recreate file if it already exists
-        if not os.path.exists(filename):
+        # don't recreate file if it already exists, unless user has specified a non-db mean path
+        if not os.path.exists(filename) or self.mean_path is not None:
```
gheinrich: I might be missing something but I don't understand the need to change this condition. The mean file must be created once per model job, either from the dataset mean or from the custom mean. In both cases `self.get_mean_path()` returns the correct file. Note that `self.job_dir` is the model job dir. This is different from the dataset job dir.

digits/model/images/classification/views.py
```diff
@@ -267,6 +267,7 @@ def create():
                         pretrained_model = pretrained_model,
                         crop_size = form.crop_size.data,
                         use_mean = form.use_mean.data,
+                        mean_path = form.mean_path.data,
```
gheinrich: I'd rename this to e.g. `custom_mean_path` since it could be confusing to have a `mean_path` field set to `None` when we are using the dataset mean.

digits/model/tasks/train.py
```diff
@@ -58,6 +58,9 @@ def __init__(self, job, dataset, train_epochs, snapshot_interval, learning_rate,
         self.shuffle = kwargs.pop('shuffle', None)
         self.network = kwargs.pop('network', None)
         self.framework_id = kwargs.pop('framework_id', None)
+        self.mean_path = kwargs.pop('mean_path', None)
```
gheinrich: I think it makes more sense to have `mean_path` be a property of the model job rather than a property of the train task. The mean file is used during inference too so it's not just required during training.

digits/model/tasks/caffe_train.py
```diff
@@ -1388,11 +1388,12 @@ def get_transformer(self):
             # XXX see issue #59
             channel_swap = (2,1,0)
 
-        if self.dataset.get_mean_file():
+        if self.get_mean_path() or self.mean_path is not None:
+
             if self.use_mean == 'pixel':
-                mean_pixel = self.get_mean_pixel(self.dataset.path(self.dataset.get_mean_file()))
+                mean_pixel = self.get_mean_pixel(self.dataset.path(self.get_mean_path()))
```
gheinrich: if we're going to use `get_mean_path()` in model jobs I think it makes more sense to do `self.dataset.path(...)` from `get_mean_path()` otherwise we might be calling `self.dataset.path()` on non-dataset paths (which happens to do what we want in the case of an absolute path but we'd better not rely on it).

digits/templates/models/images/classification/new.html
```diff
@@ -366,6 +366,11 @@ <h4 style="display:inline-block;">Python Layers</h4>
                     {{form.use_mean.tooltip}}
                     {{form.use_mean(class='form-control')}}
                 </div>
+                <div class="form-group{{mark_errors([form.mean_path])}}">
```
gheinrich: Question to all reviewers: don't we need to support both server and client-side image files (like we do for Python layers)?

gheinrich:
Thanks for the PR I think it's a nice feature. I suggest we add it to the generic path too.

gheinrich:
> I think it also makes sense just to have an option to not use mean images at all but just use our fixed [0:255] scale to [-1:1]

Good idea. This can be done with a `Power` layer in Caffe or a combination of `nn.MulConstant` and `nn.AddConstant` in Torch though.

TimZaman:
Yes like you did in the LeNet standard network.

digits/model/tasks/torch_train.py
djoubert: I agree with @gheinrich. I'll remove the modified condition.

digits/templates/models/images/classification/new.html
djoubert: Support for client-side images could be implemented, but I guess you would then have to copy the specified image file to some directory on the server. Which directory should be used?

djoubert:
I've moved the get_mean_path() function to the model jobs class, but now it falls over when trying to infer an image classification.

It seems than when "Classifying One Image", the pickle file of the model job is loaded, which then sets the dataset to None (model/job.py, line 42), which then prevents the get_mean_path() function from accessing the self.dataset.path(). 

Does anyone know why the pickle file needs to be loaded on an infererence task? Or why the dataset then needs to be cleared? I'm going to reload the dataset if it isn't available, but this feels like it's going to break something further down the line.

I think it would perhaps be easier if I just gave access to the fork for whoever wants to add to it. How would I go about doing so?

digits/model/job.py
```diff
@@ -63,3 +67,31 @@ def download_files(self):
...
+
+        if self.dataset is None:
+            self.load_dataset()
+
+        if self.custom_mean_path is None:
+            # The user never set the mean path, so the dataset's mean path should be used
+            return self.dataset.path(self.dataset.get_mean_file())
+        else:
+            # The user has specified the file path to use
+            return self.dataset.path(self.custom_mean_path)
```
gheinrich: I think `self.dataset.path(...)` is not required here as the custom path is absolute (right?)

gheinrich:
> Does anyone know why the pickle file needs to be loaded on an infererence task? Or why the dataset then needs to be cleared?

We need to load the model job to figure out the network's input dimensions, the list of model snapshots, etc. The `dataset` field is cleared from the model job object to refrain from saving data that pertain to another job. If this doesn't sound too convoluted you may use the train task's dataset field, which should always point to the right thing. Admittedly, some clean-up would be useful (e.g. perhaps we should load the dataset automatically when loading a pickled model job instead of relying on the scheduler to do so).

lukeyeager:
I'm closing this due to inactivity. I'll treat this as a feature request with a partial implementation. We'll put it on our backlog.

PR #881, opened 2016-07-04
Support variable input size

gheinrich:
Caffe and Torch pre-processors are updated to handle the case where the input image dimensions do not match those of the training set.

Depends on #806 and #880

### DIFF
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -39,6 +39,50 @@ CAFFE_SNAPSHOT_PREFIX = 'snapshot'
 CAFFE_DEPLOY_FILE = 'deploy.prototxt'
 
 @subclass
+class DigitsTransformer(caffe.io.Transformer):
+    """
+    A subclass of caffe.io.Transformer (an old-style class)
+    Handles cases when we don't want to resize inputs
+    """
+    def __init__(self, resize, **kwargs):
+        """
+        Arguments:
+        resize -- whether to resize inputs to the network default
+        """
+        self.resize = resize
+        caffe.io.Transformer.__init__(self, **kwargs)
+
+    def preprocess(self, in_, data):
+        """
+        Preprocess an image
+        See parent class for details
+        """
+        if not self.resize:
+            # update target input dimension such that no resize occurs
+            self.inputs[in_] = self.inputs[in_][:2] + data.shape[:2]
+            # do we have a mean?
+            if in_ in self.mean:
+                # resize mean if necessary
+                if self.mean[in_].size > 1:
+                    # we are doing mean image subtraction
+                    if self.mean[in_].size != data.size:
+                        # mean image size is different from data size
+                        # => we need to resize the mean image
+                        transpose = self.transpose.get(in_)
+                        if transpose is not None:
+                            # detranspose
+                            self.mean[in_] = self.mean[in_].transpose(
+                                np.argsort(transpose))
+                        self.mean[in_] = caffe.io.resize_image(
+                            self.mean[in_],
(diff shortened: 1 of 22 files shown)


digits/model/tasks/caffe_train.py
```diff
@@ -1156,14 +1200,25 @@ def can_view_weights(self):
         return False
 
     @override
-    def infer_one(self, data, snapshot_epoch=None, layers=None, gpu=None):
+    def infer_one(self,
```
lukeyeager: If I can start a useless style discussion for a moment ...

Which do you prefer? I'm not convinced either way.
```py
# fewer lines used
# arguably more natural looking
def hello_world(greeting,
                name):
    print '%s %s!' % (greeting, name)

# fewer total characters uses (less spaces)
# diff is robust to changes in function name
# diff is robust to changes in function arguments
# weird close parens at end
def hello_world(
        greeting,
        name,
):
    print '%s %s!' % (greeting, name)
```
flake8 allows both, which is no help.

digits/model/tasks/caffe_train.py
```diff
@@ -39,6 +39,50 @@
 CAFFE_DEPLOY_FILE = 'deploy.prototxt'
 
 @subclass
+class DigitsTransformer(caffe.io.Transformer):
```
lukeyeager: I'm nervous about what could happen if Caffe makes changes to this class.
https://github.com/BVLC/caffe/commits/master/python/caffe/io.py

digits/model/tasks/caffe_train.py
gheinrich: It's a tough one :-) Like you, I think the first one looks better but the fact that the second one is more robust to diffs is appealing too. If I was forced to cast a vote I think I'd go for the first one (I'm tempted to think that beauty is more important than robust diffs).

digits/model/tasks/caffe_train.py
gheinrich: > Probably the worst that would happen is we wouldn't make use of some new functionality or bugfix in preprocess().

I am calling `caffe.io.Transformer.preprocess()` on line 82 - wouldn't we then benefit from a bug fix in this function?

If `resize==True` (the default) then I believe we are doing exactly the same thing as before. If `resize==False` then in this class's `preprocess()` the dimensions of `self.inputs[in_]` are changed to avoid resizing the input data and if we are doing mean image subtraction then the mean image is resized to the dimensions of the input data.

lukeyeager: Oh sorry - I didn't read carefully.

With this approach, it doesn't look like you're restoring the original values of `self.inputs` and `self.mean`. What's going to happen to the mean if it keeps getting resized on every call to `preprocess`? Or what happens when you want to use the same transformer to do a non-resizing transformation, then use the same object to do a resizing transformation? The image size will be dependent on the last used configuration instead of how the transformer was originally set up.

gheinrich: Inference is running in a dedicated process therefore `resize` always has a fixed value, for a given run. The mean image is only resized if it has a different shape from the image to perform inference on. So if you are doing inference on a batch of images of the same size, mean image resizing should only occur once. On the next inference job everything gets reset so hopefully the issue you described can't occur.

Note: doing inference on a batch of images that have variable size will cause problems. Perhaps this should be listed as a known limitation - supporting this is very tricky.

lukeyeager: Ok yeah, I guess this is fine. The only problem would be if `preprocess()` had a `resize` argument. But if the class has a `resize` argument then we're ok. Nevermind!

PR #889, opened 2016-07-07
Use get_nvml() in tests

lukeyeager:
Use more sophisticated check for nvml in tests.

Thanks to @flx42 for the report!


gheinrich:
Looks good to me!

PR #891, opened 2016-07-07
Get stats & original network def from "Download Model" button

### LINKED ISSUE #228: Provide information file with downloaded models
We should store some information about the model when downloading.

* Caffe version
* Location of deploy, model, mean files (so filenames can change)
* Image dimensions
* Resize mode
* etc.

Lucaszw:
closes #228

### DIFF
--- a/digits/model/images/job.py
+++ b/digits/model/images/job.py
@@ -1,7 +1,10 @@
 # Copyright (c) 2014-2016, NVIDIA CORPORATION.  All rights reserved.
 from __future__ import absolute_import
 
+import os
+import datetime
 from ..job import ModelJob
+from digits.utils import subclass, override
 
 # NOTE: Increment this everytime the pickled object changes
 PICKLE_VERSION = 1
@@ -17,3 +20,17 @@ class ImageModelJob(ModelJob):
         super(ImageModelJob, self).__init__(**kwargs)
         self.pickver_job_model_image = PICKLE_VERSION
 
+    @override
+    def json_dict(self, verbose=False, epoch=-1):
+        d = super(ImageModelJob, self).json_dict(verbose)
+        task = self.train_task()
+        creation_time = str(datetime.datetime.fromtimestamp(self.status_history[0][1]))
+
+        d.update({
+            "job id": self.id(),
+            "creation time": creation_time,
+            "username": self.username,
+        })
+
+        d.update(task.get_task_stats(epoch))
+        return d
--- a/digits/model/tasks/caffe_train.py
+++ b/digits/model/tasks/caffe_train.py
@@ -33,6 +33,7 @@ PICKLE_VERSION = 3
 
 # Constants
 CAFFE_SOLVER_FILE = 'solver.prototxt'
+CAFFE_ORIGINAL_FILE = 'original.prototxt'
 CAFFE_TRAIN_VAL_FILE = 'train_val.prototxt'
 CAFFE_SNAPSHOT_PREFIX = 'snapshot'
 CAFFE_DEPLOY_FILE = 'deploy.prototxt'
@@ -83,10 +84,15 @@ class CaffeTrainTask(TrainTask):
         self.solver = None
 
         self.solver_file = CAFFE_SOLVER_FILE
+        self.original_file = CAFFE_ORIGINAL_FILE
         self.train_val_file = CAFFE_TRAIN_VAL_FILE
         self.snapshot_prefix = CAFFE_SNAPSHOT_PREFIX
         self.deploy_file = CAFFE_DEPLOY_FILE
         self.log_file = self.CAFFE_LOG
+
+        self.digits_version = digits.__version__
+        self.caffe_version  = config_value('caffe_root')['ver_str']
+        self.caffe_flavor   = config_value('caffe_root')['flavor']
 
     def __getstate__(self):
         state = super(CaffeTrainTask, self).__getstate__()
@@ -228,6 +234,10 @@ class CaffeTrainTask(TrainTask):
         """
         Save solver, train_val and deploy files to disk
         """
+        # Save the origin network to file:
+        with open(self.path(self.original_file), 'w') as outfile:
+            text_format.PrintMessage(self.network, outfile)
+
         network = cleanedUpClassificationNetwork(self.network, len(self.get_labels()))
         data_layers, train_val_layers, deploy_layers = filterLayersByState(network)
 
@@ -522,6 +532,10 @@ class CaffeTrainTask(TrainTask):
         val_label_db_path = self.dataset.get_label_db_path(constants.VAL_DB)
 
         assert train_feature_db_path is not None, 'Training images are required'
+
+        # Save the origin network to file:
+        with open(self.path(self.original_file), 'w') as outfile:
+            text_format.PrintMessage(self.network, outfile)
 
         ### Split up train_val and deploy layers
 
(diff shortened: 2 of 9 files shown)


digits/model/images/job.py
```diff
@@ -2,6 +2,10 @@
 from __future__ import absolute_import
 
 from ..job import ModelJob
+from digits.utils import subclass, override
+
+import json
+import os
```
gheinrich: I have been contaminated by @lukeyeager and will reproduce his comment that PEP8 recommends putting standard package imports first ([see](https://www.python.org/dev/peps/pep-0008/#imports)) :-)

digits/model/images/job.py
```diff
@@ -17,3 +21,65 @@ def __init__(self, **kwargs):
         super(ImageModelJob, self).__init__(**kwargs)
         self.pickver_job_model_image = PICKLE_VERSION
 
+    def get_snapshot_filename(self,epoch=-1):
```
gheinrich: since you're creating this method you might as well move it to `digits/model/tasks/train.py`, replace the code from [there](https://github.com/NVIDIA/DIGITS/blob/v4.0.0-rc.3/digits/model/tasks/torch_train.py#L929-L946) and [there](https://github.com/NVIDIA/DIGITS/blob/v4.0.0-rc.3/digits/model/tasks/caffe_train.py#L1321-L1330) with a call to this new method, and perhaps add unit tests too.

digits/model/images/job.py
```diff
@@ -17,3 +21,65 @@ def __init__(self, **kwargs):
...
+                    break
+        if not snapshot_filename:
+            raise ValueError('Invalid epoch')
+
+        path, name = os.path.split(snapshot_filename)
+
+        return name
+
+    @override
+    def get_job_stats_as_json_string(self,epoch=-1):
```
gheinrich: Why not use `json_dict()`?

digits/model/images/job.py
```diff
@@ -17,3 +21,65 @@ def __init__(self, **kwargs):
...
+            "job id": self.id(),
+            "creation time": self.status_history[0][1],
+            "username": self.username,
+            "image dimensions": task.dataset.get_feature_dims(),
+            "mean file": mean_file,
+            "snapshot file": self.get_snapshot_filename()
+        }
+
+        # Set the torch and caffe specific attributes:
+        if task.name() == 'Train Torch Model':
```
gheinrich: I would move this to the train task. With proper class inheritance you will be able to do Torch-specific stuff in `torch_train.py` and Caffe-specific stuff to `caffe_train.py`. As a rule of thumb, never write framework-specific code outside of framework-specific files - doing this breaks framework agnosticity.

digits/model/views.py
```diff
@@ -189,6 +192,13 @@ def download(job_id, extension):
     elif 'snapshot_epoch' in flask.request.form:
         epoch = float(flask.request.form['snapshot_epoch'])
 
+    # Write the stats of the job to json,
+    # and store in tempfile (for archive)
+    stats = job.get_job_stats_as_json_string(epoch)
+    temp = tempfile.NamedTemporaryFile()
+    temp.write(stats)
+    temp.seek(0)
```
gheinrich: why don't you close the file now?

Lucaszw: I close the file after writing to the tarball as tempfile deletes the file immediately after close

Lucaszw:
Thanks Greg! I implemented your comments.

digits/model/views.py
gheinrich: Ah OK good to know, thanks!

digits/model/images/job.py
```diff
@@ -17,3 +21,16 @@ def __init__(self, **kwargs):
         super(ImageModelJob, self).__init__(**kwargs)
         self.pickver_job_model_image = PICKLE_VERSION
 
+    @override
+    def get_job_stats_as_json_string(self,epoch=-1):
```
gheinrich: Hi, I'm still wondering why you didn't just extend `json_dict()` with the additional bits of information?

digits/model/views.py
```diff
@@ -189,18 +192,20 @@ def download(job_id, extension):
...
-        epoch = task.snapshots[-1][1]
-        snapshot_filename = task.snapshots[-1][0]
-    else:
-        for f, e in task.snapshots:
-            if e == epoch:
-                snapshot_filename = f
-                break
-    if not snapshot_filename:
+
+    try:
```
gheinrich: just wondering if the try/catch is really necessary here as it's done in the callee function too

digits/model/images/job.py
Lucaszw: Sorry missed this comment! Yes that is a much better idea, I changed it in my last commit.

digits/model/views.py
Lucaszw: Its not 100% necessary but I figured:

screenshot from 2016-07-08 18 10 50

looked better then:

screenshot from 2016-07-08 18 12 30

In this particular case

Lucaszw: Yeah I suppose you are right! It should be removed now.

digits/model/images/job.py
```diff
@@ -17,3 +20,16 @@ def __init__(self, **kwargs):
...
+    def json_dict(self, verbose=False, epoch=-1):
+        d = super(ImageModelJob, self).json_dict(verbose)
+        task = self.train_task()
+
+        d.update({
+            "job id": self.id(),
+            "creation time": self.status_history[0][1],
+            "username": self.username,
+        })
+
```
gheinrich: nice!

digits/model/tasks/torch_train.py
```diff
@@ -926,23 +928,29 @@ def get_network_desc(self):
...
+
+        if hasattr(self,"digits_version"):
+            stats.update({"digits version": self.digits_version})
 
-        return file_to_load
+        if hasattr(self.dataset,"resize_mode"):
+            stats.update({"image resize mode": self.dataset.resize_mode})
 
+        if hasattr(self.dataset,"labels_file"):
+            stats.update({"labels file": self.dataset.labels_file})
```
gheinrich: A very picky reviewer might argue that this method looks for the most part like the `CaffeTrainTask` one and there is an opportunity to factor out some code by moving the common bits up in the class hierarchy into `TrainTask`. It's totally fine this way though :-)

gheinrich:
Thanks, those changes look good to me.

digits/model/views.py
```diff
@@ -214,13 +214,18 @@ def download(job_id, extension):
         with tarfile.open(fileobj=b, mode='w:%s' % mode) as tf:
             for path, name in job.download_files(epoch):
                 tf.add(path, arcname=name)
+            tf.add(temp.name,arcname="stats.json")
```
lukeyeager: I vote for `info.json` instead.

digits/model/views.py
```diff
@@ -189,19 +192,16 @@ def download(job_id, extension):
     elif 'snapshot_epoch' in flask.request.form:
         epoch = float(flask.request.form['snapshot_epoch'])
 
-    task = job.train_task()
+    # Write the stats of the job to json,
+    # and store in tempfile (for archive)
+    stats = json.dumps(job.json_dict(verbose=False,epoch=epoch))
```
lukeyeager: Can we pretty-print here?
```py
# https://docs.python.org/2.7/library/json.html
json.dumps(foo, sort_keys=True, indent=4, separators=(',', ': '))
```

lukeyeager:
Is there any way to print dates as dates instead of floats?

digits/model/tasks/caffe_train.py
```diff
@@ -1031,6 +1045,38 @@ def after_runtime_error(self):
...
+        loc, mean_file = os.path.split(self.dataset.get_mean_file())
+
+        stats = {
+            "image dimensions": self.dataset.get_feature_dims(),
+            "mean file": mean_file,
+            "snapshot file": self.get_snapshot_filename(epoch),
+            "solver file": self.solver_file,
+            "train_val file": self.train_val_file,
+            "deploy file": self.deploy_file,
+            "framework": "caffe"
```
lukeyeager: Can we add "caffe_version" and "caffe_flavor" here? Or I guess "caffe version" and "caffe flavor" since you've gone with spaces.

digits/model/tasks/caffe_train.py
```diff
@@ -1031,6 +1045,38 @@ def after_runtime_error(self):
...
+        }
+
+        if hasattr(self,"original_file"):
+            stats.update({"original file": self.original_file})
```
lukeyeager: Internally in the DIGITS tree, let's keep referring to this file as "original". But since https://github.com/BVLC/caffe/pull/3863 has been merged, I'm hopeful that we'll be able to eventually provide all-in-one networks as outputs. So I vote for "network file" here, so that when we remove "train_val file" and "deploy file" later the name of this field doesn't have to change.

lukeyeager:
This looks good to me.

As with any changes to the REST API, this will probably break someone's workflow and we should make a note of it in the next release notes.

PR #895, opened 2016-07-11
Fix path to kitti devkit in data preparation script

gheinrich:
The script seems to assume that files from `devkit_object.zip` will be uncompressed into a folder `devkit` however that is not the case when using the archive from KITTI (?).


lukeyeager:
Woah, you're right. I was working with an old zipfile. Looks like they updated the README and when they recreated the zipfile they forgot the outer directory.

lukeyeager:
Fix verified. The data zipfiles don't seem to have changed, so it doesn't seem we need to update those.

PR #896, opened 2016-07-11
Upload Pre-trained Models for Fine Tuning.

Lucaszw:
Expects original model definitions from #891


IsaacYangSLA:
This will be useful if we later need to store several pre-trained models in/alongside DIGITS.

Lucaszw:
For sure!

digits/templates/partials/home/pretrained_model_pane.html
```diff
@@ -0,0 +1,126 @@
...
+                            <small ng-if="is_today(job.submitted)"
+                                   title="{[ job.submitted * 1000 | date:'medium' ]}">
+                                {[ job.submitted * 1000 | date:'h:mm a' ]}
+                            </small>
+                        </td>
+                    </tr>
+                </tbody>
+                <tbody ng-if="jobs.length == 0">
+                    <tr>
+                        <td colspan="{[fields.length]}">
```
jmancewicz: This colspan needs to include the displayed optional columns.

Like
```
                        {[ colspan = (storage.pretrained_model_fields | filter:show(true)).length + (storage.model_output_fields | filter:show(true)).length;'' ]}
                        <td colspan="{[colspan]}">
```
screen shot 2016-07-18 at 10 01 36 am

digits/templates/partials/home/pretrained_model_pane.html
```diff
@@ -0,0 +1,126 @@
...
+                            <span class="caret"></span>
+                        </button>
+                        <ul class="dropdown-menu" ng-click="$event.stopPropagation()">
+                            <li data-ng-repeat='field in storage.pretrained_model_fields'>
+                                <label style="font-weight: normal !important;">
+                                    <input type="checkbox" ng-model='field.show'>
+                                    {[ field.name ]}
+                                </label>
+                            </li>
+                            <li data-ng-repeat='field in storage.model_output_fields'>
```
jmancewicz: The files downloaded from the Download Model button include the mean.binaryproto, the prototxt files, caffemodel, and the newly added info.json. I don't see anything that would downloaded containing data in the model_output_fields. Perhaps that was not intended to there?

jmancewicz:
What is there a mechanism for adding the models to the pre-trained model list?

Lucaszw:
Thanks Joe!

I added an option in the New Model drop-down for uploading a pretrained model. There you select your weights, model def, and labels file. The data corresponds to the pretrained model job after uploading.

digits/templates/partials/home/pretrained_model_pane.html
Lucaszw: Good point! I will fix this in my next commit.

jmancewicz:
Would it make sense to allow it to point to a tgz file, view the table, and have that as an option for uploading models? There might be too much to assume about the role of files, but if it's a model from DIGITS, it should work. Just a thought.

Lucaszw:
That would definitely be a great option to have! I did think think about it, but decided against because of what you mentioned with external models, and also because it expects the original network prototxt file, and this wasn't saved into tar files prior to #891 .

lukeyeager:
> also because it expects the original network prototxt file, and this wasn't saved into tar files prior to #891

Can't we just check the `info.json` file for an "original" file and reject the upload if it doesn't exist?

Lucaszw:
I added the suggested changes with my latest commit. Everything now should be good to go! :)

digits/pretrained_model/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright (c) 2014-2016, NVIDIA CORPORATION.  All rights reserved.
```
lukeyeager: This should just be "2016" since it's a new file.

digits/static/js/components/PretrainedModel.js
```diff
@@ -0,0 +1,251 @@
+// d3 alert menu for uploading a pretrained model:
```
lukeyeager: Missing copyright header

digits/templates/models/images/classification/partials/new/network_tab_pretrained.html
```diff
@@ -0,0 +1,81 @@
+<script>
```
lukeyeager: Missing copyright header

lukeyeager:
The file upload doesn't look very good. @jmancewicz figured out how to make this work for all browsers: https://github.com/NVIDIA/DIGITS/pull/325

But I can't figure out how to add it to this PR because you're defining the form in javascript!?

lukeyeager:
* I don't think it makes sense to display columns like "accuracy" or "loss" for pretrained models on the home page, right?
* Is there any way to edit the job name after the fact?
* In a later PR, let's do some validation of the `.prototxt` before accepting it. It's annoying to get errors like `don't set the data_param.source` two or three pages after the one where you screwed up. But no need to fix it right away.

Lucaszw:
Thanks @lukeyeager ! I implemented your comments.

lukeyeager:
I tried uploading tarballs from 2 different models and I just get a red "Upload Failed" with no more information.

Lucaszw:
My bad! The problem should be fixed now. I added some better error messages for upload as well.

digits/model/tasks/caffe_train.py
```diff
@@ -84,7 +84,7 @@ def __init__(self, **kwargs):
         self.solver = None
 
         self.solver_file = CAFFE_SOLVER_FILE
-        self.original_file = CAFFE_ORIGINAL_FILE
+        self.model_file = CAFFE_ORIGINAL_FILE
```
lukeyeager: Can we use `network_file` instead of `model_file` here? I try to be consistent in calling the .prototxt file a "network description" and not calling it a model unless it has weights attached to it.

The nomenclature I'd like to migrate to (but don't fully support yet) is:

* "network" for a .prototxt file
* "model" or "trained model" for a .prototxt and a corresponding .caffemodel file
* "training" for a group of models

lukeyeager: Oh nevermind. We do the same for Torch. Rats.

However, you will want to be more careful with the upgrade path here.

* Bump the pickle version
* In `__setstate__`, upgrade from `original_file` to `model_file` appropriately

digits/pretrained_model/tasks/upload_pretrained.py
```diff
@@ -0,0 +1,92 @@
+# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
+from __future__ import absolute_import
+
+import base64
+from collections import OrderedDict
+import h5py
```
lukeyeager: For future PRs, put this import in a second group because it doesn't come with a standard Python installation.

Don't bother to fix it now, though.

digits/pretrained_model/tasks/upload_pretrained.py
```diff
@@ -0,0 +1,92 @@
...
+                        if resource.remaining() >= 1:
+                            self.gpu = int(resource.identifier)
+                            reserved_resources[gpu_key] = [(resource.identifier, 1)]
+                            break
+                return reserved_resources
+        return None
+
+    def move_file(self,input, output):
+        args  = ["mv", input, self.job_dir+"/"+output]
+        p = subprocess.Popen(args)
```
lukeyeager: This is not the way to move files. It won't work on Windows, for one thing.
http://stackoverflow.com/q/8858008

digits/pretrained_model/test_views.py
```diff
@@ -0,0 +1,116 @@
...
+        weights_file = open(snapshot_filename, 'r')
+        model_def_file = open(os.path.join(job.dir(),task.model_file), 'r')
+        labels_file = open(os.path.join(task.dataset.dir(),info["labels file"]), 'r')
+
+        rv = self.app.post(
+            '/pretrained_models/new',
+            data = {
+                'weights_file': weights_file,
+                'model_def_file': model_def_file,
+                'labels_file': labels_file,
```
lukeyeager: Weird. I thought you needed StringIO wrapping for this. Whatever.
https://github.com/NVIDIA/DIGITS/blob/v4.0.0/digits/model/images/classification/test_views.py#L353-L355

digits/pretrained_model/test_views.py
```diff
@@ -0,0 +1,116 @@
...
+        job = digits.webapp.scheduler.get_job(self.model_id)
+
+        if job is None:
+            raise AssertionError('Failed To Create Job')
+
+        info = json.dumps(job.json_dict(verbose=False,epoch=-1), sort_keys=True, indent=4, separators=(',', ': '))
+        info_io = io.BytesIO()
+        info_io.write(info)
+
+        tmp = tempfile.NamedTemporaryFile()
```
lukeyeager: This doesn't work on Windows, remember? https://github.com/NVIDIA/DIGITS/issues/919

Lucaszw: I avoided the two lines that seemed to break on windows this time around, and I validated the process through the python console on my windows laptop so I think it should work. Although maybe tomorrow I can try testing with @IsaacYangSLA to be sure.

Lucaszw:
Thanks Luke! I made the changes, and tried out the download feature across a different model jobs created at different times to test changing from original_file to model_file .

digits/pretrained_model/test_views.py
lukeyeager: Oh, I thought we just had to avoid using `NamedTemporaryFile` entirely.

lukeyeager:
This still isn't perfect, but it's working well for me and @IsaacYangSLA and we need it to build other functionality on top, so I'm going to merge it as-is.

TODOs:
* Don't ask for a labels file?
* Deal with rendering issue with upload button on Ubuntu 16.04 + Firefox 48
* Make sure uploaded models always show up on the home page without needing refresh (race condition?)

lukeyeager:
Thanks @Lucaszw!

PR #830, opened 2016-06-10
Image segmentation workflow

gheinrich:
I am planning to break this down into smaller PRs but I am interested in getting early feedback.
This PR builds upon #806 and adds:
- more flexibility in the choice of Torch labels (now we can use anything that fits in a `Datum` object)
- the ability to use a variable image size during inference (we need this to exploit the potential of FCNs)
- a walk-through and a simple example

@pansk @TimZaman : would you like to have a look?

I understand some of these changes might conflict with #777. I am willing to wait for #777 to be merged first then I will rebase.

PR #897, opened 2016-07-11
16 bit single channel support

IsaacYangSLA:
Prompt error when encoding is not None on 16 bit source images by checking encoding before saving to DB

Note this PR modifies files in another [PR](https://github.com/NVIDIA/DIGITS/pull/830).  Therefore, it's better to postpone reviewing this until #830 is merged.  After that, we can add 16 bit support on top of that new code base.


IsaacYangSLA:
Close this to reduce unrelated PR.

PR #898, opened 2016-07-11
Remove unnecessary print messages for LMDB mapsize

lukeyeager:
There's really no need for these


gheinrich:
That looks good to me.

PR #904, opened 2016-07-12
Add a hash to cached assets (.js and .css files)

jmancewicz:
When moving between versions of DIGITS changes in javascript content can cause render issues.

git checkout 74afccb
reload the DIGITS home page, click on a job name to navigate to a job page,
git checkout d0b407a
Use the DIGITS link in the upper left corner to navigate back to the home page.

The content of the app javascript file has changed, however the file is cached in session by the name, and the name has not changed. To flush the cache a hash for the static directory is being added to the url for the .js and .css files.

### DIFF
--- a/digits/templates/models/images/generic/show.html
+++ b/digits/templates/models/images/generic/show.html
@@ -5,7 +5,7 @@
 
 {% block job_content %}
 
-<script src="{{ url_for('static', filename='js/model-graphs.js') }}"></script>
+<script src="{{ url_for('static', filename='js/model-graphs.js', ver=dir_hash) }}"></script>
 
 {% set task = job.train_task() %}
 
--- a/digits/templates/error.html
+++ b/digits/templates/error.html
@@ -3,8 +3,8 @@
 <html>
 <head>
 <link rel="icon" href="{{url_for('static', filename='images/nvidia.ico')}}" />
-<link rel="stylesheet" href="{{ url_for('static', filename='css/bootstrap.min.css') }}">
-<link rel="stylesheet" href="{{ url_for('static', filename='css/bootstrap-theme.min.css') }}">
+<link rel="stylesheet" href="{{ url_for('static', filename='css/bootstrap.min.css', ver=dir_hash) }}">
+<link rel="stylesheet" href="{{ url_for('static', filename='css/bootstrap-theme.min.css', ver=dir_hash) }}">
 </head>
 
 <body>
--- a/digits/templates/home.html
+++ b/digits/templates/home.html
@@ -5,12 +5,12 @@
 {% extends "layout.html" %}
 
 {% block head %}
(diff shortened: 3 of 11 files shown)


lukeyeager:
*Requested at https://github.com/NVIDIA/DIGITS/pull/734#issuecomment-231368397*

digits/templates/models/images/generic/show.html
```diff
@@ -1,11 +1,12 @@
+
```
lukeyeager: spurious whitespace change

digits/utils/__init__.py
```diff
@@ -10,6 +10,8 @@
 import platform
 from random import uniform
 from urlparse import urlparse
+import re
+import hashlib
```
lukeyeager: Can we adhere to PEP8 here (https://github.com/NVIDIA/DIGITS/pull/501)?

digits/utils/__init__.py
```diff
@@ -152,6 +154,33 @@ def parse_version(*args):
     except AttributeError:
         return pkg_resources.parse_version(v)
 
+def dir_hash(dir_name):
```
lukeyeager: Maybe this belongs in `digits.utils.filesystem` instead?

digits/utils/__init__.py
```diff
@@ -152,6 +154,33 @@ def parse_version(*args):
...
+
+    def reduce_hash(hashlist):
+        md5 = hashlib.md5()
+        for hashvalue in sorted(hashlist):
+            md5.update(hashvalue.encode('utf-8'))
+        return md5.hexdigest()
+
+    if not os.path.isdir(dir_name):
+        raise TypeError('{} is not a directory.'.format(dir_name))
+    hashvalues = []
```
lukeyeager: Seems like you could simplify a bit by storing a single hash and updating it as you go instead of storing an array of hashes? Any reason not to?

digits/utils/__init__.py
```diff
@@ -152,6 +154,33 @@ def parse_version(*args):
...
+    for root, dirs, files in os.walk(dir_name, topdown=True):
+        if not re.search(r'/\.', root):
```
lukeyeager: Is this supposed to exclude hidden directories? What if the dirname is `.git` instead of `foo/.git`? A comment might be appropriate here, either way.

digits/utils/__init__.py
jmancewicz: Currently, the list is sorted, requiring the list. Trying it without it being `sorted`.

jmancewicz:
I think I hit all your points and simplified the dir_hash function a bit.

lukeyeager:
Code looks good. I verified the fix. I'll merge once TravisCI passes (not because I think it won't pass, but because I like to see check marks by merged PRs :smiley:).

PR #911, opened 2016-07-15
Stop setting cache timeout in TravisCI

lukeyeager:
It refers to the caching command timeout (https://github.com/travis-ci/travis-ci/issues/6317#issuecomment-233065994), not how long before the caches expire as I had thought.


gheinrich:
That looks good to me. I had also misunderstood the purpose of this `timeout` command.

PR #921, opened 2016-07-21
[DONT MERGE] Plug-in examples

gheinrich:
Simple example to show how to create a plug-in. The plug-in registers an entry point that any Python module can retrieve through the `digits.view` key (see how this is done in `digits/extensions/view/__init__.py`).

To install the plug-in:
```
pip install $DIGITS_HOME/plugins/dummyView
```

Since the plug-in requires DIGITS modules and use them through absolute imports, DIGITS must also be installed (once):
```
pip install $DIGITS_HOME
```

### DIFF
--- a/digits/extensions/view/__init__.py
+++ b/digits/extensions/view/__init__.py
@@ -1,18 +1,24 @@
 # Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
 from __future__ import absolute_import
 
+import copy
+from pkg_resources import iter_entry_points
+
 from . import boundingBox
 from . import imageGradients
 from . import rawData
 
-view_extensions = [
-    # Set show=True if extension should be shown by default
-    # in the 'Select Visualization Method' dialog. These defaults
-    # can be changed by editing DIGITS config option
-    # 'view_extension_list'
-    {'class': boundingBox.Visualization, 'show': True},
-    {'class': imageGradients.Visualization, 'show': False},
-    {'class': rawData.Visualization, 'show': True},
+
+# Entry point group (this is the key we use to register and
+# find installed plug-ins)
+GROUP = "digits.plugins.view"
+
+
+# built-in extensions
+builtin_view_extensions = [
+    boundingBox.Visualization,
+    imageGradients.Visualization,
+    rawData.Visualization,
 ]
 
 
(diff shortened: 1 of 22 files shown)


digits/extensions/view/__init__.py
```diff
@@ -15,6 +17,13 @@
     {'class': rawData.Visualization, 'show': True},
 ]
 
+# find extensions
```
gheinrich: this is where we find all registered extensions (using `group='digits.view'`)

plugins/view/dummyView/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
```
gheinrich: this is just mostly a copy of the `raw data` view extension

plugins/view/dummyView/setup.py
```diff
@@ -0,0 +1,15 @@
...
+from setuptools import setup
+
+setup(
+    name = "digits_dummy_view_plugin",
+    version = "0.0.1",
+    author = "Greg Heinrich",
+    author_email = "gheinrich@gmail.com",
+    description = ("Simple example to show how to write a vizualisation plugin"),
+    license = "Apache",
+    entry_points = {
```
gheinrich: this is where we register the `entry_point` for our plug-in

plugins/view/dummyView/__init__.py
gheinrich: note that this directory can be located anywhere - it doesn't have to be in DIGITS tree

setup.py
```diff
@@ -0,0 +1,9 @@
+from setuptools import setup, find_packages
+
+import digits
```
gheinrich: this is the script to install DIGITS through `pip`. We need to do this as plug-ins reference some module within DIGITS and they need to do so using absolute imports

plugins/view/dummyView/view_template.html
```diff
@@ -0,0 +1,10 @@
+{# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved. #}
```
gheinrich: I think I will need to create a `MANIFEST.in` file to include those non-Python files

gheinrich:
Updated with data ingestion plug-in example

gheinrich:
Close in favour of #947

PR #923, opened 2016-07-21
Fix IOError in Windows when downloading model

### LINKED ISSUE #919: [proposed question/bug] Can not download model in DIGITS on Windows
As title says,

When using Chrome, it reports:

IOError
[Errno 13] Permission denied: 'c:\\users\\jzou\\appdata\\local\\temp\\1\\tmpevv0_b'

When using IE, it simply shows "HTTP 500 Internal Server Error" page.

Any ideas? Thanks.

The DIGITS info is copied below:

DIGITS version: 
 4.1-dev
Caffe version: 
 1.0.0-rc3
Caffe flavor: 
 BVLC

IsaacYangSLA:
Fix #919 
In Windows, tempfile behaves differently than in Linux.
Use BytesIO instead of tempfile since info.json is usually small.
Tested in Windows with default compression (tar.gz) and with curl -XPOST download.zip.  Both worked without issues.

### DIFF
--- a/digits/model/views.py
+++ b/digits/model/views.py
@@ -195,9 +195,8 @@ def download(job_id, extension):
     # Write the stats of the job to json,
     # and store in tempfile (for archive)
     info = json.dumps(job.json_dict(verbose=False,epoch=epoch), sort_keys=True, indent=4, separators=(',', ': '))
-    temp = tempfile.NamedTemporaryFile()
-    temp.write(info)
-    temp.seek(0)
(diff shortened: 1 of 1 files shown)


lukeyeager:
I haven't tested it, but this definitely looks like the right solution to me. @Lucaszw can you review and merge?

Lucaszw:
Tested and everything seems to be working fine! Thanks @IsaacYangSLA !

Issue #907, opened 2016-07-15
Classify_one strange failure 


engharat:
Hi guys,
I changed from nvcaffe 015 to nvcaffe016, cuda 7.5 to cuda 8.0, and drivers from 354 to 362 (the one linked with cuda 8.0), cudnn 4 to 5.
I got a bunch of problems, so I reverted back to nvcaffe 015 (15.7) cuda 7.5, still keeping the 362 driver and the cudnn 5.
The problem is that I get this strange error when doing classify_one with DIGITS:
`Error code -6
F0714 21:28:53.351652 15904 gpu_memory.cpp:68] Check failed: initialized_ Create GPUMemory::Scope to initialize Memory Manager`
the problem happens when I run the classify_one on a training network. It seems like it cannot allocate memory if still used by the training (but plenty of GPU RAM available)
any idea?

lukeyeager:
Submitted to NVIDIA/caffe repo: https://github.com/NVIDIA/caffe/issues/196

PR #938, opened 2016-07-28
Test for cpu inference

lukeyeager:
Add a test to detect this bug: https://github.com/NVIDIA/caffe/issues/196 (was https://github.com/NVIDIA/DIGITS/issues/907)

Verifies this fix: https://github.com/NVIDIA/caffe/pull/200

### DIFF
--- a/digits/task.py
+++ b/digits/task.py
@@ -203,7 +203,9 @@ class Task(StatusCls):
         #https://docs.python.org/2/library/subprocess.html#converting-argument-sequence
         if platform.system() == 'Windows':
             args = ' '.join(args)
-        self.logger.info('Task subprocess args: {}'.format(args))
+            self.logger.info('Task subprocess args: "{}"'.format(args))
+        else:
+            self.logger.info('Task subprocess args: "%s"' % ' '.join(args))
 
         p = subprocess.Popen(args,
                 stdout=subprocess.PIPE,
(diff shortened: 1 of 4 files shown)


lukeyeager:
@gheinrich can you test this with Torch? It isn't working for me but I don't understand the errors.

digits/task.py
```diff
@@ -203,7 +203,9 @@ def run(self, resources):
         #https://docs.python.org/2/library/subprocess.html#converting-argument-sequence
         if platform.system() == 'Windows':
             args = ' '.join(args)
-        self.logger.info('Task subprocess args: {}'.format(args))
+            self.logger.info('Task subprocess args: "{}"'.format(args))
+        else:
+            self.logger.info('Task subprocess args: "%s"' % ' '.join(args))
```
gheinrich: Thanks I always make this change locally to be able to copy/paste command lines into a command prompt! Otherwise you get the list notation (with brackets, commas et al.).

gheinrich:
Hi Luke, I added a couple of commits on top of yours on https://github.com/gheinrich/DIGITS/tree/dev/torch-cpu-inference-test. This addresses the case where you train a model on GPU then perform inference on CPU **under some condition**. The condition being that you didn't use CuDNN layers during training as it is otherwise pretty difficult to convert the model back to non-CuDNN. I'd like to skip the test in this case (as I did for LeNet test cases). We can revisit this if we determine later on that we desperately need to do this.

lukeyeager:
Added your changes, thanks!

And thanks for the info. I didn't realize we had those sorts of problems with Torch. I believe we have similar issues for Caffe with BatchNorm, too.

Issue #845, opened 2016-06-15
Multiple CUDA contexts with digits-server


lukeyeager:
To reproduce:
1. Turn off device_query caching
   - Change the default value for `force_reload` to `True` [here](https://github.com/NVIDIA/DIGITS/blob/v4.0.0-rc.1/digits/device_query.py#L159)
2. Run `digits-server`
   - After https://github.com/NVIDIA/DIGITS/pull/844, you should see this error in the server output:
     
     ```
     ...
     cudaGetDeviceProperties() failed with error #3
     ...
     return get_devices()[int(device_id)]
     IndexError: list index out of range
     ...
     ```

gunicorn does a fork of the DIGITS server, and the server creates a CUDA context in the parent before forking. Then, when the child tries to call `cudaGetDevicesProperties()`, it fails with an initialization error. We need to make sure we don't make any CUDA calls in the DIGITS server before forking when running with gunicorn, if possible.

lukeyeager:
No longer an issue because `gunicorn` was removed with https://github.com/NVIDIA/DIGITS/pull/1127.

PR #944, opened 2016-07-30
Avoid creation of CUDA context in master process operated by gunicorn

dmikushin:
```
$ cat gunicorn/arbiter.py | grep "pid = os.fork()"
        pid = os.fork()
```

Otherwise, forked worker processes will receive invalid CUDA context and crash on image testing:

```
/usr/bin/gunicorn --config gunicorn_config.py digits.webapp:app
2016-07-30 21:26:50 [5394] [INFO] Starting gunicorn 17.5
2016-07-30 21:26:50 [5394] [DEBUG] Arbiter booted
2016-07-30 21:26:50 [5394] [INFO] Listening at: http://0.0.0.0:34448 (5394)
2016-07-30 21:26:50 [5394] [INFO] Using worker: socketio.sgunicorn.GeventSocketIOWorker
2016-07-30 21:26:50 [5456] [INFO] Booting worker with pid: 5456
WARNING: Logging before InitGoogleLogging() is written to STDERR
E0730 21:27:11.797976  5456 common.cpp:110] Cannot create Cublas handle. Cublas won't be available.
E0730 21:27:11.800433  5456 common.cpp:117] Cannot create Curand generator. Curand won't be available.
E0730 21:27:11.802403  5456 common.cpp:121] Cannot create cuDNN handle. cuDNN won't be available.
F0730 21:27:11.804111  5456 syncedmem.hpp:19] Check failed: error == cudaSuccess (3 vs. 0)  initialization error
```

This patch performs necessary device querying in forked process as well, keeping master of CUDA context creation.

### DIFF
--- a/digits/device_query.py
+++ b/digits/device_query.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import
 import argparse
 import ctypes
 import platform
+from multiprocessing import Process, Pipe
 
 class c_cudaDeviceProp(ctypes.Structure):
     """
@@ -160,18 +161,7 @@ def get_nvml():
 
 devices = None
 
-def get_devices(force_reload=False):
-    """
-    Returns a list of c_cudaDeviceProp's
-    Prints an error and returns None if something goes wrong
-
-    Keyword arguments:
-    force_reload -- if False, return the previously loaded list of devices
-    """
-    global devices
-    if not force_reload and devices is not None:
-        # Only query CUDA once
-        return devices
+def get_devices_process(force_reload, conn):
     devices = []
 
     cudart = get_cudart()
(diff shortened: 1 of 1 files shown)


lukeyeager:
Have you tried actually using this code? I'm pretty sure you can't `import multiprocessing` when using gevent.

This bug has already been reported here: https://github.com/NVIDIA/DIGITS/issues/845

dmikushin:
Yes, this patch is tested and working.

lukeyeager:
The TravisCI build fails as I expected - everything just hangs.

dmikushin:
The patch is tested and working for digits v3.0

lukeyeager:
Weird, you're right - this patch works on my machine. I wonder why it causes TravisCI to freeze? I'm restarting the build just in case.

lukeyeager:
I can't reproduce your bug, though. Running `digits-server` and testing an image works fine for me. I've never actually seen a problem without editing the source code as I explained here: https://github.com/NVIDIA/DIGITS/issues/845#issue-160541961

dmikushin:
Thank you for looking into this. We have this bug on two different laptops
with a freshly installed digits v3.0 using GitHub wiki guide, Kepler and
Maxwell GPUs, GPU driver as of CUDA 8.0 RC.

lukeyeager:
Ah yes, we inadvertently fixed this with https://github.com/NVIDIA/DIGITS/pull/573 in-between v3.0 and v4.0: https://github.com/NVIDIA/nvidia-docker/issues/108#issuecomment-225327217.

v4.0 packages should be out very soon and your problem will be solved. Thanks for this PR though - it's a suggestion for how to solve #845. I'd prefer to find a solution that simply didn't create a CUDA context in the server pre-fork.

PR #927, opened 2016-07-25
Add support for plug-ins

gheinrich:
Support for data and view plug-ins. See #921 for examples.

Extra command to support plug-ins:
```
pip install -e $DIGITS_HOME
```

PR #947, opened 2016-08-02
Move image gradient extensions to plug-in area

gheinrich:
This includes several changes:
- move the data extension to the plug-in area (and wrap it into an installable package)
- move the view extension to the plug-in area (and wrap it into an installable package)
- remove config options for extensions
- update Travis script to install plug-ins
- update regression tutorial with commands to install plug-ins

This depends on PR #927


.travis.yml
```diff
@@ -48,6 +48,10 @@ install:
     - pip install -r requirements.txt
     - pip install -r requirements_test.txt
 
+    # image gradients plug-ins
+    - pip install plugins/data/imageGradients
+    - pip install plugins/view/imageGradients
```
lukeyeager: So... you don't actually have to `pip install` DIGITS? You only have to install the plugins?

lukeyeager: No wait, yes you do.
```
      File "<string>", line 1, in <module>
      File "/tmp/pip-59l1vv-build/setup.py", line 5, in <module>
        from digits.extensions.data import GROUP as DIGITS_PLUGIN_GROUP
    ImportError: No module named digits.extensions.data
```
I think this Travis build is about to fail, let's see ...
https://travis-ci.org/NVIDIA/DIGITS/builds/162241842

lukeyeager:
This was the only extension that was hidden by default, right? Can we just get rid of the `DIGITS_ALL_EXTENSIONS` envvar then?
https://github.com/NVIDIA/DIGITS/blob/f2b0b0a640/digits/config/extension_list.py

It seems to me that we can entirely replace the show/hide+envvar system with the uninstalled/installed system.

gheinrich:
> we can entirely replace the show/hide+envvar system with the uninstalled/installed system

Sure, this was my original intention but the change disappeared during the rebase. I have re-instated that change.

> you don't actually have to pip install DIGITS?

You're right - I have updated the change, thanks.

examples/regression/README.md
```diff
@@ -19,51 +20,33 @@ Image regression models may learn to predict any number of image characteristics
...
-
-Input the IDs of the extensions you would like to use, separated by commas.
-
-  Suggested values:
-  (D*) [default] image-bounding-boxes,all-raw-data
-  (A)  [all]     image-bounding-boxes,image-gradients,all-raw-data
-  (N)  [none]    <NONE>
->> A
-Using "image-bounding-boxes,image-gradients,all-raw-data"
+$ pip install -e $DIGITS_HOME
```
lukeyeager: Do you mind if we use `_ROOT` instead of `_HOME` here? See https://github.com/NVIDIA/DIGITS/commit/800e352b8525f12e92e92d472461f8d104477fda.

plugins/data/imageGradients/setup.py
```diff
@@ -0,0 +1,24 @@
+import os
+from setuptools import setup, find_packages
+
+# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
```
lukeyeager: Any reason why the copyright isn't at the top of this file?

examples/regression/README.md
gheinrich: Not at all, updated in latest commit

plugins/data/imageGradients/setup.py
gheinrich: Good catch, thanks!

gheinrich: Updated in latest commit

PR #950, opened 2016-08-03
Generic datasets with variable-shape entries

gheinrich:


lukeyeager:
This is for FCNs, right? Looks fine to me - thanks for the tests!

gheinrich:
Yes this is for even more flexibility on the use of FCNs. Thanks for the review!

PR #956, opened 2016-08-04
adds rms_decay option when RMSProp solver is selected

smichalowski:
When RMSProp solver is selected, rms_decay options appears so user can change its default value:

zrzut ekranu 2016-08-05 o 00 35 30

### DIFF
--- a/digits/model/forms.py
+++ b/digits/model/forms.py
@@ -162,7 +162,7 @@ class ModelForm(Form):
     ### Additional settings specific to selected solver
 
     rms_decay = utils.forms.FloatField('RMS decay value',
-            default = 0.02,
+            default = 0.99,
             validators = [
                 validators.NumberRange(min=0),
                 ],
--- a/digits/model/forms.py
+++ b/digits/model/forms.py
@@ -159,6 +159,16 @@ class ModelForm(Form):
                 raise validators.ValidationError(
                     'Solver type not supported by this framework')
 
+    ### Additional settings specific to selected solver
+
+    rms_decay = utils.forms.FloatField('RMS decay value',
+            default = 0.02,
+            validators = [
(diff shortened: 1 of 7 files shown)


lukeyeager:
This is excellent code. One ignorant question: why default to 0.02? Is that a typical value for this option? Searching through the Caffe tree, I'm seeing values like 0.98 and 0.95.

Also, I'll need you to [send me a CLA](https://github.com/NVIDIA/DIGITS/blob/master/.github/CONTRIBUTING.md#pull-requests) before I can accept this.

smichalowski:
I took default value from [caffe solver manual](http://caffe.berkeleyvision.org/tutorial/solver.html). It says:

> If the gradient updates results in oscillations the gradient is reduced by times 1−δ. Otherwise it will be increased by δ. The default value of δ (rms_decay) is set to δ=0.02

I'll send CLA tomorrow.

alicanb:
Caffe solver manual is wrong. You need write (1-δ) as rms_decay. see [https://github.com/BVLC/caffe/pull/4379](this PR)

digits/model/forms.py
```diff
@@ -159,6 +159,16 @@ def validate_solver_type(form, field):
                 raise validators.ValidationError(
                     'Solver type not supported by this framework')
 
+    ### Additional settings specific to selected solver
+
+    rms_decay = utils.forms.FloatField('RMS decay value',
+            default = 0.02,
```
lukeyeager: This should be 0.99 to match https://github.com/BVLC/caffe/pull/4379, right?

lukeyeager:
Thanks for the update - looks good. I'll merge once I receive your CLA.

smichalowski:
just sent it.

lukeyeager:
Thanks!

PR #964, opened 2016-08-09
REST API walk-through

gheinrich:


examples/REST-API/README.md
```diff
@@ -0,0 +1,391 @@
...
+If DIGITS is running on another host/port, please adjust the command lines accordingly.
+
+## Image classification
+
+This is another way of running the [Getting started](../../docs/GettingStarted.md) tutorial.
+If you haven't gone through this tutorial, please do so now.
+
+### Creating the classification dataset
+
+In order to create a dataset, you will first need to log in.
```
lukeyeager: It's probably worth noting that you don't need to do this with a cookie. You can just add `-F username=luke` to the end of any `curl` command.

https://github.com/NVIDIA/DIGITS/blob/v4.0.0/digits/utils/auth.py#L11-L13

gheinrich: Thanks, I have added a note about that.

lukeyeager:
Do you think this file might make more sense  under `/docs/` rather than under `/examples/`?

Either way, we should add a link to it at the end of `/docs/GettingStarted.md`.

gheinrich:
I have moved the file to `/docs` and added the link at the end of `/docs/GettingStarted.md`.

lukeyeager:
Thanks! LGTM

PR #976, opened 2016-08-15
Simplify input scaling on LeNet network

mpbrigham:
Removed scaling from data layers and kept single unified power layer. Previous definition was faster due to multi-threaded data loader, but potentially less straightforward to new users.

The scaling factor 0.0125 corresponds to 1/(standard deviation on MNIST dataset), which is ~80 per pixel (from a range of [0-255] per pixel). Scaling factor comment updated to reflect this.

### DIFF
--- a/digits/standard-networks/caffe/lenet.prototxt
+++ b/digits/standard-networks/caffe/lenet.prototxt
@@ -3,12 +3,8 @@ name: "LeNet"
 layer {
   name: "train-data"
   type: "Data"
-  top: "scaled"
+  top: "data"
   top: "label"
-  transform_param {
-    # 1/(standard deviation)
-    scale: 0.0125
-  }
   data_param {
     batch_size: 64
   }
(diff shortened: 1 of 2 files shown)


lukeyeager:
@gheinrich any issue with this? The reasons for a separate power layer aren't particularly compelling (see https://github.com/NVIDIA/DIGITS/pull/733).

TimZaman:
@mpbrigham maybe also reflect this in the torch's implementation by removing https://github.com/NVIDIA/DIGITS/blob/master/digits/standard-networks/torch/lenet.lua#L36. I recall the learning rate should also be updated to yield identical results, not sure.

mpbrigham:
@TimZaman I've synced the scaling factor in torch's implementation (commit `2c40306`). The previous value rescaled the range from [0, 255] to [0, 1] by multiplying by 1/256.

gheinrich:
That looks OK to me thanks.

lukeyeager:
Thanks for the improvement @mpbrigham!

PR #978, opened 2016-08-15
move had_labes icon from python to an angularjs directive

jmancewicz:
Trying to keep the drawing decision further down the view path


lukeyeager:
@Lucaszw can you review please?

Lucaszw:
Good call! Everything looks good :)

PR #968, opened 2016-08-10
Gradient ascent

Lucaszw:
Do Not Merge! 
Gradient Ascent for Caffe (Torch support not completed yet)
Depends on #937

Optimization Technique From Deep Vis Toolbox
```
@inproceedings{yosinski-2015-ICML-DL-understanding-neural-networks,
Author = {Jason Yosinski and Jeff Clune and Anh Nguyen and Thomas Fuchs and Hod Lipson},
Booktitle = {Deep Learning Workshop, International Conference on Machine Learning (ICML)},
Title = {Understanding Neural Networks Through Deep Visualization},
Year = {2015}}
```

PR #937, opened 2016-07-28
Layer Visualization And Weights for Pretrained Jobs

Lucaszw:
Depends on #896 and #932 
Layer Visualizations for Torch and Caffe Pretrained Models , in addition to Visualization of Weights.

PR #984, opened 2016-08-17
convert generic tasks to pretrained models

### LINKED ISSUE #972: Convert non-classification snapshots to "pretrained models"
Looks like https://github.com/NVIDIA/DIGITS/pull/932 doesn't let me convert other types of models like those for object detection or semantic segmentation.

Lucaszw:
Close #972

### DIFF
--- a/digits/model/views.py
+++ b/digits/model/views.py
@@ -5,6 +5,7 @@ from datetime import timedelta
 import io
 import json
 import math
+import os
 import tarfile
 import tempfile
 import zipfile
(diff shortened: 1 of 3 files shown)


lukeyeager:
I'm hitting this again (also here: https://github.com/NVIDIA/DIGITS/pull/937#issuecomment-239536218):
```
  File "/home/lyeager/digits/digits/model/views.py", line 216, in to_pretrained
    job.dir() + "/" + task.model_file ,
TypeError: cannot concatenate 'str' and 'NoneType' objects
```
Is this fixed in #937? Which do you want me to merge first?

Lucaszw:
Hi Luke, yes it probably would be best to merge #937 . Its also includes a fix so that all models have model_file vs just new ones.

digits/model/views.py
```diff
@@ -202,13 +202,22 @@ def to_pretrained(job_id):
     snapshot_filename = None
     snapshot_filename = task.get_snapshot(epoch)
 
+    # Set defaults:
+    labels_path = None
+    resize_mode = None
+
+    if "labels file" in info:
+        labels_path = task.dataset.dir() + "/" + info["labels file"]
```
gheinrich: as @TimZaman commented `os.path.join` is preferred.

gheinrich:
Isn't it easier to just remove the notions of `labels_path`, `resize_mode`, etc. from pre-trained jobs? I don't see where they are used. Besides, those would be properties of the dataset, not the model. Admittedly they are part of the model snapshot because they're needed for inference. However if we can't perform inference on a pre-trained model I think those fields are just getting in the way of simplicity.

Lucaszw:
@gheinrich The labels_path is used for writing the deploy file for caffe (although only for getting the number of outputs, so there are definitely better ways to do that). Although for now resize_mode isn't used (I kept it in case we decided to add test images for inference on pre-trained models). Although it might be best left for another PR as its currently expected for classification jobs as well.

gheinrich:
Ah and the deploy file is just used when visualizing layers, correct?

Lucaszw:
@gheinrich Yeah for visualizing, and for gradient ascent (#968)

PR #987, opened 2016-08-20
Add About Menu

jmancewicz:
Adding an "About" menu to the layout. This is an initial implementation of adding relevant links

https://developer.nvidia.com/digits
https://groups.google.com/forum/#!forum/digits-users
https://github.com/NVIDIA/DIGITS
mailto:digits@nvidia.com

screen shot 2016-08-19 at 8 19 32 pm


gheinrich:
That looks OK to me. Can we add a link to the docs too?
Question: why is the Nvidia logo blue (not green)?

digits/templates/layout.html
```diff
@@ -53,9 +53,21 @@
...
+                <li><span class="navbar-text">Caffe flavor: <br> {{caffe_flavor}}</span></li>
             </ul>
+            </li>
+            <li class="dropdown">
+                <a href="#" class="dropdown-toggle" data-toggle="dropdown">About<span class="caret"></span></a>
+                <ul class="dropdown-menu navbar-inverse">
+                    <li><span class="navbar-text brightness"><a href="https://developer.nvidia.com/digits">
+                        <img src="{{url_for('static', filename='images/nvidia-logo.png')}}"></img>
+                    </a></span></li>
+                    <li><span class="navbar-text"><a href="https://groups.google.com/forum/#!forum/digits-users">DIGITS User Group</a></span></li>
```
lukeyeager: Can we add `target="_blank"` so they open in a new tab? Opening a new tab when going to a different website feels natural to me.

jmancewicz:
@gheinrich, I obtained that image from developer.nvidia.com, and it was white. Seeing the image based link in white and the other links as blue made it feel to me as if the white 'nvidia develpoper' image was not a link. Changing it to solid blue then made it feel like a link again.

I considered these two other options as well:

screen shot 2016-08-22 at 10 39 14 am

screen shot 2016-08-22 at 10 41 03 am

I'm leaning toward the first.

jmancewicz:
@gheinrich, are you thinking of
 * Getting Started (https://github.com/NVIDIA/DIGITS/blob/master/docs/GettingStarted.md)
 * The docs (https://github.com/NVIDIA/DIGITS/blob/master/docs/)
 * or something else?

gheinrich:
Yes, the first is more consistent with https://developer.nvidia.com/, which I'm sure a lot of marketing people have spent a long time thinking through :-)

Regarding the link to **the** docs I am unsure. I'd say the `docs` folder (?). Perhaps it's time to add a `README.md` in this folder so it shows on Github by default and in there we'd be able to give a summary of what the various `.md` files are meant to say?

jmancewicz:
screen shot 2016-08-22 at 7 55 53 pm

PR #988, opened 2016-08-20
Make the Browse... buttons more bootstrap styled

jmancewicz:
Changing the Browse buttons in the file inputs for Image Testing from

screen shot 2016-08-19 at 9 16 34 pm

to

screen shot 2016-08-19 at 9 14 25 pm

I might try to make this into a template or an angularjs directive.


lukeyeager:
Thanks!

PR #961, opened 2016-08-08
Semantic segmentation work-flow

gheinrich:

PR #993, opened 2016-08-23
 	Allow segmentation color map to be specified explicitly

gheinrich:
Add form fields to allow users to provide separate text file with color/class mappings.
This is useful when working with datasets that have label images in RGB.

Depends on #961

### DIFF
--- a/digits/extensions/data/imageSegmentation/data.py
+++ b/digits/extensions/data/imageSegmentation/data.py
@@ -7,6 +7,7 @@ import random
 
 import numpy as np
 import PIL.Image
+import PIL.ImagePalette
 
 from digits.utils import image, subclass, override, constants
 from digits.utils.constants import COLOR_PALETTE_ATTRIBUTE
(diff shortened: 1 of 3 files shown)


lukeyeager:
I'm not sure how to test this - is there a public dataset that uses RGB labels like this?

When it's convenient, can you rebase this now that #961 is merged (just for clarify / ease of review)?

gheinrich:
Thanks. This change is now rebased on the tip of the master branch. An example of a segmentation dataset with RGB labels is [SYNTHIA](http://synthia-dataset.net/). The color/class mappings for this dataset are given in the dataset's `README.txt` and may be provided to DIGITS through a text file that contains:

```
0   0   0
128 128 128
128 0   0
128 64  128
0   0   192
64  64  128
128 128 0
192 192 128
64  0   128
192 128 128
64  64  0
0   128 192
```
Ideally you should update the FCN prototxt to reduce the number of outputs to 12 (since there are 12 classes in this dataset) though I found it that it wasn't necessary to get the network to converge.

lukeyeager:
LGTM

PR #997, opened 2016-08-25
cleaned up layerVis and gradientAscent to just include job updates

Lucaszw:
Includes all changes made outside of visualizations in LayerVis and Gradient Ascent pull requests

- Moved cleanedUpGenericNetwork to helpers file
- Change image dimensions from width -> height, to height -> width in pretrained model job
- Used constants for model definition, deploy.prototxt etc.
- Cleaned up import statements
- Added missing headers
- Fixed errors when converting old models
- Added hash_dir to PretrainedModel javascript component
- Fixed tab spacing in PretrainedModel
- Removed "path" + "/" + "file" with os.path.join
- Changed has_labels to method has_labels_file()

### DIFF
--- a/digits/framework_helpers/caffe_helpers.py
+++ b/digits/framework_helpers/caffe_helpers.py
@@ -1,0 +1,269 @@
+# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
+import caffe_pb2
+from google.protobuf import text_format
+
+# Constants
+CAFFE_SOLVER_FILE = 'solver.prototxt'
+CAFFE_ORIGINAL_FILE = 'original.prototxt'
+CAFFE_TRAIN_VAL_FILE = 'train_val.prototxt'
+CAFFE_SNAPSHOT_PREFIX = 'snapshot'
+CAFFE_DEPLOY_FILE = 'deploy.prototxt'
+CAFFE_WEIGHTS_FILE = 'model.caffemodel'
+
+class Error(Exception):
+    pass
+
+class CaffeSanityCheckError(Error):
+    """A sanity check failed"""
+    pass
+
+def filterLayersByState(network):
+    """
+    Splits up a network into data, train_val and deploy layers
+    """
+    # The net has a NetState when in use
+    train_state = caffe_pb2.NetState()
+    text_format.Merge('phase: TRAIN stage: "train"', train_state)
+    val_state = caffe_pb2.NetState()
+    text_format.Merge('phase: TEST stage: "val"', val_state)
+    deploy_state = caffe_pb2.NetState()
+    text_format.Merge('phase: TEST stage: "deploy"', deploy_state)
+
+    # Each layer can have several NetStateRules
+    train_rule = caffe_pb2.NetStateRule()
+    text_format.Merge('phase: TRAIN', train_rule)
+    val_rule = caffe_pb2.NetStateRule()
+    text_format.Merge('phase: TEST', val_rule)
+
+    # Return three NetParameters
+    data_layers = caffe_pb2.NetParameter()
+    train_val_layers = caffe_pb2.NetParameter()
+    deploy_layers = caffe_pb2.NetParameter()
+
+    for layer in network.layer:
+        included_train = _layerIncludedInState(layer, train_state)
+        included_val = _layerIncludedInState(layer, val_state)
+        included_deploy = _layerIncludedInState(layer, deploy_state)
+
+        # Treat data layers differently (more processing done later)
+        if 'Data' in layer.type:
+            data_layers.layer.add().CopyFrom(layer)
+            rule = None
+            if not included_train:
+                # Exclude from train
+                rule = val_rule
+            elif not included_val:
+                # Exclude from val
+                rule = train_rule
+            _setLayerRule(data_layers.layer[-1], rule)
+
+        # Non-data layers
+        else:
+            if included_train or included_val:
+                # Add to train_val
+                train_val_layers.layer.add().CopyFrom(layer)
+                rule = None
+                if not included_train:
+                    # Exclude from train
+                    rule = val_rule
+                elif not included_val:
+                    # Exclude from val
+                    rule = train_rule
+                _setLayerRule(train_val_layers.layer[-1], rule)
+
+            if included_deploy:
+                # Add to deploy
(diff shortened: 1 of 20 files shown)


digits/pretrained_model/tasks/torch_upload.py
```diff
@@ -20,14 +29,21 @@ def get_model_def_path(self):
         """
         Get path to model definition
         """
-        return self.job_dir+"/original.lua"
+        return self.get_deploy_path();
```
lukeyeager: Is that a semicolon I see?

Lucaszw: @lukeyeager My bad! Should be removed now

digits/templates/partials/home/upload_pretrained_model.html
```diff
@@ -1,7 +1,7 @@
 {# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved. #}
 <script type="text/javascript" src="{{ url_for('static', filename='js/components/PretrainedModel.js') }}"></script>
```
lukeyeager: Don't you want `ver=dir_hash` here? From https://github.com/NVIDIA/DIGITS/pull/904?

lukeyeager:
1. I thought this was going to include the fix where the job finally shows up in the list of jobs at the top of the page while running? I'm still just seeing nothing when I'm redirected to the home page until I refresh.

2. I got a strange error in my console that wasn't showing up anywhere in the UI:

        2016-08-25 16:06:16 [20160825-160614-e42f] [ERROR] IOError: [Errno 21] Is a directory: '/raid/jobs/dev/20160810-142009-af06/'

@Lucaszw can you make a list of the changes included in this pull request and put it in the PR header?

digits/templates/partials/home/upload_pretrained_model.html
lukeyeager: Nevermind, I did it in https://github.com/NVIDIA/DIGITS/pull/998

lukeyeager:
@gheinrich can you take a look at this tomorrow?

digits/model/tasks/caffe_train.py
```diff
@@ -1128,18 +1123,26 @@ def get_task_stats(self,epoch=-1):
...
+        image_dimensions = list(self.dataset.get_feature_dims())
+
+        # TODO: Store both original, and cropped image dimensions
+        if self.crop_size is not None:
+            image_dimensions[0] = self.crop_size
+            image_dimensions[1] = self.crop_size
+
         stats = {
-            "image dimensions": self.dataset.get_feature_dims(),
+            "image dimensions": tuple(image_dimensions),
```
gheinrich: on below line, this way of getting the mean file doesn't work for generic datasets (mean file is not located in root of job directory but in `train_db` sub-folder).

digits/framework_helpers/caffe_helpers.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+    Set a new include rule for this layer
+    If rule is None, the layer will always be included
+    """
+    layer.ClearField('include')
+    layer.ClearField('exclude')
+    if rule is not None:
+        layer.include.add().CopyFrom(rule)
+
+def save_deploy_file_classification(network,path,num_categories, crop_size=None, feature_dims=None,logger=None):
```
gheinrich: if we are saving the deploy file then perhaps we don't need the label classes anymore in the pre-trained model?

gheinrich: is there an equivalent for generic networks?

digits/framework_helpers/caffe_helpers.py
```diff
@@ -0,0 +1,269 @@
...
+    if logger:
+        logger.debug("Network sanity check - deploy")
+
+    net_sanity_check(deploy_network, caffe_pb2.TEST)
+    found_softmax = False
+    for layer in deploy_network.layer:
+        if layer.type == 'Softmax':
+            found_softmax = True
+            break
+    assert found_softmax, 'Your deploy network is missing a Softmax layer! Read the documentation for custom networks and/or look at the standard networks for examples.'
```
gheinrich: this isn't working for generic networks that don't have a Softmax...

digits/model/views.py
```diff
@@ -205,30 +195,54 @@ def to_pretrained(job_id):
...
+        epoch = float(flask.request.args['epoch'])
+
+    # POST ?snapshot_epoch=n (from form)
+    elif 'snapshot_epoch' in flask.request.form:
+        epoch = float(flask.request.form['snapshot_epoch'])
 
-    return flask.redirect(flask.url_for('digits.views.home',tab=3)), 302
+    username = auth.get_username()
 
+    create_pretrained_model(job_id,username,epoch)
```
gheinrich: Quite often the pretrained model job hasn't completed by the time we get redirected to the home page and therefore we don't see the new job unless we refresh the home page. Can we add a call to `job.wait_completion()` before the redirection?

gheinrich:
Sorry it's a bit late for making this comment but the way tasks are created in `PretrainedModelJob` breaks framework agnosticity. We should refrain from doing things like `if torch then ... elif caffe ...`. Instead we should add APIs in `framework.py` and call the common interface. This way, the code would work for any number of frameworks.

digits/framework_helpers/caffe_helpers.py
Lucaszw: @gheinrich Thats a good point! Since pretrained models can be from outside of digits I didn't include it in the upload process, plus they are also used for fine tuning. Although perhaps we could have check if a deploy file exists, and if so ignore the generate deploy step?

digits/framework_helpers/caffe_helpers.py
Lucaszw: @gheinrich Whoops! I will fix this in my next commit

digits/model/views.py
Lucaszw: @gheinrich Yeah it seems to happen when there are many jobs to load. Ill fix this with my next commit .

lukeyeager:
Closing for now (there are merge conflicts anyway).

I've backed up this work at https://github.com/lukeyeager/DIGITS/tree/backup/lucas/pretrainedModelUpdates.

PR #1001, opened 2016-08-26
Let digits can create rec database

SevenGong:
I want to add a framework 'mxnet',so I first need to add a function to
let digits can create rec data. The is the modified part including
'/tools/create_db.py' and 'dataset/image/classification/forms.py'. But
it can not work. I need your helps. Thankyou very much. What else should
I modify?

### DIFF
--- a/digits/dataset/images/classification/forms.py
+++ b/digits/dataset/images/classification/forms.py
@@ -19,6 +19,7 @@ class ImageClassificationDatasetForm(ImageDatasetForm):
     backend = wtforms.SelectField('DB backend',
             choices = [
                 ('lmdb', 'LMDB'),
+				('rec', 'REC'),
                 ('hdf5', 'HDF5'),
                 ],
             default='lmdb',
@@ -26,6 +27,8 @@ class ImageClassificationDatasetForm(ImageDatasetForm):
 
     def validate_backend(form, field):
         if field.data == 'lmdb':
+            form.compression.data = 'none'
+		elif field.data == 'rec':
             form.compression.data = 'none'
         elif field.data == 'hdf5':
             form.encoding.data = 'none'
--- a/tools/create_db.py
+++ b/tools/create_db.py
@@ -13,6 +13,7 @@ import shutil
 import sys
 import threading
 import time
+import Queue
 
 # Find the best implementation available
 try:
@@ -34,8 +35,11 @@ from digits import utils, log
 # must call digits.config.load_config() before caffe to set the path
 import caffe.io
 import caffe_pb2
+import mxnet as mx
 
 logger = logging.getLogger('digits.tools.create_db')
+
+PREFIX_MXNET='images'
 
 class Error(Exception):
     pass
@@ -217,7 +221,7 @@ def create_db(input_file, output_dir,
     image_width -- image resize width
     image_height -- image resize height
     image_channels -- image channels
-    backend -- the DB format (lmdb/hdf5)
+    backend -- the DB format (lmdb/hdf5/rec)
 
     Keyword arguments:
     resize_mode -- passed to utils.image.resize_image()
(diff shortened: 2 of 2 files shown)


gheinrich:
Can MXNET not work with LMDB or HDF5?

SevenGong:
@gheinrich  Mxnet seems to support hdf5 data. I let digits can create rec data because I want to add more frameworks as mxnet, tensorflow, keras, and cntk to work with multi-GPUs and mutil-machines. However every frameworks use own data format, like caffe with lmdb, mxnet with rec, tensorflow with tfrecord and cntk with other data format. So creating a kind of data format is a need.  
I am not familiar in deep learnng field. Can you give me some suggests? Should I try to add a function for digits to create new data format, or use the hdf5 or tansfer the lmdb which digits have supported when using other frameworks? Which way is more efficient and simpler to add ? Looking for your suggests.

tools/create_db.py
```diff
@@ -297,6 +331,121 @@ def create_db(input_file, output_dir,
...
+	
+def _create_rec(image_count,images_folder, images_list_mx,image_height,image_width,
+               resize_mode,image_channels,encoding,output_dir,prefix,num_threads,mean_files):
+	source = images_list_mx
+    tic = [time.time()]
+    total = len(source)
+	
+    def image_encode(item, q_out):
+        try:
+		    img = cv2.imread(item[1])
```
lukeyeager: You never import cv2 - this will fail

tools/create_db.py
```diff
@@ -34,9 +35,12 @@
 # must call digits.config.load_config() before caffe to set the path
 import caffe.io
 import caffe_pb2
+import mxnet as mx
```
lukeyeager: If we do add support for mxnet, it will certainly be optional. Not required for everyone like this.

digits/dataset/images/classification/forms.py
```diff
@@ -19,6 +19,7 @@ class ImageClassificationDatasetForm(ImageDatasetForm):
     backend = wtforms.SelectField('DB backend',
             choices = [
                 ('lmdb', 'LMDB'),
+				('rec', 'REC'),
```
lukeyeager: We don't use tabs in DIGITS code. You have tabs all over the place in this PR.

tools/create_db.py
```diff
@@ -297,6 +331,121 @@ def create_db(input_file, output_dir,
...
+                break
+            record.write(s)
+            sink.append(item)
+            if len(sink) % 1000 == 0:
+                cur_time = time.time()
+                print 'time:', cur_time - pre_time, ' count:', len(sink)
+                pre_time = cur_time
+
+    try:
+        import multiprocessing
```
lukeyeager: Imports go at the top of the file, unless there's a good reason not to.

lukeyeager:
@SevenGong thanks for the contribution. But if we add support for mxnet, we will do it right. We would need to support a data format, sure, but we would also need to make mxnet an optional requirement. We would need to add support for training and inference. We would need to add build instructions and usage examples. We would need to figure out how to package it up and release it alongside Caffe and Torch.

I'm closing this PR because it adds no value without at least partial support for training with mxnet. But thanks for showing us a small part of what it might take to add support for mxnet.

SevenGong:
@lukeyeager I am very sorry I just see your comment. Now I have modified this indent problem. And now I trace my PR and notice the problem is at [https://github.com/NVIDIA/DIGITS/blob/v4.0.0/digits/task.py#L208](url) . My code walk to this place and stop here, and it cannot walk next step to run the digits/tools/create_db.py. I was wondering. Do you know why? 
As for any other codes of my RP, it seems to be right, no errors.

SevenGong:
@lukeyeager I am very sorry I just see your comment. Now I have modified this indent problem. And now I trace my PR and notice the problem is at [https://github.com/NVIDIA/DIGITS/blob/v4.0.0/digits/task.py#L208](url) . My code walk to this place and stop here, and it cannot walk next step to run the digits/tools/create_db.py. I was wondering. Do you know why? 
As for any other codes of my RP, it seems to be right, no errors.

tools/create_db.py
SevenGong: @lukeyeager Thank you. I imitate the source of the mxnet. But your advice is good.

tools/create_db.py
SevenGong: Hello.Why cannot import cv2 here?

PR #1003, opened 2016-08-26
Add custom delete job timeouts

Lucaszw:
Add custom delete job timeouts

### DIFF
--- a/digits/job.py
+++ b/digits/job.py
@@ -22,6 +22,7 @@ class Job(StatusCls):
     Base class
     """
     SAVE_FILE = 'status.pickle'
+    NON_PERSISTENT_JOB_DELETE_TIMEOUT_SECONDS = 3600
 
     @classmethod
     def load(cls, job_id):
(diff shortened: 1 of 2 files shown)


digits/job.py
```diff
@@ -22,6 +22,7 @@ class Job(StatusCls):
     Base class
     """
     SAVE_FILE = 'status.pickle'
+    NON_PERSISTENT_JOB_DELETE_TIMEOUT_SECONDS = 3600
```
gheinrich: that makes it a class attribute, is this what you want? or shall this be an instance attribute?

digits/job.py
```diff
@@ -349,6 +350,12 @@ def is_persistent(self):
         """
         return self.persistent
 
+    def delete_timeout(self):
```
gheinrich: it might be easier to read if you name this `get_delete_timeout()` (and add a `set_delete_timeout()` too for symmetry). Otherwise it may sound as if this function deletes the timeout.

digits/job.py
Lucaszw: @gheinrich I feel like setting job time out would be more class specific then instance specific. Ie all get_weights jobs would have the same timeout, vs individual instances of get_weights jobs.

digits/job.py
Lucaszw: @gheinrich Totally see what you mean! I changed this in my latest commit.

lukeyeager:
Sorry, but how is this a bug? This PR doesn't appear to change any behavior.

Lucaszw:
@lukeyeager My bad, yeah this should be an enhancement.

lukeyeager:
Closing because there are merge conflicts from https://github.com/NVIDIA/DIGITS/pull/1207 and because this was a no-op change anyway.

PR #1004, opened 2016-08-26
Change file concat to os path join

Lucaszw:
Change file concat to os path join

### DIFF
--- a/digits/pretrained_model/tasks/caffe_upload.py
+++ b/digits/pretrained_model/tasks/caffe_upload.py
@@ -20,14 +20,14 @@ class CaffeUploadTask(UploadPretrainedModelTask):
         """
         Get path to model definition
         """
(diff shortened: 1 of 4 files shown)


digits/pretrained_model/tasks/caffe_upload.py
```diff
@@ -20,14 +20,14 @@ def get_model_def_path(self):
         """
         Get path to model definition
         """
-        return self.job_dir+"/original.prototxt"
+        return os.path.join(self.job_dir,"original.prototxt")
```
gheinrich: it's maybe just me but I like to see a space after commas in argument lists...

Lucaszw: @gheinrich That sounds reasonable! I added them in my last commit :)

lukeyeager:
Again, Travis failures are unrelated.

Issue #910, opened 2016-07-15
Build DIGITS on Windows


JieZou1:
@IsaacYangSLA 

I have followed the steps described in  docs / BuildDigitsWindows.md to try run DIGITS on Windows 7, but I receive the following error message in the last step, running "python digits-devserver"

------------------------------------
Traceback (most recent call last):
  File "digits-devserver", line 31, in <module>
    from digits import config
  File "D:\digits\digits\config\__init__.py", line 7, in <module>
    from .current_config import config_value
  File "D:\digits\digits\config\current_config.py", line 4, in <module>
    from .caffe_option import CaffeOption
  File "D:\digits\digits\config\caffe_option.py", line 14, in <module>
    from digits.utils import parse_version
  File "D:\digits\digits\utils\__init__.py", line 158, in <module>
    from . import constants, image, time_filters, errors, forms, routing, auth
  File "D:\digits\digits\utils\image.py", line 15, in <module>
    import PIL.Image
ImportError: No module named PIL.Image
--------------------------------------

I am using Anaconda, the Python version is:

Python 2.7.11 |Anaconda 4.0.0 (64-bit)| (default, Feb 16 2016, 09:58:36) [MSC v. 1500 64 bit (AMD64)] on win32

I do have installed all the modules listed in the requirests.txt. I also have installed image module, which is not included in Anaconda.

Any ideas what goes wrong? Many thanks

TimZaman:
Seems you should also download the Pillow package as you downloaded the other packages and then run
`python -m pip install Pillow-3.3.0-cp27-cp27m-win_amd64.whl`

IsaacYangSLA:
Yes.  What Tim said was right.  Please install Pillow.  That package is the first one in requirements.txt.  I will check if my instruction does not cover that.  Thanks.

JieZou1:
@TimZaman @IsaacYangSLA 

Many thanks, Tim and Isaac, previously, I have pillow 3.2.0-py27_0 installed. After upgrading the pillow to 3.2.0-py27_1, DIGITS does successfully run on my Win7. The version is  higher than that listed in the requirements.txt (Pillow>=2.3.0,<=3.1.2), but DIGITS does still seem running Okay. Many thanks indeed!

IsaacYangSLA:
@JieZou1 , great to hear that.  I will close this issue soon.  However, please feel free to comment in case you encounter the same issue again.

Chachay:
[...]

PR #1007, opened 2016-08-26
Add how to install Pillow in Windows.

IsaacYangSLA:
Test DIGITS master branch with new requirements.txt on Windows.  The Pillow package can't be installed via pip due to missing jpeg library.  Users can install Pillow 3.1.1 via Windows installer executable instead.

### DIFF
--- a/docs/BuildDigitsWindows.md
+++ b/docs/BuildDigitsWindows.md
@@ -72,6 +72,15 @@ Then type
 python -m pip install -r requirements.txt
 ```
 
+You may see error about Pillow, like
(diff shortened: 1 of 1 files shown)


lukeyeager:
We haven't updated the PIL requirement in a while - was this not broken before?

IsaacYangSLA:
It worked fine without issue, at least before July 15 (see https://github.com/NVIDIA/DIGITS/issues/910).  This time, I tried a fresh installation and saw the jpeg library error.  So I need to update the instruction.  Another way is to install pre-build whl from UCI, but its version is too high.

PR #1009, opened 2016-08-27
Gradient Ascent for Caffe and Torch + Layer Viz

Lucaszw:
Do Not Merge!
<b> Gradient Ascent For Torch and Caffe + Layer Visualizations. </b> 
Things left to do: 

- Add receptive field calculation for Torch . Perhaps this step could be moved out of tools, and into the views before rendering (giving the user options to dev # deviations, size of bounding box etc. 

- Customizeable parameters: Currently these are hardcoded into the python and lua scripts. It would be nice to have a few set of options available from the UI like: "Best Alexnet" or "Best Googlenet" etc. 

- Additional Regularization Options (I didn't have time to implement all the params from Deep Vis Toolbox)

- Use mean file during Gradient Ascent. (The mean file is sent as a param for both jobs but not used, as it seems to make certain black and white jobs like LeNet(MNIST) return odd regularizations) . Perhaps this could be toggles with a checkbox in the UI


Lucaszw:
screen1

Layer Visualizations - Caffe

Lucaszw:
screen2

Layer Visualizations - Torch (Notice Lack of Receptive Field Calculations)

Lucaszw:
snap3

Gradient Ascent - Torch

Lucaszw:
snap4

TimZaman:
How does this PR relate to #968 and #937?

digits/pretrained_model/tasks/torch_upload.py
```diff
@@ -30,6 +41,30 @@ def get_weights_path(self):
         return os.path.join(self.job_dir, "_Model.t7")
 
     @override
+    def get_deploy_path(self):
+        """
+        Get path to file containing model def for deploy/visualization
+        """
+        return os.path.join(self.job_dir,"original.lua")
```
TimZaman: Maybe just call this `network.lua` for consistency? Or will there be more/'unoriginal' models?

tools/torch/toGraph.lua
```diff
@@ -0,0 +1,260 @@
+-- Copyright (c) 2015-2016, NVIDIA CORPORATION. All rights reserved.
+require 'torch'
+require 'pl'
+
+opt = lapp[[
+-n,--network (default '') path to network file
+-o,--output (default '') output path of model_def.json
+]]
```
TimZaman: Neat! How hard would it be for you to include parameters the output json per node? Why is the type information included in both {'source' and 'target'} of the link and the {node} itself?
Also at this point maybe check that the `network` and `output` options are defined at all (for `cli` usage).

digits/pretrained_model/tasks/torch_upload.py
Lucaszw: @TimZaman This PR is just of my top of branch before finishing my internship. Its the same as those two combines, but rebased with master, and Gradient Ascent for Torch.

Lucaszw: @TimZaman Yeah network.lua would definitely be a better name. Ill try updating this next time I get a chance :)

tools/torch/toGraph.lua
Lucaszw: @TimZaman That should be pretty easy I think! As for the type info for the links, I think I was just using that for styling them (in nnTree.js). Although I don't think I used those properties in generating the graph itself.

digits/static/js/layer_visualizations/readCaffeModel.js
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  // *** Read Prototxt File ****
+  if (type == 'file'){
+    fileReader = new XMLHttpRequest();
+    fileReader.open("GET", text, false);
+
+    fileReader.onreadystatechange = function (){
+      if(fileReader.readyState === 4){
+        if(fileReader.status === 200 || fileReader.readyState == 4){
```
TimZaman: In line 50 you say
`if(fileReader.readyState === 4){`
and immediatelly after
`if(fileReader.status === 200 || fileReader.readyState == 4){`
So the latter is always true if the former is true. I bet you meant to say 
`(fileReader.status === 200 || fileReader.status == 0)` here.

digits/static/js/layer_visualizations/readCaffeModel.js
```diff
@@ -0,0 +1,163 @@
+function getTreeData(type,text){
```
TimZaman: Maybe add usage instructions

tools/get_weights_caffe.py
```diff
@@ -0,0 +1,144 @@
...
+        FD.TYPE_SINT64: long,
+    }
+    js = {}
+    fields = pb.ListFields()	#only filled (including extensions)
+    for field,value in fields:
+        if field.type == FD.TYPE_MESSAGE:
+            ftype = pb2json
+        elif field.type in _ftype2js:
+            ftype = _ftype2js[field.type]
+
```
TimZaman: else: unknown-type error, or use logger

lukeyeager:
Closing for now (there are merge conflicts anyway). I'd like to revisit this.

I've backed up this work at https://github.com/lukeyeager/DIGITS/tree/backup/lucas/gradientAscentWithLayerVis.

Lucaszw:
@lukeyeager @TimZaman Sorry school has been super busy. But hopefully soon I'll get some time to revisit this and create some smaller pull requests (such as layer visualizations for Torch models, etc.) once I have a machine up and running.

PR #1012, opened 2016-08-29
Add threshold option in image output extension

gheinrich:
Add option to apply binary threshold on output image:

vis

Values lower than threshold will be zeroed.
Values higher than threshold will be set to 255. 

cc @jbarker-nvidia


gheinrich:
Not need anymore, closing.

PR #1015, opened 2016-08-29
Fix mean subtraction in Object Detection walk-through

gheinrich:


lukeyeager:
You're right - this is a documentation bug. The mean subtraction is already done in the transform layer:
https://github.com/NVIDIA/caffe/blob/v0.15.9/examples/kitti/detectnet_network.prototxt#L100

PR #1021, opened 2016-08-30
allow DIGITS to function behind an SSL-terminating proxy

rubysolo:


lukeyeager:
Can you provide a little context for this PR?

rubysolo:
Sure.  We are running DIGITS behind nginx configured to run in SSL (only).  Without these changes, links inside the DIGITS app would lose the scheme and send requests to `http://server:443/$request_uri`, so we would get the error message "The plain HTTP request was sent to HTTPS port”.  The fix was to ensure that nginx was setting the `X-Forwarded-Proto` header and to add the `ProxyFix` module so that Flask would use the `X-Forwarded-Proto` header to determine the scheme for links/forms.

lukeyeager:
Are you sure this makes sense to add in general? I'm not sure what this means, but it doesn't sound good:

> Do not use this middleware in non-proxy setups for security reasons.
http://werkzeug.pocoo.org/docs/0.11/contrib/fixers/#werkzeug.contrib.fixers.ProxyFix

rubysolo:
Maybe we could have the `ProxyFix` stuff commented out with a note to only enable in this scenario?

IsaacYangSLA:
I guess if all traffic to DIGITS is SSL-terminated at nginx, the $schme in X-Forwarded-Proto does not change per each request.  Will listen 443 SSL; work in this case?

rubysolo:
Our nginx config already had `listen 443 ssl;`, but we still encountered the issue.

lukeyeager:
> Maybe we could have the ProxyFix stuff commented out with a note to only enable in this scenario?

We could do that, but obviously we wouldn't be testing the commented-out code and we wouldn't know when/if we break something. Thanks for making a public pull request for this - I'll point anyone else who has this issue here.

PR #1024, opened 2016-08-31
Modularization of inference

gheinrich:
If a dataset extension implements its get_inference_form() method then the standard image inference form is replaced with the extension's inference form:

inference-form

During inference, an inference dataset is created using the data from the inference form and the appropriate visualization can be used to render the output:

view

Tasks:

- [x] Create new `infer_extension` route
- [x] Create inference dataset
- [x] Add example (text classification)
- [x] Add tests

### DIFF
--- a/digits/extensions/data/imageGradients/forms.py
+++ b/digits/extensions/data/imageGradients/forms.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import
 
 from digits import utils
 from digits.utils import subclass
+from digits.utils.forms import validate_required_iff
 from flask.ext.wtf import Form
 import wtforms
 from wtforms import validators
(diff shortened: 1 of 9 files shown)


lukeyeager:
This approach looks great to me

gheinrich:
Added tests and an example "modularized" inference form for the image gradient dataset (where the user can specify the x and y gradient and get an image generated on the fly):
gradient-inference

digits/extensions/data/imageGradients/forms.py
```diff
@@ -55,3 +56,37 @@ class DatasetForm(Form):
...
+    A form used to perform inference on a gradient regression model
+    """
+
+    gradient_x = utils.forms.FloatField(
+        'Gradient (x)',
+        validators=[
+            validate_required_iff(test_image_count=None),
+            validators.NumberRange(min=-1, max=1),
+            ],
+        tooltip="Specify a number between -1 and 1"
```
lukeyeager: Range should be [-0.5,0.5] instead. Otherwise you get this for (1,1):

gradients-clipping

gheinrich: Whoops, that looks good, but is not what we want :-)

lukeyeager:
LGTM

PR #1025, opened 2016-08-31
Text classification workflow

gheinrich:
~~Depends on #927 (support for plug-ins) and #1024 (modularization of inference).~~


IsaacYangSLA:
@gheinrich , in encode_entry(), if one char is encoded as k, and another char is encoded as k+1, does that imply these two char's are close to each other?  I mean encoding characters into scalar, rather than one-hot-encoding, seems not common.  For text classification, word2vec seems a more popular to encode document into vector space.

gheinrich:
Hi @IsaacYangSLA, in the example network that I provide, the [first layer](https://github.com/NVIDIA/DIGITS/blob/v4.0.0/examples/text-classification/text-classification-model.lua#L38-Lundefined) is doing one-hot encoding of the characters. I chose to do the one-hot encoding in the network rather than in the dataset because that results in a **much** more compact dataset, especially if you have a large alphabet.

It's just my opinion, but I think "word2vec" kind of defeats the purpose of deep learning: you need logic outside of the network, like stemming algorithms, to identify words. I suppose the popularity of word2vec comes down to the limited memory/compute capabilities. I think a character-level representation of the data should be ultimately more powerful, similar to how Deep Neural Nets outperform HOG+SVM in image classification.

IsaacYangSLA:
Hi @gheinrich , thanks for the information of first layer.  That's a better design, I agree.

For the word2vec or character part, it seems more people use word2vec in NLP applications and the idea behind it is also reasonable, i.e. the concept of DC - USA + France ~= Paris.  However, I see increasing researches are now on character-based text processing.  Maybe in a few years, it will outperform word2vec in NLP applications.

gheinrich:
rebased on tip of master branch

plugins/data/textClassification/setup.py
```diff
@@ -0,0 +1,24 @@
...
+def read(fname):
+    return open(os.path.join(os.path.dirname(__file__), fname)).read()
+
+setup(
+    name="digits_text_classification_data_plugin",
+    version="0.0.1",
+    author="Greg Heinrich",
+    description=("A data ingestion plugin for text classification"),
+    long_description=read('README'),
+    license="Apache",
```
lukeyeager: Why Apache and not BSD-3?

plugins/view/textClassification/digitsViewPluginTextClassification/view.py
```diff
@@ -0,0 +1,118 @@
...
+        for idx in input_data[0]:
+            c = self.alphabet[idx - 1] if idx < self.alphabet_len else '.'
+            input_characters.append(c)
+        input_string = ''.join(input_characters)
+
+        # assume the only output is from a probability distribution
+        scores = output_data[output_data.keys()[0]].astype('float32')
+
+        if self.terminal_layer_type == "logsoftmax":
+            scores = np.exp(scores)
```
lukeyeager: Could you just check to see if the values sum to 1 instead of having this form field?

gheinrich: good idea (I guess I can also check if values are positive or negative)!

gheinrich:
I have updated the text classification example to show how to use the plug-ins

plugins/view/textClassification/digitsViewPluginTextClassification/view.py
```diff
@@ -0,0 +1,117 @@
...
+            c = self.alphabet[idx - 1] if idx < self.alphabet_len else '.'
+            input_characters.append(c)
+        input_string = ''.join(input_characters)
+
+        # assume the only output is from a probability distribution
+        scores = output_data[output_data.keys()[0]].astype('float32')
+
+        if np.max(scores) < 0:
+            # terminal layer is a logsoftmax
+            scores = np.exp(scores)
```
lukeyeager: :+1:

examples/text-classification/README.md
```diff
@@ -27,6 +29,41 @@ The following sample is an example from the "company" class:
...
+Point the `$DIGITS_ROOT` envvar to the top of your DIGITS installation then do:
+```sh
+$ pip install -e $DIGITS_ROOT
+```
+
+Install the text classifications plug-ins:
+```sh
+$ pip install $DIGITS_ROOT/plugins/data/textClassification
+$ pip install $DIGITS_ROOT/plugins/view/textClassification
+```
```
lukeyeager: :+1:

PR #1027, opened 2016-08-31
Update BuildCaffe.md

jianghaitao:
I was bitten by Coffee compilation errors when I found out that my gcc version is 4.9.3 due to the fact that I was trying to compile CUDA - which doesn't support gcc 5.0+. I think it is worth mentioning it here to save people's time on troubleshooting. It should be made clear that gcc 5 is required, at least on Ubuntu 16.04.

### DIFF
--- a/docs/BuildCaffe.md
+++ b/docs/BuildCaffe.md
@@ -39,6 +39,12 @@ If you hit some errors about missing imports, then use this command to install t
 cat $CAFFE_HOME/python/requirements.txt | xargs -n1 sudo pip install
 ```
 
+### Check gcc version
+On Ubuntu 16.04, you would need to make sure your gcc/g++ is of version 5 (5.4.0) not 4.9 or lower. Otherwise, you would encounter issues compiling Caffee, such as https://github.com/BVLC/caffe/issues/3438.
+```
+gcc --version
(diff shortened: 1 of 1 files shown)


lukeyeager:
Hi @jianghaitao. Thanks for trying to improve the documentation, but this is not what we recommend.

If you're running 16.04 (as I am), then you want to use CUDA 8.0 and gcc 5.x rather than CUDA 7.5 and gcc 4.x. Downgrading your gcc version is likely to cause a bunch of issues on your system. CUDA 8.0 is still in the release candidate phase, but the final release should come out relatively soon. Once CUDA 8.0 is released, that will work on either Ubuntu version.

Until then, you can download the RC here: https://developer.nvidia.com/cuda-toolkit

jianghaitao:
HI, Luke

Yes. I am running 16.04 / CUDA 8. I think the key thing is that we need to
make it clearly that user should make sure their gcc is of 5.0+ not 4.9 or
below when they compile Nvidia/Caffe. This was not clearly stated in the
original instruction I think...took me a bit time to figure it out.

Thanks

Haitao

Thanks

Haitao

lukeyeager:
Uh, how did you even end up with GCC 4.x on Ubuntu 16.04? The default version is 5.3:
http://packages.ubuntu.com/xenial/gcc

PR #1028, opened 2016-08-31
Print a little more info for errors in job.save()

lukeyeager:
I always end up with weird errors because I switch around between PR branches so often. This just adds a little more info to help debug the error messages. Shouldn't affect normal users at all.


gheinrich:
That looks OK to me.

PR #1045, opened 2016-09-06
Fix inconsistencies in view interface parameters

gheinrich:
Should be purely cosmetic.


lukeyeager:
I can't find where either `dataset` or `input_id` are ever used. Is that just a placeholder?

gheinrich:
`input_id` is the index of each sample from the list of samples to perform inference on. I thought this could be useful if one day we multi-thread the processing of visualizations and we need indices to get a notion of ordering. But you're right that for the time being, it's not used.

lukeyeager:
Ok, thanks for the explanation.

PR #1048, opened 2016-09-08
Rearranged group elements and pills on the front page

jmancewicz:
Moved the "Group Jobs" button to be closer to the well containing the table.
Moved the New Dataset and New Model pills to their respective tabs.
The "Upload Pretrained Model" has been moved to a new "Load Model" on the "Pretrained Models" tab.
Changed the "Group" button to be disabled when the "Group Jobs" is unchecked.

From this:
screen shot 2016-09-07 at 5 04 55 pm

To this:
screen shot 2016-09-07 at 5 12 57 pm

### DIFF
--- a/digits/templates/home.html
+++ b/digits/templates/home.html
@@ -52,10 +52,6 @@
     {[jc.model_jobs = (jobs | filter:is_model);'']}
     {[jc.pretrained_model_jobs = (jobs | filter:is_pretrained_model);'']}
     <section ng-controller="tab_controller as tab" ng-init="init({{request.args['tab']}})">
-        <label>Group Jobs:
-            <input type="checkbox"
-                   ng-model="jc.storage.show_groups">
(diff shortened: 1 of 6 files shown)


TimZaman:
Still seems a bit out of place. I don't have a better alternative though :disappointed:

jmancewicz:
Yeah, I think it's better than where it had been, but seems like it could be better. I'd considered another options cog dropdown for the entire page, but that seemed odd with only one option and like it obscured the functionality too much.

lukeyeager:
*(merge conflict)*

FWIW, I think the new layout is better. But I'm not very picky about UI.

IsaacYangSLA:
Thanks.  This will make it much easier to add model store UI.

lukeyeager:
@jmancewicz - can you deal with the merge conflicts?

PR #1049, opened 2016-09-08
Block Delete Jobs popup when no jobs are selected

jmancewicz:
Both the Delete and Abort buttons bring up the confirmation popup when the buttons is disabled.

### DIFF
--- a/digits/static/js/home_app.js
+++ b/digits/static/js/home_app.js
@@ -384,6 +384,7 @@ try {
 
         $scope.delete_jobs = function() {
             var job_ids = $scope.get_selected_job_ids();
+            if (job_ids.length == 0) return;
             bootbox.confirm(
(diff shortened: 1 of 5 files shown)


lukeyeager:
Verified bug and fix. Thanks!

digits/static/js/home_app.js
```diff
@@ -384,6 +384,7 @@ try {
 
         $scope.delete_jobs = function() {
             var job_ids = $scope.get_selected_job_ids();
+            if (job_ids.length == 0) return;
             bootbox.confirm(
                 ('Are you sure you want to delete the selected' +
```
TimZaman: You forgot a white space after 'selected'

jmancewicz: Thanks.

PR #1058, opened 2016-09-12
Move digits.__version__ into new file and add tests

lukeyeager:
This lets you query the version from the command-line (useful for debian packaging) and inside a `setup.py` script (useful for PyPI/conda packaging) without actually having to import DIGITS. We might actually add some code to `__init__.py` someday, and moving this into a separate file now helps with tool maintainability.

Also, rather than simply writing a comment about how the file is supposed to be maintained and updated, I wrote some tests to enforce it.


gheinrich:
That looks good to me!

PR #1060, opened 2016-09-13
Various small ObjectDetection dataset improvements

### LINKED ISSUE #940: Object detection file names required to be numeric
When creating an Object Detection dataset, it seems like DIGITS requires the file names to be numeric. When I tried to create a dataset with filenames that contained alphabetic characters, I got the error `ERROR: ValueError: invalid literal for int() with base 10: 'ba'` in the sidebar under Create train_db DB where `ba.txt` and `ba.bmp` were the names of two of my files. When I changed the filenames to include only numeric characters (eg. `00.txt` or `00.bmp`), the dataset was successfully created.

I'm not sure if this requirement actually exists or is the result of a different bug, but it would be nice if DIGITS could handle alphabetic file names or if there was at least a warning on the UI.

lukeyeager:
* Allow non-numeric filenames (close #940)
  * This should also address the issue on Windows where, because symlinks are not supported, the `prepare_kitti_data.py` script creates filenames like `000000039_002994.png`.
* Improve documentation about case sensitivity
* Improve documentation about mapping classes to 0 during dataset creation

### DIFF
--- a/digits/extensions/data/objectDetection/README.md
+++ b/digits/extensions/data/objectDetection/README.md
@@ -160,4 +160,4 @@ Class name | Class ID
 dontcare | 0
 pedestrian | 1
 
-All labeled objects other than "pedestrian" in your dataset will be mapped to 0.
+All labeled objects other than "pedestrian" in your dataset will be mapped to 0, along with any objects explicitly labeled as "dontcare".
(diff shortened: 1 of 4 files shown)


gheinrich:
That looks OK to me.

PR #1061, opened 2016-09-13
Improve handling of empty/invalid object detection label files

### LINKED ISSUE #1054: Validation images with no objects
Hello,
I was trying to create a validation object detection dataset which contains images with no objects. First i tried to create it using empty label files, but digits gave me the error ""Label key 0000 not found in label folder".

I handled this issue by adding a "dummy" label for these images. But i think this is not the best solution.

lukeyeager:
*Close #1054*

An empty list of labels is a valid use case.


gheinrich:
That looks OK to me. Maybe add a reference to #1054 ?

PR #1070, opened 2016-09-15
cookie.tmp replaced by digits.cookie

jpfairbanks:
Hello digits maintainers,

I was working through this example and noticed that the cookie file didn't match in the api examples. Was it supposed to be digits.cookie or cookie.tmp? I just changed it to digits.cookie everywhere and it worked for me. Here is a patch if you would like to make them match.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -43,21 +43,21 @@ In order to create a dataset, you will first need to log in.
 The following command will log us in as user `fred`:
 
 ```sh
-$ curl localhost/login -c cookie.tmp -XPOST -F username=fred
+$ curl localhost/login -c digits.cookie -XPOST -F username=fred
 <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
 <title>Redirecting...</title>
 <h1>Redirecting...</h1>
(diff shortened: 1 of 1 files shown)


gheinrich:
Hello, I don't understand. You can name this file whatever you like as long as you're consistent in its use. Are you sure you didn't accidentally misspell the file in one of the commands?

jpfairbanks:
According to the instructions, you were supposed to save the login cookie into cookie.tmp but then when you used it, you were supposed to load the cookie from digits.cookie. I think it was renamed but not updated everywhere. 

For example, line 46 uses `cookie.temp` and line 71 uses `digits.cookie` as the filename for the cookie to be stored in.

line 43 on master
```sh
curl localhost/login -c cookie.tmp -XPOST -F username=fred
```
line 71 on master
```sh
curl localhost/datasets/images/classification.json -b digits.cookie -XPOST -F folder_train=$MNIST_PATH/train -F encoding=png -F resize_channels=1 -F resize_width=28 -F resize_height=28 -F method=folder -F dataset_name=mnist_dataset
```

We agree that these filenames should be the same, right?

gheinrich:
@jpfairbanks you're correct, sorry! I was the one being inconsistent. I apologize.

jpfairbanks:
It's all good.

PR #965, opened 2016-08-09
Add support for Ubuntu 16.04

lukeyeager:
Several key packages have been upgraded from 14.04 to 16.04.

* gevent 1.0 -> 1.1.0
* gunicorn 17.5 -> 19.4.5

The gevent upgrade is particularly troublesome because it requires upgrading Flask-Socketio. Which isn't so bad in and of itself, but causes problems for packaging because of a naming conflict (see https://github.com/NVIDIA/DIGITS/pull/476#issuecomment-165268248, https://github.com/miguelgrinberg/python-socketio/issues/23). I can get around the packaging issue, but it's not pretty.

Another issue is that some versions of gunicorn and gevent conflict. So while requirements says that you can use `gevent>=1.0,<=1.1.0` and/or `gunicorn>=17.5,<=19.4.5`, it should really read `(gevent==1.0 AND gunicorn==17.5) OR (gevent==1.1.0 AND gunicorn==19.4.5)`. Otherwise you get this weird behavior where the Caffe subprocess goes defunct.

I also found a funny pydot packaging bug along the way (see commit msg).

PR #1072, opened 2016-09-15
Install a few more deb packages on TravisCI

lukeyeager:
Should shave off a little time.

`python-matplotlib` because https://github.com/NVIDIA/DIGITS/pull/961
`python-gevent-websocket` because https://github.com/NVIDIA/DIGITS/pull/965


lukeyeager:
Yep, shaved off about a minute.

PR #1073, opened 2016-09-15
Remove changes to socketio logger

lukeyeager:
I can't remember why I had added this originally. Some older version of socketio wasn't interacting well with my custom logger class or something. I want to remove it now because there are a bunch of socketio messages clogging up the testsuite output.


lukeyeager:
Well this helps a little bit. It removes the messages from the "captured stdout" section but not the "captured logging" section.

PR #1076, opened 2016-09-15
[Tests] More time for data aug model to converge

lukeyeager:
Like https://github.com/NVIDIA/DIGITS/pull/1074, this is trying to get rid of all the meaningless TravisCI errors.


lukeyeager:
Ran into this again. Merging.

PR #1077, opened 2016-09-16
Model store

IsaacYangSLA:
Depends on #1048, will rebase/squash later after #1048 is merged.


IsaacYangSLA:
Just rewrote the store option to utilize env vars to be consistent with PR1091.  Will re-verify the conflict after PR1091 is merged.

digits/config/store_option.py
```diff
@@ -0,0 +1,33 @@
...
+    def config_file_key():
+        return 'model_store'
+
+    @classmethod
+    def prompt_title(cls):
+        return 'Model Store URL list'
+
+    @classmethod
+    def visibility(cls):
+        return config_option.Visibility.HIDDEN
```
lukeyeager: Can you make this `DEFAULT` instead? Otherwise I have to use `python -m digits.config.edit -v` to edit it.

digits/config/store_option.py
```diff
@@ -0,0 +1,33 @@
...
+    def suggestions(self):
+        self.url_list = ['http://localhost/modelstore/']
+        return [prompt.Suggestion(self.url_list, 'H', desc='URL list')]
+
+    def apply(self):
+        self._config_dict_value = {
+            'url_list': self.url_list
+        }
+
+
```
lukeyeager: I think you're going to want to implement `validate()` somehow. Right now, it accepts "1234", "foo", etc.

digits/config/store_option.py
```diff
@@ -0,0 +1,33 @@
...
+    @classmethod
+    def visibility(cls):
+        return config_option.Visibility.HIDDEN
+
+    def optional(self):
+        return True
+
+    def suggestions(self):
+        self.url_list = ['http://localhost/modelstore/']
```
lukeyeager: How is the user going to input a list? This needs to be a string, possibly separated by a separator ([a comma seems like a good choice](http://stackoverflow.com/a/3867794/2404152)).

digits/static/store.html
```diff
@@ -0,0 +1,165 @@
+<!doctype html>
+<html lang="en">
+<head>
+    <link rel="icon" href="/static/images/nvidia.ico"/>
+    <meta charset="UTF-8">
+    <title>Model Store</title>
+    <!-- Latest compiled and minified CSS -->
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css"
+          integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous">
```
lukeyeager: Please don't add links that require internet access to download. Either check them into our source tree under `digits/static/js/` or pick another library.

digits/static/store.html
```diff
@@ -0,0 +1,165 @@
...
+    <!-- jQuery (necessary for Bootstrap's JavaScript plugins) -->
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
```
lukeyeager: This one is already checked into our tree.

digits/static/store.html
```diff
@@ -0,0 +1,165 @@
...
+    <!-- Latest compiled and minified JavaScript -->
+    <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.min.js"
+            integrity="sha384-Tc5IQib027qvyjSMfHjOMaLkfuWVxZxUPnCJA7l2mCWNIpG9mGCD8wGNIcPD7Txa"
+            crossorigin="anonymous"></script>
+    <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.5.8/angular.min.js"></script>
+    <script src="/static/js/store.js"></script>
```
lukeyeager: Please use this format for files stored under `static/js/`:
https://github.com/NVIDIA/DIGITS/blob/v4.0.0/digits/templates/layout.html#L12

digits/store/views.py
```diff
@@ -0,0 +1,125 @@
+# Copyright (c) 2014-2016, NVIDIA CORPORATION.  All rights reserved.
+from __future__ import absolute_import
+
+import flask
```
lukeyeager: Sorry to nitpick, but can you put this `import flask` below the standard imports? And above the digits imports?

digits/templates/layout.html
```diff
@@ -69,6 +69,7 @@
                     <li><span class="navbar-text"><a target="_blank" href="https://github.com/NVIDIA/DIGITS/tree/master/docs">DIGITS docs</a></span></li>
                     <li><span class="navbar-text"><a target="_blank" href="https://github.com/NVIDIA/DIGITS">DIGITS on github</a></span></li>
                     <li><span class="navbar-text"><a target="_blank" href="https://groups.google.com/forum/#!forum/digits-users">DIGITS User Group</a></span></li>
+                    <li hidden><span class="navbar-text"><a target="_blank" href="https://developer.nvidia.com/digits/nvzoo">NVidia model store for DIGITS</a></span></li>
```
lukeyeager: Can we use NVIDIA here instead of NVidia?

digits/config/store_option.py
IsaacYangSLA: Sure.  That's also what I got confused about.  Thanks for the suggestion.

digits/templates/layout.html
IsaacYangSLA: No problem.

digits/store/views.py
IsaacYangSLA: Sure.  Sorry for the improper ordering of imports.

digits/static/store.html
IsaacYangSLA: I will look into those three 'scripts' and 'link' tags and modify them accordingly.

digits/config/store_option.py
IsaacYangSLA: I think I will implement it later in this PR.  The code should ensure that URL is supposed to be well-formatted, not random strings.

digits/views.py
```diff
@@ -90,6 +90,11 @@ def home(tab=2):
                     'url': flask.url_for(
                         'digits.pretrained_model.views.new'),
                     },
+                'access-model-store': {
+                    'title': 'Retrieve from Model Store',
+                    'id': 'retrieveModelStore',
+                    'url': '/static/store.html',
```
lukeyeager: Let's make this an actual Flask route and not just a link to a static HTML file.

TimZaman:
I have some trouble setting the custom URL. I do this:
`python -m digits.config.edit -v`
then I input as Model Store URL List: `['http://foo.bar']`

Then in `digits/store/views.py` I do for debugging `print(app.config['store_url_list'])` it prints out: `['http://localhost/modelstore/']` ?

So

1. For me (I might be using it wrong) it does not want to retain the static url
2. The bogus url should communicate to the front-end that it cannot be reached (404).

digits/views.py
IsaacYangSLA: Thanks.  Yes, it will be much better to do flask.url_for on this.  I changed codes to reflect this.

IsaacYangSLA:
@TimZaman , my bad.  The prompt portion of store codes was broken.  I rewrote that portion and it should work now.

digits/config/store_option.py
```diff
@@ -15,19 +18,34 @@ def prompt_title(cls):
...
+
 
     def optional(self):
         return True
 
     def suggestions(self):
-        self.url_list = ['http://localhost/modelstore/']
-        return [prompt.Suggestion(self.url_list, 'H', desc='URL list')]
+        self.url_list = ''
+        return [prompt.Suggestion(self.url_list, 'M', desc='Model Store URL list (comma separated)')]
```
lukeyeager: Please put the format requirements in `prompt_message()` above. I don't think you actually need any suggestions for now.

TimZaman:
I can see why we want this in the next release, it's really cool!

digits/config/store_option.py
```diff
@@ -15,19 +18,34 @@ def prompt_title(cls):
...
+        return config_option.Visibility.DEFAULT
+
+    @classmethod
+    def validate(cls, value):
+        valid_url_list = list()
+        if isinstance(value, str):
+            url_list = value.split(',')
+            for url in url_list:
+                if url is not None and urlparse(url).scheme != "" and not os.path.exists(url):
+                    valid_url_list.append(url)
```
lukeyeager: If it's not valid, throw an error.

IsaacYangSLA: Yes.  I should throw BadOption and let user to enter the URL again.

digits/config/store_option.py
IsaacYangSLA: I also wondered what should be the suggested URL.  Now it's much easier for me for this option.

digits/config/store_option.py
IsaacYangSLA: The validate method is now throwing BadValue exception whenever it encounters invalid url inside url_list loop.

TimZaman:
I do think we need to make the download progress visible (and a loading indicator when retrieving details from the particular model store)

docs/ModelStore.md
```diff
@@ -0,0 +1,54 @@
...
+
+
+## Create your own Model Store server
+At the top directory, create a master.json file (if your server does not support directory listing).
+The following is a sample master.json file.
+```
+{'msg':'This is my own model store server.', 'children':['Model01','Model02']}
+```
+Model01 and Model02 are subdirectories containing the actual models.
+Each model must consist one weight file, one info.json file.
```
gheinrich: also mention that the file pointed to by `model_file` in `info.json` (e.g. `original.protoxt`) must be there.
otherwise a lot of things silently fail (network customization, ...)

docs/ModelStore.md
```diff
@@ -0,0 +1,54 @@
+# Model Store
+
+
+## Introduction
+Model Store is a new feature in DIGITS.
```
IsaacYangSLA: Yes, it's been a long time.  Let me remove that. :-)

docs/ModelStore.md
```diff
@@ -0,0 +1,54 @@
...
+If multiple model stores are available, specify their url's, separated by the comma (,).
+``` shell
+export DIGITS_MODEL_STORE_URL='http://localhost/mymodelstore,http://dlserver/teammodelstore'
+```
+
+
+## Browse and import model
+First launch DIGITS.
+Click Pretrained Models tab, and select 'Retrieve from Model Store' under 'Load Model.'
+The new page shows models available in model stores.
```
IsaacYangSLA: I also think screenshots would be much better than sentences after sentences (for most readers like me).  I will add some.

IsaacYangSLA:
@gheinrich , really thank you for reviewing the codes and excellent comments.  I just pushed the new commits to address each of your comment.  Since there are lots of areas to improve, and to save some space, I didn't add my reply to each of your comments.

gheinrich:
Thanks @IsaacYangSLA this looks fine though it still feels like the model store is somewhat hidden and like @TimZaman I feel it would be nice to show progress notifications, when downloading large models.

IsaacYangSLA:
@gheinrich , I believed you also found importing a model took a long time.  A progress notification will be really useful to users.  I will create another PR for that after this is merged.

digits/static/js/store.js
```diff
@@ -0,0 +1,30 @@
+(function(angular) {
```
lukeyeager: Missing copyright notice here

digits/config/store_option.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) 2015-2016, NVIDIA CORPORATION.  All rights reserved.
```
lukeyeager: Should be just "2016"

digits/store/views.py
```diff
@@ -0,0 +1,142 @@
+# Copyright (c) 2014-2016, NVIDIA CORPORATION.  All rights reserved.
```
lukeyeager: Should be just "2016"

digits/templates/store.html
```diff
@@ -0,0 +1,127 @@
+{# Copyright (c) 2014-2016, NVIDIA CORPORATION.  All rights reserved. #}
```
lukeyeager: Should be just "2016"

digits/utils/store.py
```diff
@@ -0,0 +1,54 @@
+# Copyright (c) 2014-2016, NVIDIA CORPORATION.  All rights reserved.
```
lukeyeager: Should be just "2016"

IsaacYangSLA:
@lukeyeager , thanks a lot.  To be honest, I didn't know the meaning of those years until you pointed them out.  It's great to know more.

IsaacYangSLA:
I will merge this and later create another PR for adding the Python layer file support on importing part, which depends on this PR.

PR #1084, opened 2016-09-18
Enhancement: Marking Mandatory fields in Dataset/Model forms

prabindh:
Mandatory fields are not marked, causing post-submit validation errors in both new Dataset and new Model forms. This minor patch adds marking of mandatory fields.

-Prabindh (psundareson@nvidia)

### DIFF
--- a/digits/dataset/forms.py
+++ b/digits/dataset/forms.py
@@ -12,7 +12,7 @@ class DatasetForm(Form):
     (abstract class)
     """
 
-    dataset_name = utils.forms.StringField(u'Dataset Name',
+    dataset_name = utils.forms.StringField(u'Dataset Name *',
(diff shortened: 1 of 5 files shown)


jmancewicz:
Although I like the idea, this really needs to be implemented based on the settings of the forms rather than explicitly per form. This is currently incomplete and would be difficult to maintain.

If we are going to add the 'required' asterisk, then it needs to be exhaustive and correct. Unfortunately, this is complicated by conditional validators.

There are attempts at javascript versions of wtforms validators, but the custom conditional validators may present a problem there too.

lukeyeager:
> This is currently incomplete and would be difficult to maintain.

I agree. Thanks for the contribution @prabindh - we'll keep this in mind for future enhancements.

PR #1088, opened 2016-09-19
Fix path given to utils.filesystem.dir_hash()

lukeyeager:
1. Don't assume `digits/` is in the CWD (found while testing setup.py)
2. Don't use platform-specific paths (doesn't this break on Windows?)


lukeyeager:
@jmancewicz can you give this a look and merge if it looks OK?

jmancewicz:
Looks good. I'll merge after TravisCI passes.

PR #1091, opened 2016-09-19
Use envvars for configuration instead of a file

### LINKED ISSUE #180: Add command-line overrides for configuration options
It would be nice to be able to override a configuration option from a config file with a command-line option.

The config file would not need to be rewritten - this would be a one-time change.

### LINKED ISSUE #941: instance-specific config capability needed
We work in a Linux-based HPC cluster environment. Users will use the queueing system to gain access to systems/resources, then run software out of a central installation (NFS mount). When we attempt to run digits-devserver, we run into errors because individual users can't create a jobs directory as they don't have write permissions into the software install. We've worked around this short-term by creating the digits/jobs directory (and opening it up for full read/write access), but to be able to support this in our environment we need the ability to have instance-specific configurations.

### LINKED ISSUE #1090: Subprocesses get incorrect configuration when DIGITS can't write to config file
To reproduce:
* Make digits/digit.cfg non-writeable

        chmod a-w digits/jobs/ digits/digits.cfg

* Start DIGITS with "digits-devserver" and change the jobs_dir to some place that you can write to
* Try to create a dataset

        Default value for jobs_dir "/home/lyeager/digits/digits/jobs" invalid:  You do not have write permission
        RuntimeError: No valid default value found for configuration option "jobs_dir"

lukeyeager:
*Close #941, close #180, fix #1090*

> Env vars are easy to change between deploys without changing any code; unlike config files, there is little chance of them being checked into the code repo accidentally; and unlike custom config files, or other config mechanisms such as Java System Properties, they are a language- and OS-agnostic standard.
https://12factor.net/config

Also,

* Once we add a working setup.py, multiple users will be able to start their own customized instance from the same installed codebase (#941)
* The ability to set the configuration interactively doesn't scale well as we continue to add options


gheinrich:
This seems to improve the time it takes to start tools (and this should be reflected on Travis test time too):

before:
```sh
$ time ./tools/inference.py --help
real	0m0.821s
```

After:
```sh
$ time ./tools/inference.py --help
real	0m0.438s
```

digits/config/jobs_dir.py
```diff
@@ -4,57 +4,32 @@
...
-        if not os.access(os.path.dirname(value), os.W_OK):
-            raise config_option.BadValue('You do not have write permission')
-        return value
-
-    def apply(self):
-        if not os.path.exists(self._config_file_value):
-            # make the directory
-            os.mkdir(self._config_file_value)
+
+if 'DIGITS_MODE_TEST' in os.environ:
```
gheinrich: maybe move this to a function like you are doing for other options like `log_file`?

digits/frameworks/caffe_framework.py
```diff
@@ -34,13 +34,13 @@ class CaffeFramework(Framework):
     # whether this framework can shuffle data during training
     CAN_SHUFFLE_DATA = False
 
-    if config_value('caffe_root')['flavor'] == 'NVIDIA':
-        if config_value('caffe_root')['version'] > parse_version('0.14.0-alpha'):
+    if config_value('caffe')['flavor'] == 'NVIDIA':
+        if parse_version(config_value('caffe')['version']) > parse_version('0.14.0-alpha'):
```
gheinrich: tested this path - ok

digits/config/caffe.py
```diff
@@ -0,0 +1,236 @@
...
+    match = re.match(r'%s(.*)\.so\.(\S+)$' % (libname), filename)
+    if match:
+        return match.group(2)
+    else:
+        return None
+
+
+
+if 'CAFFE_ROOT' in os.environ:
+    executable, version, flavor = load_from_envvar('CAFFE_ROOT')
```
gheinrich: tested this path - OK

digits/config/torch.py
```diff
@@ -0,0 +1,56 @@
...
+    if executable is None:
+        raise ValueError('Torch executable not found at "%s" (TORCH_ROOT)'
+                         % os.environ['TORCH_ROOT'])
+elif 'TORCH_HOME' in os.environ:
+    executable = find_executable(os.environ['TORCH_HOME'])
+    if executable is None:
+        raise ValueError('Torch executable not found at "%s" (TORCH_HOME)'
+                         % os.environ['TORCH_HOME'])
+else:
+    executable = find_executable()
```
gheinrich: tested this path - OK

digits/config/jobs_dir.py
lukeyeager: Do you really think that's more legible? I thought I'd just keep it simple when I can.

lukeyeager:
> Can we document the full list of environment variables?

I added a doc at `docs/Configuration.md`.

digits/config/jobs_dir.py
gheinrich: No I think it's fine this way, forget that comment.

lukeyeager:
Thanks for the review Greg! I'm going to merge this now, to enable #1093 (which enables #927).

lukeyeager:
Oops, I meant to squash before merging. Sorry.

PR #1093, opened 2016-09-19
Add a setup.py script

lukeyeager:
*Merge #1091 first*

TODO:

- [x] `digits-server` doesn't work because it can't find `gunicorn_config.py`
- [x] Nothing actually works because the `tools/` directory isn't copied
- [ ] ~~Try uploading to PyPI~~
  * Problems with requirements - will solve later
- [x] Try simplifying the debian packaging scripts
  * Works great!
- [ ] ~~Stop copying jobs directory around on pip install~~
  * It still technically works - so we're good to go with packaging
  * For local development, just use `pip install -e .`


lukeyeager:
@gheinrich this seems to be working now. Will you take a look and/or test it with your plugin work?

lukeyeager:
Justification for removing `requirements{,_test}.txt`:

I wanted to leave the requirements files alone, and read them in `setup.py` like this:
```py
requirements = []
with open(os.path.join(LOCAL_DIR, 'requirements.txt'), 'r') as infile:
    for line in infile:
        line = line.strip()
        if line and not line[0] == '#':  # ignore comments
            requirements.append(line)

setuptools.setup(
    # ...
    install_requires=requirements,
)
```
That worked fine for local installs, but failed when trying to download from PyPI:
```
  File "<string>", line 1, in <module>
  File "/tmp/pip-build-m3gKX_/digits/setup.py", line 13, in <module>
    with open(os.path.join(LOCAL_DIR, 'requirements.txt'), 'r') as infile:
IOError: [Errno 2] No such file or directory: '/tmp/pip-build-m3gKX_/digits/requirements.txt'
```
Apparently, pip has to re-calculate the requirements based on `setup.py` at install time, and `requirements.txt` isn't present at that time. As far as I can tell, there's no way to fix this. So I had to move the requirements specs into `setup.py`.

This was the most helpful resource I found: https://caremad.io/posts/2013/07/setup-vs-requirement/

gheinrich:
When I do `pip install .` after a while the tool exhausts my disk space and I get spates of errors like `('/home/greg/ws/digits/digits/jobs/20160907-011804-d09c', '/tmp/pip-KcNJhU-build/digits/jobs/20160907-011804-d09c', "[Errno 28] No space left on device: '/tmp/pip-KcNJhU-build/digits/jobs/20160907-011804-d09c'")`. Is it trying to copy the `jobs` directory?

`pip install -e .` is working though and I can add plugins.

lukeyeager:
> Is it trying to copy the jobs directory?

Oh yeah, I forgot about that issue. I gave up on it and just removed all the jobs and the data and the install seems to work fine. I'm not sure this is fix-able. I'll try to dig into it more...

gheinrich:
That looks OK to me!

PR #1106, opened 2016-09-21
WebGL Image Segmentation Visualization

jmancewicz:
Adding a WebGL image segmentation visualization. The current visualization shows the color coded segmentation image.

screen shot 2016-09-21 at 11 33 38 am

The new visualization overlays the segmentation image over input image.

screen shot 2016-09-21 at 11 35 41 am

The options menu has
- Opacity (of the detected pixels)
- Mask (the darkness of the non-detected pixels. May help highlight small segments.)
- Line Width (of the perimeter)

screen shot 2016-09-21 at 11 37 13 am

Moving the pointer over an image magnifies the area under the pointer.

screen shot 2016-09-21 at 11 43 09 am

I will be moving this from imageSegmentationGL to imageSegmentation soon. I'l be commenting and may be moving code around.


gheinrich:
Excellent! Let me try this on other datasets.

jmancewicz:
I've come across an issue with the 16 context limit. If there are more than 16 images, only the last 16 will draw the data over top.

I thought it automatically swapped contexts to the drop the those that haven't been used lately. The request for a context is supposed to only deliver if the canvas in on screen.

I'll look into this.

gheinrich:
Hi Joe on Ubuntu/Firefox I get an error "unable to initialize webGL". I see this in my console:
```
Error in parsing value for 'background-image'.  Declaration dropped. infer_one:1:7413
TypeError: gl is null
Stack trace:
GL.prototype.resize@http://localhost:5000/static/js/gl/gl-utils.js?ver=f463cc03e590bf141e8e598d6929c6fa:126:13
GL.prototype.run_webgl@http://localhost:5000/static/js/gl/gl-utils.js?ver=f463cc03e590bf141e8e598d6929c6fa:94:9
GL.prototype.load_shader_file/<@http://localhost:5000/static/js/gl/gl-utils.js?ver=f463cc03e590bf141e8e598d6929c6fa:280:17
 gl-utils.js:129:13
Error: WebGL: Error during native OpenGL init. gl-utils.js:124:40
Error: WebGL: WebGL creation failed. gl-utils.js:124:40
TypeError: gl is null
 gl-utils.js:219:9
TypeError: this.sink is null network-monitor.js:275:5
Use of getPreventDefault() is deprecated.  Use defaultPrevented instead.
```

gheinrich:
Can we show the legend closer to the overlay? It's difficult to see class names if we need to scroll up and down.

jmancewicz:
I agree.  I'm considering a legend with only the relevant classes under each image.

More like this.

screen shot 2016-09-21 at 1 02 55 pm

gheinrich:
With the ability to click on the class in the legend to set the opacity to 0 for that class? That would be totally awesome!

jmancewicz:
That would have to be a new story.

jmancewicz:
So, the issue is going to be that I'm using one GL context per image, and I run out of contexts. I need to manage the contexts by either moving swapping them between visible images or using a single scrollable context in which all items are shown. The single allows more tabs to be opened, the swapping may be simpler to implement and prevent reinventing interactivity.

gheinrich:
> That would have to be a new story.

Sure. What we have now is already great. 

Do you know why WebGL is [not working for me](https://github.com/NVIDIA/DIGITS/pull/1106#issuecomment-248722831)?

digits/model/views.py
```diff
@@ -96,6 +95,21 @@ def customize():
             'snapshot': snapshot
             })
 
+@blueprint.route('/extension-static/<extension_id>/<path:filename>')
```
gheinrich: does it have to be in the **model** views? Can you move this one level up maybe?

jmancewicz:
@gheinrich, With respect to the 'unable to initialize webGL' error that you received. I ran it on my Ubuntu 14.04 without a problem. Were you running this on the machine itself, or were you using NX or VNC or something to view it remotely? Those don't always deal with GL very well.

digits/model/views.py
jmancewicz: There can be a data and a view extension with the same name, so it needs to be disambiguated. 

You are right that it should move to digits/views.py, but I'll need to add a view/data segment to the path.

gheinrich:
> Were you running this on the machine itself, or were you using NX or VNC or something to view it remotely?

I was running locally on the machine (Ubuntu 14.04 Firefox 45.0.1).

jmancewicz:
This has been reworked to avoid running out of gl contexts. A considerable amount of it a framework for making other webgl based extensions.  It's still not as performant as I'd like, and will look into that in the near future.

It's been tested on
OSX
- Google Chrome 53.0.2785.116 (64-bit)
- Firefox 49.0.1
- Opera 40.0.2308.75
- Vivaldi 1.1.453.52

Windows
- Firefox 49.0.1

Ubuntu
- Firefox 47.0
- Chrome 51.0.2704.103 (64-bit)

jmancewicz:
> I was running locally on the machine (Ubuntu 14.04 Firefox 45.0.1)

I've tested with Firefox 47.0 on Ubuntu 14.04 with no issue. I've not installed anything to enable webgl. Can you try this commit, @gheinrich ?

jmancewicz:
> Do you know why WebGL is not working for me?

@gheinrich, can you check your firefox preferences. Options -> Advanced -> General -> Use hardware acceleration when available

screen shot 2016-09-28 at 6 12 04 pm

I've read that some people find disabling hardware acceleration allows WebGL to be used in Firefox. It may just be anecdotal but worth a try.

digits/extensions/view/imageSegmentationGL/view.py
```diff
@@ -0,0 +1,190 @@
...
+            legend.append({'image': image, 'text': text})
+        return template, {'legend': legend}
+
+    @override
+    def get_ng_templates(self):
+        """
+        Implements get_ng_templates() method from view extension interface
+        """
+        extension_dir = os.path.dirname(os.path.abspath(__file__))
+        header = open(os.path.join(extension_dir, APP_BEGIN_TEMPLATE), "r").read()
```
jmancewicz: To obviate the need for the extensions route?

digits/model/views.py
jmancewicz: moved up one level.

digits/extensions/view/imageSegmentationGL/view.py
gheinrich: yes, because if the extension comes from a 3rd party plug-in (outside of DIGITS project) we don't know how to construct the path to the resource files.

gheinrich:
Hi Joe, I am not seeing the WebGL anymore (with/without hardware acceleration) on the last commit. Thanks!

digits/views.py
```diff
@@ -653,6 +653,29 @@ def path_autocomplete():
...
 
+@blueprint.route('/extension-static/<extension_type>/<extension_id>/<path:filename>')
+def extension_static(extension_type, extension_id, filename):
+    """
+    Returns static files from an extension's static directory.
+    '/extension-static/view/image-segmentation-gl/js/gl.js'
+    would send the file
+    'digits/extensions/view/imageSegmentationGL/static/js/gl.js'
+    """
+    print extension_type, extension_id
```
gheinrich: just to remind you to remove this in final version

lukeyeager:
@jmancewicz are we closing this in favor of https://github.com/NVIDIA/DIGITS/pull/1131?

jmancewicz:
Yes

PR #1110, opened 2016-09-22
Small changes to the "About" dropdown

lukeyeager:
* Why is `class=brightness` there for just one element?
* I feel like "DIGITS on GitHub" and "DIGITS docs" are nearly redundant
* Capitalization changes

/cc @jmancewicz


jmancewicz:
Yeah, I was torn about having both "DIGITS on GitHub" and "DIGITS docs".

Looks all good.

PR #1113, opened 2016-09-23
Deb packaging

lukeyeager:
Borrowed many things from [nvidia-docker](https://github.com/NVIDIA/nvidia-docker/tree/v1.0.0-rc.3/tools) (thanks @3XX0!).


deb/build.sh
```diff
@@ -0,0 +1,67 @@
...
+if [ $(git tag -l $GIT_TAG | wc -l) -ne 1 ]; then
+    echo "$GIT_TAG is not a git tag"
+    exit 1
+fi
+DESCRIBE_VERSION=$(git describe --match $GIT_TAG)
+SOURCE_VERSION=${DESCRIBE_VERSION:1}
+UPSTREAM_VERSION=$(echo $SOURCE_VERSION | sed '0,/-/{s/-/~/}' | sed 's/-/\./g')
+echo UPSTREAM_VERSION: $UPSTREAM_VERSION
+
+GIT_REV=$(git stash create)
```
flx42: Maybe you should just abort if the environment is dirty.

deb/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM ubuntu:14.04
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        build-essential \
+        dh-make \
+        devscripts \
+        equivs \
+        lintian
```
flx42: It's recommend to clean the apt cache afterwards: https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/apt-get

deb/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+ENV DEBFULLNAME "DIGITS Development Team"
+ENV DEBEMAIL "digits@nvidia.com"
+ENV DEB_BUILD_OPTIONS nocheck
+
+ARG UPSTREAM_VERSION
+ARG DEBIAN_VERSION
+
+WORKDIR /build
+COPY tarball/* .
```
flx42: I don't think you need the `/*` here

deb/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+    -f ../*.orig.tar.gz -p digits_${UPSTREAM_VERSION}
+RUN dch --create --package digits -v $DEBIAN_VERSION "v${DEBIAN_VERSION}"
+RUN dch -r ""
+RUN cp -R deb/extras/* debian/
+RUN apt-get update && echo y | mk-build-deps -i -r debian/control
+RUN debuild --no-lintian -i -uc -us -b
+RUN debuild --no-lintian -i -uc -us -S -sa
+RUN lintian ../*.changes
+RUN mkdir -p /dist
+RUN cp ../* /dist/ || true
```
flx42: No need to create so many layers, you should have fewer `RUN` commands.

deb/build.sh
```diff
@@ -0,0 +1,67 @@
...
+    exit 1
+fi
+DOCKER_ID="digits-debbuild"
+cd $LOCAL_DIR
+docker build -t $DOCKER_ID \
+    --build-arg UPSTREAM_VERSION=$UPSTREAM_VERSION \
+    --build-arg DEBIAN_VERSION=$DEBIAN_VERSION \
+    .
+
+set +e
```
flx42: Why?

deb/build.sh
lukeyeager: Yeah good call.

deb/Dockerfile
lukeyeager: Ugh. It makes my build time slower. Ok.

deb/Dockerfile
lukeyeager: I don't think this really matters if I'm not actually distributing the docker image. But ok.

deb/build.sh
lukeyeager: Yeah this is dumb. Changing ...

PR #1120, opened 2016-09-26
Move the download_data scripts into digits/

lukeyeager:
* Data download scripts get distributed automatically with setup.py
* Allows "python -m digits.download_data" - independent of installation method


lukeyeager:
(needed for upcoming v4.1 documentation update)

PR #1127, opened 2016-09-28
Add packaging support for 16.04 (and remove gunicorn)

lukeyeager:
*Continuation of https://github.com/NVIDIA/DIGITS/pull/1113*

As explained in the commit messages, gunicorn isn't doing any good for us at present.

Since https://github.com/NVIDIA/DIGITS/pull/1121, `digits-devserver` isn't really required. Because we're already removing `digits-server` in this PR, I think it makes sense to remove both at once.


IsaacYangSLA:
Just did a sanity check and it worked without issue.
Details:
install DIGITS 4.0 via apt-get install
clean git clone of DIGITS
master branch merged with PR1127
git checkout pr1127
run build.sh
sudo dpkg -i DIGITS_PACKAGE.deb

lukeyeager:
@gheinrich any concerns with this?

docs/BuildDigits.md
```diff
@@ -49,39 +49,10 @@ sudo pip install -e $DIGITS_HOME
...
-
-DIGITS development server
-
-optional arguments:
-  -h, --help            show this help message and exit
-  -p PORT, --port PORT  Port to run app on (default 5000)
-  -d, --debug           Run the application in debug mode (reloads when the
-                        source changes and gives more detailed error messages)
-  --version             Print the version number and exit
+python -m digits
```
lukeyeager: 1. Technically, it is still optional. You can run `python -m digits` without `pip install digits` as long as you're in `$DIGITS_ROOT`.

2. I was having problems with that, but I think it's probably fixed by https://github.com/NVIDIA/DIGITS/pull/1124/files. Let me check again...

lukeyeager: Actually, I think I'll leave `digits-devserver` in there for now and keep telling people to use it. Next step can be:

* PyPI support
* Console entry point (it does work now)

PR #1130, opened 2016-09-30
Bugfix - Torch and CUDA_VISIBLE_DEVICES

lukeyeager:
I think this has been broken forever. But it's a bigger deal now that https://github.com/NVIDIA/DIGITS/pull/1091 encourages users to use `CUDA_VISIBLE_DEVICES` to choose the GPUs that will be used in DIGITS.

To test:
1. `export CUDA_VISIBLE_DEVICES=1,0`
2. `./digits-devserver`
3. Train a model with torch
4. You'll see 0 utilization for the selected GPU. `nvidia-smi` will tell you that Torch is using the other GPU.

/cc @gheinrich

### DIFF
--- a/digits/model/tasks/torch_train.py
+++ b/digits/model/tasks/torch_train.py
@@ -29,6 +29,28 @@ PICKLE_VERSION = 1
 TORCH_MODEL_FILE = 'model.lua'
 TORCH_SNAPSHOT_PREFIX = 'snapshot'
 
+
+def subprocess_visible_devices(gpus):
+    """
+    Calculates CUDA_VISIBLE_DEVICES for a subprocess
+    """
+    if not isinstance(gpus, list):
+        raise ValueError('gpus should be a list')
+    gpus = [int(g) for g in gpus]
+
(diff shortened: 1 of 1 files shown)


digits/model/tasks/torch_train.py
```diff
@@ -29,6 +29,29 @@
...
+    gpus = [int(g) for g in gpus]
+
+    old_cvd = os.environ.get('CUDA_VISIBLE_DEVICES', None)
+    if old_cvd is None:
+        real_gpus = gpus
+    else:
+        map_visible_to_real = {}
+        for visible, real in enumerate(old_cvd.split(',')):
+            map_visible_to_real[visible] = int(real)
+        print map_visible_to_real
```
gheinrich: that little `print` was forgotten here :-)

lukeyeager: Oops, thanks for the catch!

lukeyeager:
I can't get Travis to pass because of https://github.com/NVIDIA/DIGITS/issues/1134.

PR #1131, opened 2016-10-01
Image segmentation view improvement

jmancewicz:
Changed the imageSegmentation view extension so that rather than displaying the segmentation image,
screen shot 2016-10-01 at 10 21 28 am

The segmentation image overlays the input image with a solid outline around each class.
screen shot 2016-10-01 at 10 21 07 am

The opacity of the segmented regions and the opacity of the mask of the non-segmented regions are adjustable.
screen shot 2016-10-01 at 10 18 27 am
screen shot 2016-10-01 at 10 18 45 am

I've removed the color to class legend from the header and put a legend or relevant classes beneath each image. When the pointer is over a segmented region, the corresponding class in the legend is highlighted.

screen shot 2016-10-01 at 10 28 28 am


digits/extensions/view/imageSegmentation/view.py
```diff
@@ -6,6 +6,8 @@
 import matplotlib as mpl
 import numpy as np
 import PIL.Image
+import json
+import skfmm
```
gheinrich: @lukeyeager usually prefers seeing non-standard package imports below (line 14)

digits/extensions/view/imageSegmentation/view.py
```diff
@@ -90,24 +120,20 @@ def get_header_template(self):
...
-                image[:,:] = rgb_color
-            else:
-                image = np.full(shape=(50, 50), fill_value=c)
-            image = image.astype('uint8')
-            image = digits.utils.image.embed_image_html(image)
-            text = "Class #%d" % c
-            if self.class_labels:
-                text = "%s (%s)" % (text, self.class_labels[int(c)])
-            legend.append({'image': image, 'text': text})
+        legend = self.get_legend_for(self.found_classes)
```
gheinrich: I think we're not using the legend anymore in the header template, correct?

digits/extensions/view/imageSegmentation/view.py
```diff
@@ -134,24 +171,68 @@ def process_data(self, input_id, input_data, output_data):
...
         # of classes, H and W are the height and width of the image
-        data = output_data[output_data.keys()[0]].astype('float32')
+        class_data = output_data[output_data.keys()[0]].astype('float32')
         # retain only the top class for each pixel
-        data = np.argmax(data,axis=0)
+        class_data = np.argmax(class_data, axis=0).astype('uint8')
 
         # remember the classes we found
-        found_classes = np.unique(data)
+        found_classes = np.unique(class_data)
```
gheinrich: do we still need `self.found_classes` below? Now that we have one legend per image I think this can go away.

digits/extensions/view/imageSegmentation/view.py
```diff
@@ -134,24 +171,68 @@ def process_data(self, input_id, input_data, output_data):
...
-        return image
+            fill_data = (self.map.to_rgba(class_data)*255).astype('uint8')
+        else:
+            fill_data = np.ndarray((class_data.shape[0], class_data.shape[1], 4), dtype='uint8')
+            for x in xrange(3):
+                fill_data[:, :, x] = class_data.copy()
+
+        zeros = np.zeros(class_data.shape)
+        # Assuming that class 0 is the background
+        mask = np.greater(class_data, zeros)
```
gheinrich: do you have to create the `zeros` array? I think you can use `np.greater` to compare against a scalar:
```
>>> x=np.array([0,1,2,0,4])
>>> mask=np.greater(x,0)
>>> mask
array([False,  True,  True, False,  True], dtype=bool)
```

digits/extensions/view/imageSegmentation/view.py
```diff
@@ -134,24 +171,68 @@ def process_data(self, input_id, input_data, output_data):
...
+        fill_data[:, :, 3] = mask * 255
+
+        # Black mask of non-segmented pixels
+        mask_data = np.zeros(fill_data.shape, dtype='uint8')
```
gheinrich: do you need this line? I think the below one is enough (?)

digits/extensions/view/imageSegmentation/view.py
```diff
@@ -134,24 +171,68 @@ def process_data(self, input_id, input_data, output_data):
...
+        mask_data[:, :, 3] = (1 - mask) * 255
+
+        # Generate outlines around segmented classes
+        if len(found_classes) > 1:
+            # Assuming that class 0 is the background.
+            line_mask = np.zeros(class_data.shape, dtype=bool)
+            for c in (x for x in found_classes if x != 0):
+                c_mask = np.equal(class_data, zeros + c)
```
gheinrich: I think you can also do without `zeros` here:
```
>>> x=np.array([0,1,1,2])
>>> np.equal(x,1)
array([False,  True,  True, False], dtype=bool)
```

digits/extensions/view/imageSegmentation/view.py
```diff
@@ -134,24 +171,68 @@ def process_data(self, input_id, input_data, output_data):
...
+                # Find the signed distance from the zero contour
+                distance = skfmm.distance(c_mask.astype('float32') - 0.5)
+                # Accumulate the mask for all classes
+                line_mask |= c_mask & np.less(distance, zeros + 3)
```
gheinrich: neat!

digits/inference/job.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, model, images, epoch, layers, resize=True, **kwargs):
 
     @override
     def __getstate__(self):
-        fields_to_save = ['_id', '_name']
+        fields_to_save = ['_id', '_name', 'status_history']
```
gheinrich: can you explain this change?

digits/extensions/view/imageSegmentation/view.py
jmancewicz: Correct. Great catch.

digits/extensions/view/imageSegmentation/view.py
jmancewicz: Damn, you're good.

digits/extensions/view/imageSegmentation/view.py
jmancewicz: Nice!

digits/extensions/view/imageSegmentation/view.py
jmancewicz: According to PEP8, I need to add a blank line. (https://www.python.org/dev/peps/pep-0008/#imports). Are you saying @lukeyeager prefers third party imports between application and local imports?

digits/extensions/view/imageSegmentation/view.py
jmancewicz: The np.zeros create RGBA data. the line below sets the A component. Is there some pythonic trickery that I'm missing?

digits/extensions/view/imageSegmentation/view.py
jmancewicz: Done.

digits/inference/job.py
jmancewicz: Not for this commit. I needed for something a while back, which I no longer recall. Removing.

digits/extensions/view/imageSegmentation/view.py
gheinrich: my mistake, sorry.

digits/extensions/view/imageSegmentation/view.py
```diff
@@ -134,24 +165,66 @@ def process_data(self, input_id, input_data, output_data):
...
+                c_mask = np.equal(class_data, c)
+                # Find the signed distance from the zero contour
+                distance = skfmm.distance(c_mask.astype('float32') - 0.5)
+                # Accumulate the mask for all classes
+                line_width = 3
+                line_mask |= c_mask & np.less(distance, line_width)
+
+        # add the outlines to the input image
+        for x in xrange(3):
+            input_data[:, :, x] = (input_data[:, :, x] * (1 - line_mask) +
```
gheinrich: this doesn't work if `input_data` is a grayscale image (with shape e.g. `[256,256]`)

gheinrich: also, this should be done within the `if len(found_classes)>1` otherwise `line_mask` is uninitialized

digits/extensions/view/imageSegmentation/view.py
```diff
@@ -134,24 +165,66 @@ def process_data(self, input_id, input_data, output_data):
...
+                                   fill_data[:, :, x] * line_mask)
+
+        # Input image with outlines
+        input_image = PIL.Image.fromarray(input_data)
```
gheinrich: does this work if `input_data.dtype!='uint8'`?

digits/extensions/view/imageSegmentation/view.py
jmancewicz: @gheinrich, did you try this with a (256, 256)?

gheinrich: @jmancewicz yes, this was an image from the Sunnybrook dataset (with channel conversion = 'none')

digits/extensions/view/imageSegmentation/view.py
jmancewicz: No, but that's in an upcoming PR.

PR #1136, opened 2016-10-03
Supporting DICOM via pydicom

IsaacYangSLA:
1. Importing dicom if dicom module can be loaded.
2. Storing dicom image in LMDB data field as float.
3. Several small modifications on dealing with PIL.Image mode
4. Notify users to use None on higher bit images.


digits/utils/image.py
```diff
@@ -11,6 +11,20 @@
 except ImportError:
     from StringIO import StringIO
 
+try:
```
gheinrich: don't you want to add `pydicom` to `requirements.txt`? If the package is not installed, DIGITS will silently refuse to read DICOM files, which could be confusing to the user.

digits/utils/image.py
```diff
@@ -35,6 +49,27 @@
...
+
+    Arguments:
+    path -- file system path to the image
+    """
+    try:
+        dm = dicom.read_file(path)
+    except dicom.errors.InvalidDicomError:
+        raise errors.LoadImageError, 'Invalid Dicom file'
+    pixels = dm.pixel_array
+    image = PIL.Image.fromarray(pixels.astype(np.float))
```
gheinrich: how do you deal with 3D/4D data (which seem to be the most common use of DICOM)?

digits/utils/image.py
IsaacYangSLA: Let me check if 0.9.7 works.  Pip installed 0.9.9 and it's the one I tested.

digits/tools/create_db.py
```diff
@@ -693,7 +725,10 @@ def _save_means(image_sum, image_count, mean_files):
             with open(mean_file, 'wb') as outfile:
                 outfile.write(blob.SerializeToString())
         elif mean_file.lower().endswith(('.jpg', '.jpeg', '.png')):
-            image = PIL.Image.fromarray(mean)
+            if np.issubdtype(mean.dtype, np.float):
+                image = PIL.Image.fromarray(mean*255/mean.max()).convert('L')
```
IsaacYangSLA: You are right.  I don't think any pixel in mean can be greater than 65535.  We should be able to do that for PNG case.

digits/tools/create_db.py
```diff
@@ -579,12 +602,15 @@ def _load_thread(load_queue, write_queue, summary_queue,
...
-        else:
-            write_queue.put((image, label))
-
+        try:
+            if backend == 'lmdb':
+                datum = _array_to_datum(image, label, encoding)
+                write_queue.put(datum)
+            else:
+                write_queue.put((image, label))
+        except IOError:  # try to save 16-bit images with PNG/JPG encoding
```
IsaacYangSLA: Let me catch the error message, too, and set _notification with that error message.

digits/tools/create_db.py
IsaacYangSLA: @gheinrich , it looks like if the max value is greater than 255, but much smaller than 65536 (for example, around 800), and we save it to png file.  That image is almost black.  The reason, I believe, is the dynamic range of 16-bit PNG is much higher than mean pixel value.  We can either scale it up to full dynamic range of 16-bit PNG, or scale it down to 8-bit png/jpg.  Both look the same on screen.

gheinrich: Indeed it's impossible to know how to best visualize data on server side. This is something we discussed with @jmancewicz. Joe thinks the best way to deal with this is to have the user choose the amount of contrast, etc. though sliders on client side. I agree with him.

digits/utils/image.py
IsaacYangSLA: I think we can support that in the future.  Adding 3D/4D data support from one single DICOM file would involve many modifications.  Maybe that's because my limited understanding on the current codes.

digits/utils/image.py
gheinrich: so you can add `pydicom` to `requirements.txt` and `import dicom` without a `try/except` then?

digits/utils/image.py
```diff
@@ -35,6 +49,27 @@
 # List of supported file extensions
 # Use like "if filename.endswith(SUPPORTED_EXTENSIONS)"
 SUPPORTED_EXTENSIONS = ('.png','.jpg','.jpeg','.bmp','.ppm')
+if dicom is not None:
```
gheinrich: if we always import `dicom` maybe you don't need this `if`?

digits/utils/image.py
gheinrich: Sure that can be a future improvement. If `pixels` is a 3-D array, how do you determine whether it's an RGB image (OK in DIGITS) or a volumetric grayscale image (not OK in DIGITS if more than there aren't exactly 3 slices)?

digits/utils/image.py
IsaacYangSLA: Sure.

digits/utils/image.py
IsaacYangSLA: Each slice should have its meta data.  Therefore, we should be able to tell if it's a multi-slice DICOM, or a multi-channel DICOM.  I need to dig out that information, but I remember I saw that before.

digits/templates/datasets/images/generic/summary.html
```diff
@@ -21,6 +21,7 @@
         <li><b>Image Count</b> - {{task.image_count}}</li>
         <li><b>Image Dimensions</b> -
         {{task.image_width}}x{{task.image_height}}x{{task.image_channels}}</li>
+        <li><b>Image bit-depth</b></li>
```
IsaacYangSLA: I added that because it's part of UI selection.  However, you are right about the usefulness.  Old dataset did not have this info and the summary showed 'Image bit-depth' but nothing below it.  Maybe it causes more confusing.

digits/utils/image.py
IsaacYangSLA: That's correct.  We can remove that ``if``` after we require ```pydicom``` package.

IsaacYangSLA:
If an inference task applies to a high bit-depth image a model which is trained with 8-bit images, it's likely to generate incorrect results.  If resize is needed for test image, the function will follow test image's bit-depth to preserve its dynamic range.

gheinrich:
When you create a classification dataset, all images are resized to the specified dimensions and converted to the specified color scheme (grayscale or RGB). It would feel awkward if this wasn't the case for the bit depth too. I understand it's probably more work than was originally planned but I think the bit depth should be enforced. Besides, doing this (both during dataset creation and inference) would avoid a lot of issues that could arise if a user mixes 8bbc and 18bbc images.

IsaacYangSLA:
@gheinrich, thanks a lot.  I think we are getting closer to finish it.  Let me explain it and see if it makes sense to you.
An dicom image, after being loaded, will have image mode 'F.'  And during resize, that 'F' is maintained, and no scaling is done on pixel values.  At the very last stage, if user selected bpp = 32, and unencoded for LMDB, DIGITS save the complete information.  The trained model remembers its dataset and its bpp.  At inferencing, only the combination of mode 'F' images and model trained with bpp=8 will raise exception.  If no exception, the resize stage of test images will apply original dataset's bpp (any image modes with bpp=32, or non-'F' modes with bpp=8).  So the model will see the test images processed by the same bpp.

digits/dataset/images/forms.py
```diff
@@ -44,3 +44,8 @@ class ImageDatasetForm(DatasetForm):
             choices=ImageDatasetJob.resize_mode_choices(),
             tooltip = "Options for dealing with aspect ratio changes during resize. See examples below."
             )
+    resize_bpp = utils.forms.SelectField(u'Bits per pixel',
+            default='8',
+            choices=[('8', '8-bit (color or grayscale)'), ('32', '32-bit floating point (grayscale only)')],
+            tooltip="Storing 32-bit floating point for certain medical images."
```
gheinrich: may be useful to have a more verbose tooltip to state the fact that bit depth is only checked, not enforced through conversion

gheinrich:
@IsaacYangSLA I think it makes sense indeed.

PR #1142, opened 2016-10-04
[TravisCI] build parallelism

lukeyeager:
Apparently we only have access to 2 processors on Travis - not 4.
I was hoping this would speed the Torch build up more - it only shaves off a minute or so.

/cc @flx42

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ python:
     2.7
 compiler: gcc
 env:
-    global:
-        - NUM_THREADS=4
     matrix:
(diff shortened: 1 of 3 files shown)


flx42:
Slightly overcommitting the number of threads compared to the number of cores is usually harmless: it can be slightly faster or slower, it depends.
But at least now you have the potential of being much faster if you end up on an instance with more cores.

flx42:
Oh, and for torch, many small projects are installed sequentially, even if the project itself is built in parallel, it's not going to help that much.

PR #1143, opened 2016-10-05
Separate Travis builds for caffe/torch/deb

lukeyeager:
digits-travis-debbuild

* Updated the test suite to check for the envvar `DIGITS_TEST_FRAMEWORK` (caffe, torch or none)
  * This could be useful for every-day testing - not just on Eris
  * The hacks I had to add for dealing with inheritance properly are pretty gnarly
* Test deb package build on Travis (preparing for automatic PPA uploads)


lukeyeager:
One big positive of this approach is that it's clear when the tests are failing because of Torch (which is usually the case).

    [ERROR] torch classify many task failed with error code -11

Looks like Torch is segfaulting!? That's a new one.

lukeyeager:
> Looks like Torch is segfaulting!? That's a new one.

Re-ran the build and it passed.

lukeyeager:
Thanks for the feedback Joe. I've refactored some code into a new file: `digits/test_utils.py`.

lukeyeager:
When the builds are cached everything finishes in 10 minutes now! :tada: 
https://travis-ci.org/NVIDIA/DIGITS/builds/165384804

Issue #1153, opened 2016-10-11
Make coding style compatible with flake8


lukeyeager:
Also add flake8 to TravisCI so this is enforced

lukeyeager:
For motivation, a quick look through the output of pyflakes discovered some real bugs (see https://github.com/NVIDIA/DIGITS/pull/1154). Staying lint-free would likely help avoid some more bugs in the future.

PR #1154, opened 2016-10-11
Fix errors found with pyflakes

lukeyeager:
While investigating https://github.com/NVIDIA/DIGITS/issues/1153, pyflakes pointed out a few real errors (not just style issues).


lukeyeager:
I had originally started fixing everything from pyflakes. But I stopped because (1) there are a TON of things to fix and (2) we should have a discussion about this first - if we fix all the errors we'll touch all the files and likely make merge conflicts with all the open pull requests.

lukeyeager:
@gheinrich will you take a quick look? Most of the errors are mine but a few are yours.

PR #1156, opened 2016-10-12
Including Python layer file during importing models

IsaacYangSLA:
Download python layer file if specified in info.json
Pretrained model copies python layer file
Automatically fill server-side python layer file name when users click 'Customize'


digits/pretrained_model/job.py
```diff
@@ -47,6 +47,16 @@ def get_weights_path(self):
     def get_model_def_path(self):
         return self.tasks[0].get_model_def_path()
 
+    def get_python_layer_path(self):
+        tmp_dir = os.path.dirname(self.tasks[0].get_model_def_path())
+        python_layer_file_name = 'digits_python_layers.py'
```
lukeyeager: Don't we have a constant for this filename somewhere?

lukeyeager: Ok. We can fix it later I guess.

IsaacYangSLA: Yes.  It can be done when we spend sometime refactoring my messy codes.

IsaacYangSLA:
I will merge it after Model Store UI improvement is merged.

IsaacYangSLA:
Merged manually with commandline.

PR #1159, opened 2016-10-12
Partial python lint checking with flake8

lukeyeager:
To avoid being too invasive, for now it only checks for lint in `setup.py` and under `examples/`.

But it puts in place all the automation needed to check Python lint throughout the project (https://github.com/NVIDIA/DIGITS/issues/1153).

Example lint failure: https://travis-ci.org/lukeyeager/DIGITS/builds/167121132


gheinrich:
Looks great! Can the lint checker be another target of the test matrix? Something like this so we can easily spot Lint errors:
```
env:
    matrix:
        - DIGITS_TEST_FRAMEWORK=caffe CAFFE_FORK=NVIDIA
        - DIGITS_TEST_FRAMEWORK=caffe CAFFE_FORK=BVLC
        - DIGITS_TEST_FRAMEWORK=torch
        - DIGITS_TEST_FRAMEWORK=none
        - DEB_BUILD=true
        - LINT_CHECKER=true
```

lukeyeager:
Excellent suggestion, greg!

* Created a new test target
* Moved it to the top so you get quick feedback
* Refactored apt package installation out of `.travis.yml` so we don't install unnecessary packages for simple test targets

Updated example lint failure: https://travis-ci.org/lukeyeager/DIGITS/builds/169032940

lukeyeager:
Refactored travis code again, borrowing [ideas from ethereum](https://github.com/ethereum/go-ethereum/blob/v1.4.18/.travis.yml). This moves `LINT_CHECK` back down to the bottom, but it sets the weird tests (lint and debuild) apart nicely and tees us up for Python3 testing, if we ever get there.

TimZaman:
This is great luke. Can we still do in-line overrides?

.travis.yml
```diff
@@ -1,24 +1,51 @@
...
+        - OMP_NUM_THREADS=1
+        - OPENBLAS_MAIN_FREE=1
     matrix:
         - DIGITS_TEST_FRAMEWORK=caffe CAFFE_FORK=NVIDIA
         - DIGITS_TEST_FRAMEWORK=caffe CAFFE_FORK=BVLC
         - DIGITS_TEST_FRAMEWORK=torch
         - DIGITS_TEST_FRAMEWORK=none
-        - DEB_BUILD=true
+
+matrix:
```
gheinrich: I don't really understand how this second definition of `matrix` works in conjunction with the first one. Should the one on line `16` not start with an `include` directive and be merged with this one on line `22`?

setup.cfg
```diff
@@ -0,0 +1,3 @@
+[flake8]
+max-line-length = 120
+exclude = venv/,digits/,plugins/
```
gheinrich: If `LINT_CHECK` is added to an `allow_failures` section maybe we can include everything?

gheinrich:
> This is great luke. Can we still do in-line overrides?

@TimZaman I think you do this by appending `# noqa` to a line

lukeyeager:
> I think you do this by appending # noqa to a line

That's right. But you can also ignore a specific rule on a line (`# noqa: E234`) or ignore a specific rule for an entire file ([example](https://github.com/lukeyeager/flask-sqlalchemy-socketio-demo/blob/7fe63b2604/myapp/comm/__init__.py#L1)).

.travis.yml
lukeyeager: It's a little funky. We create a matrix with `env: matrix` that is our standard build matrix. If we were to change `python: 2.7` to `python: 2.7, 3.5` then we would have 8 standard build targets (2x4). Then, after this standard matrix is constructed, we can "include" or "exclude" some specific targets. I'm using it to include `LINT_CHECK` and `DEB_BUILD`, but we could also exclude "python3 and FRAMEWORK=tensorflow" or something else that was part of the standard matrix.

setup.cfg
lukeyeager: That's worth considering. I think I'll just merge this as-is, but consider moving to that approach on the next incremental lint-fixing PR.

PR #1160, opened 2016-10-12
Luarock installation

lukeyeager:
Minor changes to torch installation instructions.

/cc @3XX0, @gheinrich


lukeyeager:
I wanted to install `lua-pb` like normal - just `luarocks install lua-pb`. But it's only listed as a "dev" package on luarocks.org:

https://luarocks.org/modules/akhaustov/lua-pb

And it also hasn't been updated in almost a year, whereas the GitHub repo was updated about 8 months ago.

So, we'll have to stick with the GitHub URL for this one.

lukeyeager:
I verified that the TravisCI build passes in my fork even after clearing all caches and forcing a new build.

PR #1161, opened 2016-10-12
Modelstore ui

jmancewicz:
Making the Model Store page more consistent with the DIGITS front page and adding a download progress bar.

Moving from this look;
old

To this look;
new

I also stopped the redirection to the front page after download was complete.

The progress bar is a progress line at the moment, under the model name to save real estate.

### DIFF
--- a/digits/templates/store.html
+++ b/digits/templates/store.html
@@ -12,116 +12,153 @@
 <script type="text/javascript" src="{{ url_for('static', filename='js/jquery.sparkline.min.js', ver=dir_hash) }}"></script>
 <script type="text/javascript" src="{{ url_for('static', filename='js/store.js', ver=dir_hash) }}"></script>
 <script>
-    $(document).ready(function(){
-        $('[data-toggle="tooltip"]').tooltip();
-    });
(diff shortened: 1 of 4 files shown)


jmancewicz:
@lukeyeager and @IsaacYangSLA maybe have a look.

digits/templates/store.html
```diff
@@ -12,116 +12,153 @@
...
-                        <th>License</th>
-                    </tr>
-                    </thead>
-                    <tbody>
-                    <tr ng-repeat="model in value.model_list | filter: query">
-                        <td>
-                            <a href="/store/push?id={[model.id]}"> Import </a>
-                        </td>
-                        <td>
+        </div>
```
IsaacYangSLA: I also have 'd' around here.  However, the original HTML does not have that text.

IsaacYangSLA: Yes.  It's not there any more.  It should be OK to merge.
@lukeyeager how do you think?

IsaacYangSLA:
I will merge it since it looks great and I didn't see anything broken.

PR #1163, opened 2016-10-13
Fix the image segmentation view line_mask bug

jmancewicz:
A few lines need to be indented to the correct block


jmancewicz:
I have this fixed on a branch that I needed to pause, but this needs to go out.

PR #1167, opened 2016-10-14
Fixes for segmentation extensions

lukeyeager:


IsaacYangSLA:
Tested on a few images from Sunnybrook validation set.  All of them worked well.

PR #1188, opened 2016-10-19
Additional image segmentation view options

jmancewicz:
Adding view options for image segmentation. 

screen shot 2016-10-19 at 12 20 01 pm

The existing Opacity slider changes the opacity of the fill region.
screen shot 2016-10-19 at 12 21 43 pm

The Mask slider changes the opacity of the non-segmented area. I find this useful to find small segments.
screen shot 2016-10-19 at 12 22 00 pm

The Threshold slider changes the moves in the distance field around segmentation boundaries.
screen shot 2016-10-19 at 12 22 18 pm

The Line Width does what you think
screen shot 2016-10-19 at 12 22 32 pm

Gamma along with the Colormap can make the grey scale medical images a little easier to read.
screen shot 2016-10-19 at 12 23 16 pm

Also, if the inference feature image comes through as a float array, and PIL won't convert it to an image, then the data is normalized to its range and converted to uint8 so that it will be converted to a PIL.Image.

### DIFF
--- a/digits/extensions/view/imageSegmentation/app_begin_template.html
+++ b/digits/extensions/view/imageSegmentation/app_begin_template.html
@@ -2,6 +2,64 @@
 
 <div ng-app="visualization_app" class="ng-cloak">
     <div ng-controller="display_controller as dispaly">
+        <?xml version="1.0"?>
+        <!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
+        "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+        <svg xmlns="http://www.w3.org/2000/svg" version="1.1" height="0px">
+            <defs>
+                <filter id="Gamma">
+                    <feComponentTransfer>
+                        <feFuncR type="gamma" amplitude="1" exponent="{[storage.gamma]}" offset="0"/>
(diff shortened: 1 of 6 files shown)


gheinrich:
Looks great! Do you think we could also apply this to the [binary segmentation example](https://github.com/NVIDIA/DIGITS/tree/digits-5.0/examples/binary-segmentation)? In this example the network output has shape `Height x Width` and each "pixel" value is a score (0->strongly think it's foreground, 255->strongly think it's background - we could reverse this if it's easier to visualize).

jmancewicz:
I'm sure we can get it to work with the binary example. I'll have a look now.

jmancewicz:
@gheinrich, I just went through the binary segmentation example, and cam up with a result similar to what you show.
screen shot 2016-10-20 at 2 07 49 pm

I thought Jon Barker's SpaceNet example had smoother gradients (I might be wrong), but moving along the gradient would be very uninteresting in this case. What exactly did you have in mind?

gheinrich:
Hi Joe, when we display the binary segmentation output as an image we get the image you show above with different shades of grey. Suppose you want to make a hard decision whether a given pixel is inside or outside a triangle. In that case you'll need to choose a threshold (below threshold: inside of triangle; above threshold:outside of triangle). I think it would be helpful to display this network output using the binary segmentation extension because the slider allows the user to interactively set the threshold - as in 2-class image segmentation. It's just a thought - I think your PR is already great now.

PR #1190, opened 2016-10-19
Described how to select OpenBLAS / MKL for Caffe build

ajvengo:
Hope this will save time for newcomers like me

### DIFF
--- a/docs/BuildCaffe.md
+++ b/docs/BuildCaffe.md
@@ -49,3 +49,12 @@ cd build
 cmake ..
 make --jobs=4
 ```
+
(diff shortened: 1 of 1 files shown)


TimZaman:
Thanks for the PR. Looks good. However, where do we draw the line between Caffe and DIGITS. Clearly this is a Caffe option and optimization. I'm not sure it's important enough to bother newcomers with. @ajvengo did you add this as just an optimization or was this critical for you?

ajvengo:
@TimZaman, no it is not critical. Just optimization, cause I'm using MAGMA library with MKL.

lukeyeager:
I think we'll keep the instructions simple for now. As I understand it, Atlas vs. OpenBLAS vs. MKL shouldn't make much of a difference as long as CUDA + cuDNN is used.

Thanks though!

PR #1191, opened 2016-10-19
check for custom_classes

jmancewicz:
There's an error reading classification models from before custom_classes was added.

### DIFF
--- a/digits/extensions/data/objectDetection/data.py
+++ b/digits/extensions/data/objectDetection/data.py
@@ -32,7 +32,7 @@ class DataIngestion(DataIngestionInterface):
         # this instance is automatically populated with form field
         # attributes by superclass constructor
 
-        if self.custom_classes != '':
+        if 'custom_classes' in self.__dict__ and self.custom_classes != '':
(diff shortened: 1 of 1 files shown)


jmancewicz:
@gheinrich, do you have any problem with this check?

digits/extensions/data/objectDetection/data.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, **kwargs):
         # this instance is automatically populated with form field
         # attributes by superclass constructor
 
-        if self.custom_classes != '':
+        if 'custom_classes' in self.__dict__ and self.custom_classes != '':
```
lukeyeager: I think you want this instead:
```py
hasattr(self, 'custom_classes')
```

gheinrich:
> @gheinrich, do you have any problem with this check?

None at all

digits/extensions/data/objectDetection/data.py
jmancewicz: I merged before I saw this. Making new PR.

PR #1201, opened 2016-10-21
Check JS lint with gjslint

lukeyeager:
*Working on https://github.com/NVIDIA/DIGITS/issues/1153*

This change should be purely stylistic - entirely non-functional. I've broken it up into multiple commits in an attempt to make this easier to review.


jmancewicz:
This is beautiful.

.gjslintrc
```diff
@@ -0,0 +1,3 @@
+--max_line_length=120
+--exclude_directories=3rdparty
+--disable=0121,0220
```
jmancewicz: Thanks for 0121.

lukeyeager: [The 16.04 package](http://packages.ubuntu.com/xenial/closure-linter) doesn't even check for this rule. I only had to turn it off for [the 14.04 package](http://packages.ubuntu.com/trusty/closure-linter).

PR #1204, opened 2016-10-21
Use hasattr when checking for custom_classes

jmancewicz:
@lukeyeager commented on #1191, and I merged it before seeing the comment.


lukeyeager:
Thanks. Not a big deal, but it's more consistent this way. And I think it's also a bit more readable.

jmancewicz:
Consistent is good.

PR #1207, opened 2016-10-22
Fix remaining Python lint

lukeyeager:
*Working on https://github.com/NVIDIA/DIGITS/issues/1153*


lukeyeager:
The only thing I can imagine worse than implementing this PR is reviewing it... any takers? :stuck_out_tongue_closed_eyes: 

I think our best strategy may be to just merge it and deal with any bugs we run into as they come up.

PR #1223, opened 2016-10-27
Use view extension for Exploring the databases (not ready to merge)

jmancewicz:
This is a move to use the view extensions to explore the databases.

As it is, it breaks the original image classification exploration.

data extensions need two methods, `can_explore`, which is used to enable the "explore the db" button, and `get_data` which retrieves data from the databases, and returns it in the same format that inference would.

Included in this PR is the imageSegmentation data view extension changes.

screen shot 2016-10-27 at 10 46 28 am


jmancewicz:
@gheinrich can you have a look at this. I'm moving on the the other data extensions, and wanted to get your feedback before I finish them.

digits/extensions/data/interface.py
```diff
@@ -109,3 +109,27 @@ def itemize_entries(self, stage):
         this function in no particular order
         """
         raise NotImplementedError
+
+    @staticmethod
+    def can_explore():
```
gheinrich: why don't you have `get_data()` return `None` by default and check the return value on caller side? This way, sub-classes only have to override `get_data()`

digits/extensions/view/imageSegmentation/view.py
```diff
@@ -36,14 +36,14 @@ def __init__(self, dataset, **kwargs):
 
         # view options
         if kwargs['colormap'] == 'dataset':
-            if not COLOR_PALETTE_ATTRIBUTE in dataset.extension_userdata or \
+            if COLOR_PALETTE_ATTRIBUTE not in dataset.extension_userdata or \
```
gheinrich: whoops, thanks!

digits/templates/datasets/generic/show.html
```diff
@@ -52,6 +99,19 @@
         {% endif %}
     </dl>
 </div>
+<script>
+    $('#view_extension_id_{{ index }}').on("change", function(){
+        var view_extension_id= $(this).val();
+        var href = "{{url_for('digits.dataset.generic.views.explore')}}" +
```
gheinrich: can't you use the same syntax as in line 82 above?

digits/extensions/data/interface.py
jmancewicz: I didn't because get_data is expensive to compute and the data would be thrown away in that context.

jmancewicz: Also, when all the view extensions are converted, then the can_explore might just go away.

PR #1229, opened 2016-10-31
[Packaging] Upload to PPA from TravisCI

lukeyeager:
Example build: https://travis-ci.org/lukeyeager/DIGITS/jobs/172138202
Packages uploaded to: https://launchpad.net/~nvidia-digits/+archive/ubuntu/dev/+packages

I haven't tested tagged builds yet, but I think they should be pretty fool-proof.

PR #1230, opened 2016-11-01
[Packaging] Upload to PyPI from TravisCI

lukeyeager:
*Depends on https://github.com/NVIDIA/DIGITS/pull/1229*

We can't test this for real until we tag something.

Example build: https://travis-ci.org/lukeyeager/DIGITS/jobs/172173617
Example package: https://testpypi.python.org/pypi?name=digits&:action=display


lukeyeager:
Rebased on `master` after #1229 was merged, and fixed a bug with source distributions by adding `requirements*.txt` to MANIFEST.in.

PR #1236, opened 2016-11-02
Fix calculation of Dice metric in medical imaging example

gheinrich:
The Dice metric should be computed for each sample and then averaged across the samples of a batch.
The previous implementation was computing the Dice metric for a full batch, which led to slightly incorrect results.


gheinrich:
@IsaacYangSLA can you review please?

PR #1244, opened 2016-11-06
Fix Caffe search paths in Windows

lunzueta:
This pull request fixes the problem that happens in Windows platforms when DIGITS tries to find where Caffe is installed, based on CAFFE_ROOT environment variable. Currently, the official Windows Caffe version is (https://github.com/bvlc/caffe/tree/windows), in which once the installation folder is generated (typically it should be a folder called "install" within the Caffe root folder), the caffe.exe executable is in the folder called "bin" and not in "build/tools" and pycaffe files are in the "python" folder inside it.

### DIFF
--- a/digits/config/caffe.py
+++ b/digits/config/caffe.py
@@ -19,8 +19,12 @@ def load_from_envvar(envvar):
     """
     value = os.environ[envvar].strip().strip("\"' ")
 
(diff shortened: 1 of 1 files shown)


lukeyeager:
Thanks!

PR #1262, opened 2016-11-11
Fix batch accumulation

### LINKED ISSUE #1240: Incorrect iteration_to_epoch calculation with batch_accumulation
Hello,
I'm not a 100% sure, but I think the calculation of iteration_to_epoch is incorrect when using batch_accumulation. I guess this is due to the calculation of solver.max_iter since it doesn't take batch_accumulation into account.

lukeyeager:
*Fix #1240*

### DIFF
--- a/digits/dataset/images/classification/test_imageset_creator.py
+++ b/digits/dataset/images/classification/test_imageset_creator.py
@@ -15,22 +15,18 @@ import numpy as np
 import PIL.Image
 
 
-IMAGE_SIZE = 10
-IMAGE_COUNT = 10  # per category
-
-
-def create_classification_imageset(folder, image_size=None, image_count=None, add_unbalanced_category=False):
+def create_classification_imageset(
+        folder,
+        image_size=10,
+        image_count=10,
+        add_unbalanced_category=False,
+):
     """
     Creates a folder of folders of images for classification
 
     If requested to add an unbalanced category then a category is added with
     half the number of samples of other categories
     """
-    if image_size is None:
-        image_size = IMAGE_SIZE
-    if image_count is None:
-        image_count = IMAGE_COUNT
-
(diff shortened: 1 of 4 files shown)


kostinalexey:
You need to modify following lines to fix test and display parameters for caffe

542-544, 
-            solver.test_iter.append(int(math.ceil(float(self.dataset.get_entry_count(constants.VAL_DB)) / val_data_layer.data_param.batch_size)))
+            solver.test_iter.append(int(math.ceil(float(self.dataset.get_entry_count(constants.VAL_DB)) / (val_data_layer.data_param.batch_size * self.batch_accumulation))))
 
603, 
-                int(math.ceil(5000.0 / train_data_layer.data_param.batch_size))
+                int(math.ceil(5000.0 / (train_data_layer.data_param.batch_size * self.batch_accumulation)))

772-773
-            solver.test_iter.append(int(math.ceil(float(self.dataset.get_entry_count(constants.VAL_DB)) / val_image_data_layer.data_param.batch_size)))
+            solver.test_iter.append(int(math.ceil(float(self.dataset.get_entry_count(constants.VAL_DB)) / (val_image_data_layer.data_param.batch_size * self.batch_accumulation))))

828
-                int(math.ceil(5000.0 / train_image_data_layer.data_param.batch_size))
+                int(math.ceil(5000.0 / (train_image_data_layer.data_param.batch_size * self.batch_accumulation)))

lukeyeager:
@kostinalexey as far as I can tell, batch accumulation has no effect during the TEST phase. Do you have proof to the contrary?

You're right about the display bug though - thanks! Let me fix that...

kostinalexey:
You are right, test phase is not effected by batch_accumulation. My mistake.

lukeyeager:
No problem at all - thanks for double-checking my work!

PR #1270, opened 2016-11-14
Default to PNG in HTML embeddings

### LINKED ISSUE #1200: Quality in inference intermediate results
While testing autoencoders for image processing, I noticed that some typical artifacts of JPEG compression were appearing in the results.
I couldn't find an option to fix this, what I did is modify the embed_image_html to produce PNGs by default. Here's the diff:

```
@@ -300,7 +300,7 @@ def embed_image_html(image):
     fmt = image.format
     if not fmt:
         # default to JPEG
-        fmt = 'jpeg'
+        fmt = 'png'
     else:
         fmt = fmt.lower()

```

I add here request for image processing: could you put a flag somewhere in the infer_one to show results as an image instead of as an array? And (if this last flag is selected) an option to avoid renormalization of the image?

gheinrich:
fix #1200

### DIFF
--- a/digits/utils/image.py
+++ b/digits/utils/image.py
@@ -311,8 +311,8 @@ def embed_image_html(image):
     # Read format from the image
     fmt = image.format
     if not fmt:
-        # default to JPEG
-        fmt = 'jpeg'
+        # default to PNG
+        fmt = 'png'
     else:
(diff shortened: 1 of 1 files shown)


gheinrich:
Verified using sample images from @pansk

PR #1275, opened 2016-11-15
Import pyplot explicitly

### LINKED ISSUE #1117: Segmentation Inference - 'module' object has no attribute 'pyplot'
Hi,
when i want to infer a segmentation image and choose colormap from dataset, i get the following error:
```
 AttributeError

'module' object has no attribute 'pyplot'
    rv = self.dispatch_request()
  File "/home/udo/programs/digits/digits/webapp.py", line 72, in decorated
    return f(_args, *_kwargs)
  File "/home/udo/programs/digits/digits/model/images/generic/views.py", line 394, in infer_one
    outputs)
  File "/home/udo/programs/digits/digits/model/images/generic/views.py", line 794, in get_inference_visualizations
    extension = extension_class(dataset, **extension_form.data)
  File "/home/udo/programs/digits/digits/extensions/view/imageSegmentation/view.py", line 44, in **init**
    cmap = mpl.colors.ListedColormap(palette)
AttributeError: 'module' object has no attribute `'pyplot'
```
matplotlib seems installed.
Best Regards,
Udo

gheinrich:
fix #1117

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -124,6 +124,8 @@ before_install:
     - source ~/venv/bin/activate
 
 install:
+    - mkdir -p ~/.config/matplotlib
+    - echo "backend:agg" > ~/.config/matplotlib/matplotlibrc
     - ./scripts/travis/install-caffe.sh $CAFFE_ROOT
     - if [ "$DIGITS_TEST_FRAMEWORK" == "torch" ]; then travis_wait ./scripts/travis/install-torch.sh $TORCH_ROOT; else unset TORCH_ROOT; fi
     - pip install -r ./requirements.txt
--- a/digits/dataset/generic/views.py
+++ b/digits/dataset/generic/views.py
@@ -10,6 +10,7 @@ except ImportError:
 import caffe_pb2
 import flask
 import matplotlib as mpl
+import matplotlib.pyplot as plt
 import numpy as np
 import PIL.Image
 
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -124,6 +125,7 @@ before_install:
     - source ~/venv/bin/activate
 
 install:
+    - echo "backend:agg" > $MATPLOTLIBRC/matplotlibrc
```
lukeyeager: It might clutter up the log a bit less to do this without an envvar:
```sh
mkdir -p ~/.config/matplotlib
echo "backend: agg" > ~/.config/matplotlib/matplotlibrc
```
Looks like there are several ways to set the config: http://matplotlib.org/users/customizing.html#the-matplotlibrc-file

digits/dataset/generic/views.py
```diff
@@ -10,6 +10,7 @@
 import caffe_pb2
 import flask
 import matplotlib as mpl
+import matplotlib.pyplot as plt
```
lukeyeager: FYI, I learned recently that you can do this:
```py
>>> import matplotlib as mpl
>>> mpl.pyplot
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'pyplot'
>>> import matplotlib.pyplot
>>> mpl.pyplot
<module 'matplotlib.pyplot' from '/usr/lib/python2.7/dist-packages/matplotlib/pyplot.pyc'>
```
That would let you change less code, but I think it's debatable whether that's more legible. Up to you.

gheinrich: Nice! I tried but it seems Lint doesn't know that we can do that (and complains `F401 'matplotlib.pyplot' imported but unused`) so I think I'll leave it as is.

.travis.yml
gheinrich: Thanks I have updated the patch with your suggestion

digits/dataset/generic/views.py
lukeyeager: Oh. Darn linter.

PR #1307, opened 2016-11-29
[Caffe] Fix batch accumulation bug (digits-5.0)

### LINKED ISSUE #1240: Incorrect iteration_to_epoch calculation with batch_accumulation

lukeyeager:
This is a rebase of #1262 (which fixed #1240) for the `digits-5.0` branch.


TimZaman:
LGTM ;)

PR #1323, opened 2016-12-03
Update Object Detection example doc

jmancewicz:
Updated the example README.md with respect to changes in the view options


jmancewicz:
@gheinrich or @lukeyeager can you review this?

digits/extensions/data/objectDetection/data.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, **kwargs):
         # this instance is automatically populated with form field
         # attributes by superclass constructor
 
-        if self.custom_classes != '':
+        if hasattr(self, 'custom_classes') and self.custom_classes != '':
```
lukeyeager: Changing the code at this point in the release cycle is a big deal. Updating the documentation is not.

This is a rebase of https://github.com/NVIDIA/DIGITS/pull/1191 (and https://github.com/NVIDIA/DIGITS/pull/1204), correct? If this is unlikely to affect most users (and I think it is), then I think we should leave it out of this release. It's too late.

jmancewicz: Removed

lukeyeager:
Cherry-picked into `master`.

Issue #1256, opened 2016-11-11
Tests require ImageGradient plugins


DarylWM:
I'm running digits-test on the v5.1-dev release (having built Caffe using the NVcaffe instructions), but I get this failure. Can anyone tell me what might cause that?
```
======================================================================
ERROR: test suite for <class 'digits.dataset.generic.test_views.TestImageGradientCreated'>
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/nose/suite.py", line 209, in run
    self.setUp()
  File "/usr/local/lib/python2.7/dist-packages/nose/suite.py", line 292, in setUp
    self.setupContext(ancestor)
  File "/usr/local/lib/python2.7/dist-packages/nose/suite.py", line 315, in setupContext
    try_run(context, names)
  File "/usr/local/lib/python2.7/dist-packages/nose/util.py", line 471, in try_run
    return func()
  File "/home/daryl/DIGITS-5.1-dev/digits/dataset/generic/test_views.py", line 375, in setUpClass
    image_height=cls.IMAGE_HEIGHT)
  File "/home/daryl/DIGITS-5.1-dev/digits/dataset/generic/test_views.py", line 183, in setUpClass
    cls.dataset_id = cls.create_dataset(json=True, **kwargs)
  File "/home/daryl/DIGITS-5.1-dev/digits/dataset/generic/test_views.py", line 89, in create_dataset
    'Dataset creation failed with %s' % rv.status_code)
RuntimeError: Dataset creation failed with 500
-------------------- >> begin captured stdout << ---------------------
WARNING: log_file config option not found - no log file is being saved
2016-11-11 12:01:55 [INFO ] Loaded 0 jobs.

--------------------- >> end captured stdout << ----------------------
-------------------- >> begin captured logging << --------------------
engineio: INFO: Server initialized for gevent.
digits: INFO: Loaded 0 jobs.
--------------------- >> end captured logging << ---------------------
```

lukeyeager:
Thanks for reporting this. We're aware of it. You should be able to solve it by installing the ImageGradient plugins:
```sh
pip install -e .
pip install -e plugins/data/imageGradients
pip install -e plugins/view/imageGradients
```
That's why our TravisCI tests pass:
https://github.com/NVIDIA/DIGITS/blob/v5.0.0-rc.1/scripts/travis/install-digits.sh#L15-L17

DarylWM:
[...]

PR #1337, opened 2016-12-07
[Tests] Skip if extension not installed (part 2)

lukeyeager:
https://github.com/NVIDIA/DIGITS/pull/1263 was incomplete.

TravisCI build without plugins and **without** this fix (fails):
https://travis-ci.org/lukeyeager/DIGITS/builds/182047123
TravisCI build without plugins and **with** this fix (passes):
https://travis-ci.org/lukeyeager/DIGITS/builds/182050494

/cc @cliffwoolley

PR #1338, opened 2016-12-07
[TravisCI] Test all plugins and no plugins

lukeyeager:
To catch bugs like https://github.com/NVIDIA/DIGITS/issues/1256


lukeyeager:
The TravisCI build correctly detected that the current testsuite is busted:
https://travis-ci.org/NVIDIA/DIGITS/builds/182060877

Need to merge https://github.com/NVIDIA/DIGITS/pull/1337. Once I rebase this on top of that, all tests should pass.

lukeyeager:
@gheinrich do you have an issue with this? Seems like the most reasonable way to test both with and without plugins, while not adding a bunch of new test targets.

If you're cool with it, please merge.

gheinrich:
That looks good to me thanks!

PR #1347, opened 2016-12-12
Add steps to specify the Python layer file

IsaacYangSLA:
Provide more details about how to specify the path of Python layer file when using pretrained models.

### DIFF
--- a/docs/ModelStore.md
+++ b/docs/ModelStore.md
@@ -21,6 +21,7 @@ Click on the small download icon (first column) to download a model from the sto
 
 After downloading a model, you can return to the homepage to see the list of pretrained models.
 To make use of the downloaded model for training, select it as a "Pretrained Network" in the "New Model" form.
+> NOTE: In order to automatically popluate to Python Layers the path of the Python layer file which comes with "Pretrained Models," you need to click "Customize" next to the selected "Pretrained Model."
 
 ## Create your own Model Store server
 


docs/ModelStore.md
```diff
@@ -21,6 +21,10 @@ Click on the small download icon (first column) to download a model from the sto
 
 After downloading a model, you can return to the homepage to see the list of pretrained models.
 To make use of the downloaded model for training, select it as a "Pretrained Network" in the "New Model" form.
+Please remember to specify the path of the customized Python layer file if it is required by the downloaded model.
+You are welcome to provide your own Python layer file by uploading it from your browser (Use client-side file).
+Alternatively, you may click "Customize" next to the "Pretrained Model" you select.
+This will automatically populate the path of the Python layer file to the input field of "Python Layers."
```
lukeyeager: I don't think we need to explain about Python layers in general - we already have a doc for that at `examples/python-layer/`. There's not really any alternative, right? The only way to get the Python layer that comes with the model is to click "Customize". This is very odd and I think deserves some sort of special formatting like:

> *NOTE*: message

(see the note about GPU memory for detectnet [here](https://github.com/NVIDIA/DIGITS/tree/digits-5.0/examples/object-detection#training-detectnet-in-digits))

IsaacYangSLA: Yes.  The special NOTE is a better way to describe that odd operating steps.  I am updating the document.

lukeyeager:
Cherry-picked into `master`

PR #1405, opened 2017-01-20
Fix Siamese example

gheinrich:
Broadcast -1 into all elements that equal 0 in original label.


lukeyeager:
Cherry-picked into `digits-5.0`.

lukeyeager:
Bug 200257814

PR #1417, opened 2017-01-26
Fix documentation typo. 

JacoCronje:
train.txt and test.txt was swapped and shown in the wrong folders for mnist and cifar10 data sets.


lukeyeager:
Thanks!

PR #1422, opened 2017-01-28
Document a cuDNN workaround for the text classification example

lukeyeager:
Bug 200271785

### DIFF
--- a/examples/text-classification/README.md
+++ b/examples/text-classification/README.md
@@ -83,6 +83,23 @@ After a few hours of training, your network loss and accuracy may look like:
 
 ![loss](dbpedia-loss.png)
 
+**NOTE**: There is a minor bug in cuDNN v5.1.10 that you may run into with this example.
+On Kepler GPUs (e.g. K80), you may have very bad performance (time estimates in the hundreds of days!).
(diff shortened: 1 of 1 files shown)


examples/text-classification/README.md
```diff
@@ -83,6 +83,23 @@ After a few hours of training, your network loss and accuracy may look like:
...
+You can override this easily - simply replace this line:
+```lua
+    -- [(1018-3)/3+1=339] x 256
+    net:add(backend.TemporalConvolution(256, 256, 7))
+```
+With these lines:
+```lua
+     -- [(1018-3)/3+1=339] x 256
+    local conv2 = backend.TemporalConvolution(256, 256, 7)
+    conv2:setMode(nil, 0, nil)
```
gheinrich: `setMode()` is undefined for the `nn` backend (and calling it would yield an exception) but since this note specifically states that this patch applies to CuDNN v.5 then I believe it's OK.

lukeyeager: Ah, thanks for point this out. So it's a good thing we're putting it in the documentation and not into the example model directly.

lukeyeager:
Cherry-picked into `master`.

PR #1397, opened 2017-01-17
Add optional prefix for all URLs

lukeyeager:
*Close https://github.com/NVIDIA/DIGITS/issues/1395*

Updating the test suite to allow `DIGITS_URL_PREFIX=/custom-prefix ./digits-test` would be a big undertaking, so I haven't done it yet.

PR #1434, opened 2017-02-03
socket.io prefix is not required

neerajshukla1911:
socket.io prefix is not required

### DIFF
--- a/packaging/deb/extras/digits.nginx-site
+++ b/packaging/deb/extras/digits.nginx-site
@@ -20,7 +20,7 @@ server {
 
     # Socketio
     location /socket.io {
-        proxy_pass http://127.0.0.1:34448/socket.io;
+        proxy_pass http://127.0.0.1:34448;
         proxy_redirect off;
         proxy_buffering off;
 


lukeyeager:
I just copied that from the Flask-SocketIO documentation:
https://flask-socketio.readthedocs.io/en/latest/#using-nginx-as-a-websocket-reverse-proxy

Are you seeing any problems?

neerajshukla1911:
With current master branch code it works fine. But if digits is run with url-prefix (#1397) code then it throws error since its not correct url. Removing socket.io prefix will make it work in both cases.

lukeyeager:
Yeah but if you're using #1397 then you're probably going to be changing several other things in your nginx config, right?

However, I'm not opposed to merging this if it doesn't break anything for the standard case. When I get back in on Monday I'll test it out.

BTW, I'll need you to send a signed CLA before I can accept this contribution.
https://github.com/NVIDIA/DIGITS/blob/digits-5.0/.github/CONTRIBUTING.md#pull-requests

neerajshukla1911:
Change in this pull request also makes same proxy_pass for both location '/' and '/socket.io'. So further redundancy can be removed using upstream directive like below
upstream digits_server {
        server localhost:34448;
}
This will make config clean. Should I make upstream changes in this pull request only ?
I will send signed CLA today.

lukeyeager:
Why don't you update the PR to show me what you mean. I'm still not convinced anything of the changes you're proposing would change any behavior (unless possibly paired with #1397).

PR #1450, opened 2017-02-10
Correct shebang for prepare_pascal_voc_data.sh

cgerum:
This just a minor change. But the incorrect #/bin/bash creates problems for users of (t)csh


lukeyeager:
Oops. Thanks!.

PR #1498, opened 2017-03-08
Copy labels.txt inside the dataset (Fix #1462)

### LINKED ISSUE #1462: Copy labels.txt inside the dataset
Migrating datasets, created by using text files, from one computer to another is a pain because new model jobs will fail with error:

> [ERROR] AssertionError: labels_file does not exist

Why not save the labels file inside the job folder and then use a relative path?

rodrigoberriel:
This aims to fix #1462 

To test it, I created and moved Image Classification datasets between two machines and everything worked as expected. In addition, I created Classification Models using these datasets in both machines and they worked as expected as well.

The change made in `digits/model/tasks/train.py` was just to make the error more informative.

### DIFF
--- a/digits/dataset/images/classification/views.py
+++ b/digits/dataset/images/classification/views.py
@@ -2,6 +2,7 @@
 from __future__ import absolute_import
 
 import os
+import shutil
 
 # Find the best implementation available
 try:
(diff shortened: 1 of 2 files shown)


digits/dataset/images/classification/views.py
```diff
@@ -156,12 +156,15 @@ def from_files(job, form):
     """
     # labels
     if form.textfile_use_local_files.data:
-        job.labels_file = form.textfile_local_labels_file.data.strip()
+        labels_file_from = form.textfile_local_labels_file.data.strip()
+        labels_file_to = os.path.join(job.dir(), utils.constants.LABELS_FILE)
+        import shutil
```
gheinrich: maybe you can move this import to the top (line 5), where the other standard import packages are

rodrigoberriel:
@gheinrich  Done! Moved the import to the top and squashed.

PR #1500, opened 2017-03-08
Fix Distribution Graph

### LINKED ISSUE #1473: Image Load Failures lead to mismatch between DB entries and Distribution Graph
### tl;dr

The distribution data (used to plot the Distribution Graph) is incorrect when images fail to load.

---
In case of image load failures (I only tested on Classification Datasets), there will be a mismatch between the **DB entries** and the **Distribution Graph** shown in the dataset page.

This happens because **DB entries** use `task.entries_count` that comes from the [number of images written to the database](https://github.com/NVIDIA/DIGITS/blob/master/digits/dataset/tasks/create_db.py#L212), and it takes the number of images loaded into account, i.e. DB entries consider only images loaded. On the other hand, the **Distribution Graph** uses `task.distribution_data()` that comes from [an information](https://github.com/NVIDIA/DIGITS/blob/master/digits/dataset/tasks/create_db.py#L194) that is calculated from [parsing the text file](https://github.com/NVIDIA/DIGITS/blob/master/digits/tools/create_db.py#L506), i.e. the Distribution Graph doesn't take load failures into account.

I think two actions are required:
1. Fix the Distribution Graph to show the correct number of images loaded on each class
2. Show a warning with the number of failures after finishing the job

`LoadImageError` can be thrown for [many reasons](https://github.com/NVIDIA/DIGITS/blob/master/digits/utils/image.py#L40):
- LoadImageError: "{image_path}" not found
- LoadImageError: IOError: image file is truncated (20 bytes not processed)
- LoadImageError: IOError: cannot identify image file
- etc.

I think `not found` errors are unexpected for the user and they should be checked during the initial parse. Also, these errors should stop the job<sup>1</sup>. I've made a test and for a text file with 200k image paths, it increased the time spent on [`_fill_load_queue()`](https://github.com/NVIDIA/DIGITS/blob/master/digits/tools/create_db.py#L465) ~15 seconds. The job itself took ~2 hours. I think it's worth it.

The other ones require `try-except` on `Image.open(...)` which might be too expensive for very long text files, i.e. very large databases. Therefore, the current behavior + displaying a message at the end (warning the user) must be fine.

<sup>1</sup> <sub>I lost ~2 hours of a database creation job to discover that the script I used to generate the text files had a bug and ~20% of the paths were broken. Besides that, I looked at the Distribution Graph and everything looked as I was expecting :(</sub>

---
DIGITS: 5.1-dev (I had these problems when using DIGITS 4 too)
Caffe: 0.15.14
Caffe flavor: NVIDIA

rodrigoberriel:
This aims to fix #1473 

If any `LoadImageError` is raised, dataset pages will look like this:
- notice the `(%d failed to load)` on the **DB Entries**

image

Datasets without any `LoadImageError` will look exactly the same:

image

### DIFF
--- a/digits/dataset/tasks/create_db.py
+++ b/digits/dataset/tasks/create_db.py
@@ -1,7 +1,6 @@
 # Copyright (c) 2014-2017, NVIDIA CORPORATION.  All rights reserved.
 from __future__ import absolute_import
 
-import operator
 import os.path
 import re
 import sys
@@ -61,6 +60,7 @@ class CreateDbTask(Task):
             self.image_channel_order = None
 
         self.entries_count = None
+        self.entries_error = None
         self.distribution = None
         self.create_db_log_file = "create_%s.log" % db_name
 
@@ -98,6 +98,14 @@ class CreateDbTask(Task):
             self.backend = 'lmdb'
         if not hasattr(self, 'compression') or self.compression is None:
             self.compression = 'none'
+
+        if not hasattr(self, 'entries_error'):
+            self.entries_error = 0
+            for key in self.distribution.keys():
+                self.distribution[key] = {
+                    'count': self.distribution[key],
+                    'error_count': 0
+                }
 
     @override
     def name(self):
@@ -169,8 +177,6 @@ class CreateDbTask(Task):
 
     @override
     def process_output(self, line):
-        from digits.webapp import socketio
-
         self.create_db_log.write('%s\n' % line)
         self.create_db_log.flush()
 
@@ -191,19 +197,22 @@ class CreateDbTask(Task):
             if not hasattr(self, 'distribution') or self.distribution is None:
                 self.distribution = {}
 
-            self.distribution[match.group(1)] = int(match.group(2))
+            self.distribution[match.group(1)] = {
+                'count': int(match.group(2)),
+                'error_count': 0
+            }
+            self.update_distribution_graph()
+            return True
 
-            data = self.distribution_data()
-            if data:
-                socketio.emit('task update',
-                              {
-                                  'task': self.html_id(),
-                                  'update': 'distribution',
-                                  'data': data,
-                              },
-                              namespace='/jobs',
-                              room=self.job_id,
-                              )
+        # add errors to the distribution
+        match = re.match(r'\[(.+) (\d+)\] LoadImageError: (.+)', message)
+        if match:
+            self.distribution[match.group(2)]['count'] -= 1
+            self.distribution[match.group(2)]['error_count'] += 1
+            if self.entries_error is None:
+                self.entries_error = 0
+            self.entries_error += 1
+            self.update_distribution_graph()
             return True
 
         # result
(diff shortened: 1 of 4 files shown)


digits/dataset/tasks/create_db.py
```diff
@@ -191,19 +189,19 @@ def process_output(self, line):
...
-                socketio.emit('task update',
-                              {
-                                  'task': self.html_id(),
-                                  'update': 'distribution',
-                                  'data': data,
-                              },
-                              namespace='/jobs',
-                              room=self.job_id,
-                              )
+            self.distribution[match.group(1)] = [int(match.group(2)), 0]
```
gheinrich: I find this a little bit difficult to read. Could you use a dict with two keys for example `{'count': int(match.group(2)), 'error_count':0}`?

digits/dataset/tasks/create_db.py
```diff
@@ -191,19 +189,19 @@ def process_output(self, line):
...
+            self.update_distribution_graph()
+            return True
+
+        # add errors to the distribution
+        match = re.match(r'\[(.+) (\d+)\] LoadImageError: (.+)', message)
+        if match:
+            self.distribution[match.group(2)][0] -= 1
+            self.distribution[match.group(2)][1] += 1
+            if self.entries_error is None:
```
gheinrich: why don't you set `self.entries_error` to 0 initially?

digits/static/css/style.css
```diff
@@ -150,3 +150,7 @@ ul.inline li {
     margin-top:10px;
     margin-bottom:10px;
 }
+
+div.exploration {
+    margin-top: 10px;
+}
```
gheinrich: ha I think there's a missing line return here :-)

digits/dataset/tasks/create_db.py
rodrigoberriel: I tried to mimic the `self.entries_count`, but I can change to init to 0 if you think it is better

digits/static/css/style.css
rodrigoberriel: Done!

digits/dataset/tasks/create_db.py
gheinrich: Ah yes in that case I think you're right. Consistency is preferred!

digits/dataset/tasks/create_db.py
```diff
@@ -301,24 +302,60 @@ def distribution_data(self):
...
+        label_error = 'LoadImageError'
+
+        # backwards compatibility
+        if not hasattr(self, 'entries_error'):
+            self.entries_error = 0
+            for key in self.distribution.keys():
+                self.distribution[key] = {
+                    'count': self.distribution[key],
+                    'error_count': 0
+                }
```
lukeyeager: For consistency, it's better to do this kind of backwards-compatibility upgrade in the `__setstate__` function above

lukeyeager:
Also haven't tested this. Also looks pretty great to me besides the one nitpick - thanks @rodrigoberriel!

digits/dataset/tasks/create_db.py
rodrigoberriel: I agree. Done!

rodrigoberriel:
After this change, I did the following tests and they were working:

- Create Dataset using Image Folder
- Create Dataset using Image Files (without failures)
- Create Dataset using Image Files (with failures)
- Create Dataset on another machine (without this changes) and moved the job to my machine (job was loaded properly and the backwards-compatibility upgrade worked)

PR #1511, opened 2017-03-15
Enable selection of GPU on inference page

### LINKED ISSUE #1418: Define which GPU for inference?
Hi,

How can we define which GPU to use in inference? 
(same way as we can do for training)

It seems that digits always tries to use GPU0 even in the case where GPU0 is busy but GPU1 is free.

rodrigoberriel:
This PR attempts to fix #1418, see https://github.com/NVIDIA/DIGITS/issues/1418#issuecomment-285680223. I tried to minimize changes to avoid side-effects.

If the user doesn't have more than one GPU (or only one CUDA_VISIBLE_DEVICES for DIGITS), nothing is going to change:

image

But if the user has multiple GPUs (2 or more), it is going to look like this:

image

where `Next Available` is the default choice and has the same behavior as the current implementation. If the user selects a GPU (only one can be selected), then this GPU will be used.

PS: I modified [images/generic/show.html](https://github.com/NVIDIA/DIGITS/compare/master...rodrigoberriel:select_gpu_inference?expand=1#diff-f16ada5c581ed885973db1abdde5fae5) the same way I did for [images/classification/show.html](https://github.com/NVIDIA/DIGITS/compare/master...rodrigoberriel:select_gpu_inference?expand=1#diff-41696ab7b1c04e66a23ac76fd94bb75f), but I didn't test the former as much as I did on the latter.

### DIFF
--- a/digits/model/forms.py
+++ b/digits/model/forms.py
@@ -339,6 +339,23 @@ class ModelForm(Form):
         tooltip="The job won't start until all of the chosen GPUs are available."
     )
 
+    # Select 1 of several GPUs
+    select_one_of_gpus = utils.forms.SelectField(
+        'Select which GPU you would like to use',
+        choices=[('next', 'Next available')] + [(
+            index,
+            '#%s - %s (%s memory)' % (
+                index,
+                get_device(index).name,
+                sizeof_fmt(
+                    get_nvml_info(index)['memory']['total']
+                    if get_nvml_info(index) and 'memory' in get_nvml_info(index)
+                    else get_device(index).totalGlobalMem)
+            ),
+        ) for index in config_value('gpu_list').split(',') if index],
+        default='next',
+    )
+
     # XXX For testing
     # The Flask test framework can't handle SelectMultipleFields correctly
     select_gpus_list = wtforms.StringField('Select which GPU[s] you would like to use (comma separated)')
--- a/digits/model/forms.py
+++ b/digits/model/forms.py
@@ -305,55 +305,31 @@ class ModelForm(Form):
                 if filename and not os.path.exists(filename):
                     raise validators.ValidationError('File "%s" does not exist' % filename)
 
+    # List of GPUs
+    gpu_list = [(
+        index,
+        '#%s - %s (%s memory)' % (
+            index,
+            get_device(index).name,
+            sizeof_fmt(
+                get_nvml_info(index)['memory']['total']
+                if get_nvml_info(index) and 'memory' in get_nvml_info(index)
+                else get_device(index).totalGlobalMem)
+        ),
+    ) for index in config_value('gpu_list').split(',') if index]
+
     # Select one of several GPUs
     select_gpu = wtforms.RadioField(
         'Select which GPU you would like to use',
-        choices=[('next', 'Next available')] + [(
-            index,
-            '#%s - %s (%s memory)' % (
-                index,
-                get_device(index).name,
-                sizeof_fmt(
-                    get_nvml_info(index)['memory']['total']
-                    if get_nvml_info(index) and 'memory' in get_nvml_info(index)
-                    else get_device(index).totalGlobalMem)
-            ),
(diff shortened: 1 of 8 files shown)


digits/model/forms.py
```diff
@@ -339,6 +339,23 @@ def validate_custom_network_snapshot(form, field):
...
+            '#%s - %s (%s memory)' % (
+                index,
+                get_device(index).name,
+                sizeof_fmt(
+                    get_nvml_info(index)['memory']['total']
+                    if get_nvml_info(index) and 'memory' in get_nvml_info(index)
+                    else get_device(index).totalGlobalMem)
+            ),
+        ) for index in config_value('gpu_list').split(',') if index],
+        default='next',
```
lukeyeager: What happens here if there are no GPUs on the system?

rodrigoberriel: @lukeyeager the same behavior that you'd get if you were to start a training job without a GPU:
1. If all setup was made with CPU only, you'd get the same page you currently get (without the multi-gpu form)
2. If you manage to have Caffe compiled with CUDA and has no GPU on the system (suppose you removed it afterwards -- I tested it masking all my GPUs using an invalid id on `CUDA_VISIBLE_DEVICES`), then you'd get the page without multi-gpu form and your job will fail exactly as the train job fails currently:

```
Check failed: error == cudaSuccess (38 vs. 0)  no CUDA-capable device is detected
```

lukeyeager: Sounds reasonable to me!

lukeyeager (approved):

digits/model/images/generic/views.py
```diff
@@ -309,8 +309,10 @@ def show(job, related_jobs=None):
             template, context = extension.get_inference_template(form)
             inference_form_html = flask.render_template_string(template, **context)
 
+    generic_form = GenericImageModelForm()
     return flask.render_template(
         'models/images/generic/show.html',
+        form=generic_form,
```
gheinrich: I am a bit uneasy about passing the form here when all you need is the list of GPUs. Wouldn't it more explicit and self-explanatory to pass the list instead?

digits/templates/models/images/classification/show.html
```diff
@@ -139,9 +137,36 @@ <h4 class='text-center'>Dataset</h4>
...
+                            formenctype="multipart/form-data"
+                            class="btn btn-success">
+                            Make Pretrained Model
+                        </button>
+                        {% endif %}
+                    </div>
+                    {% if show_multi_gpu_form %}
+                    <div class="col-sm-6">
+                        <div class="form-group{{mark_errors([form.select_one_of_gpus])}}">
+                            {{form.select_one_of_gpus.label}}
```
gheinrich: it's indeed better to use the WTForm fields here but it's  inconsistent with the rest of the file... so I'm not sure about that...

gheinrich (commented):
sorry for late response and thanks for another great PR! Do you think you could add some unit tests to verify the new functionality?

digits/model/images/generic/views.py
rodrigoberriel: I agree. Do you think [this way](https://github.com/NVIDIA/DIGITS/pull/1511/commits/019414fd0ae849b7970dbba2c4f08966ed201efe#diff-46bf113ddd823b115faf1a7edd0675d0) is okay?

digits/templates/models/images/classification/show.html
rodrigoberriel: I've [changed](https://github.com/NVIDIA/DIGITS/pull/1511/commits/019414fd0ae849b7970dbba2c4f08966ed201efe#diff-41696ab7b1c04e66a23ac76fd94bb75f) that too. What you think?

rodrigoberriel:
@gheinrich I'm on a deadline (March 27), but I could work on the unit tests after that. I have to say that I am not familiar with unit tests using flask, but I can give it a try.

RSly:
Hi, any update on the merge for this PR?
tnx

PR #1530, opened 2017-03-22
Fix table formatting in README.md

thinkski:
Fix table formatting


lukeyeager:
Yuck, has that been broken the whole time? Maybe GitHub got more strict with their table parsing? Either way, thanks for the fix!

I'll merge once TravisCI passes. Thanks for the CLA.

PR #1531, opened 2017-03-24
Clamp distance values from segementation boundaries before begin conv…

jmancewicz:
…erted to uint8. That was causing banding in the image because of wrapping at V % 256
screen shot 2017-03-23 at 9 20 18 pm

Without
screen shot 2017-03-23 at 9 20 51 pm


gheinrich:
@adnphllps @4575759ww do you mind testing this patch?

adnphllps:
I have tested the patch. Now the inference visualization works as expected. Thanks

Issue #1492, opened 2017-03-04
DIGITS image classification view does not correctly handle output of FCNs


AliaMYH:
I've trained a network and the validation accuracy is very good for it, almost 100 percent. However when I come to 'classify many', using the val.txt for the dataset, the accuracy returned is terrible. Only one of the classes is predicted every time. I'm not exactly sure what the problem is. 
My validation set is 20% of my training set, I've used the split option in digits. 

![screen shot 2017-03-04 at 6 48 01 pm](https://cloud.githubusercontent.com/assets/10338687/23580475/5871427c-010b-11e7-92ba-49f89f98bb13.png)
![screen shot 2017-03-04 at 6 48 14 pm](https://cloud.githubusercontent.com/assets/10338687/23580476/58a23454-010b-11e7-887d-c66a737ca3c4.png)

AliaMYH:
I tried using mean pixel as recommended by https://github.com/NVIDIA/DIGITS/issues/625 , but the same thing is happening.

AliaMYH:
@samansarraf According to your comment on #625, am I meant to preprocess all my data outside of digits before training?

samansarraf:
@AliaMYH your understanding is correct. To solve the issue at that time , I came up with the idea to preprocess my data and made them totally independent from classify many preprocessing module . There should be something on their codes that is not working in some cases. After I preprocessed data outside (including image resizing ...) the classify many exactly produced the same results as I was getting during training and validation . Don't give up ! you are pretty close to it .

AliaMYH:
A curious thing though is that I use the exact same dataset and cropping with two different networks and one gave the above results, while the other one worked perfectly. The one with the above results is a fine-tuned Squeezenet, and the one that gave proper results was a fine-tuned AlexNet which I trained for comparison. I'm not exactly sure why this is. 

Also, does this mean that it's only the displayed results that are incorrect or the entire training of the network.. considering that the 'Classify One' results are correct.  Is the deploy.prototxt what is used during the inference? 

@samansarraf @gheinrich

samansarraf:
@AliaMYH No , the entire training (including validation) process is correct. Only classify many has some preprocessing problem in my opinion. I had the same kind of problem with GoogleNet so I what I did I cropped the data before training and used the cropped images for both training and classify many module then it worked.

AliaMYH:
I agree as well. @gheinrich Can you confirm this?

gheinrich:
[...]

PR #1536, opened 2017-03-27
FCN confusion matrix visualization fix

SlipknotTN:
Issue #1492.

To reproduce the bug:
- Train a fully convolutional network (e.g [SqueezeNet 1.1](https://github.com/DeepScale/SqueezeNet))
- Run "classify many" with a list of images

Depending on the version of numpy you will get a warning or an error.

With numpy 1.11 you get this warning and the confusion matrix has all the results on the first column.
```
VisibleDeprecationWarning: converting an array with ndim > 0 to an index will result in an error in the future
  result.append((labels[i], round(100.0 * scores[image_index, i], 2)))
```
With numpy 1.12 you get an error at the same line and no confusion matrix is shown.

The problem is the shape of "scores" that must be (number_of_images, number_of_classes), but with FCNs the shape has additional dimensions. e.g. SqueezeNet v1.1 final shape is (number_of_images, number_of_classes, 1, 1).


SlipknotTN:
Added the same fix for "TOPN category". Without this fix, topn returns an error about dimensions.

tmatas:
Hi! I'm facing the same issue..any news about this task?

Issue #1545, opened 2017-03-29
Configuration Issues with Digits 5.0 on Ubuntu 16.04


S4WRXTTCS:
I installed Digits 5.0 using the deb based network installer for CUDA 8, and Ubuntu 16.04

It runs correctly, but I'm having issues starting/stopping the service. I'm also having issues reconfiguring it to put the Jobs somewhere else other than the default.

When I try to stop the nvidia-digits-server it comes back with this error message

"stop: Unable to connect to Upstart: Failed to connect to socket /com/ubuntu/upstart: Connection refused"

When I try to make configuration changes according to the UbuntuInstall.md documentation I get this error message when running the sudo python -m digits.config.edit command

/usr/bin/python: No module named digits.config.edit

I did try setting up an environmental variable at startup, and that didn't work either. I believe that's only valid if it's manually installed. Or maybe I didn't put it early enough in the start up.

lukeyeager:
Aha, so our DIGITS 5.0 deb packages finally made it on the public repo! Time to update our docs. These are outdated (still intended for the 4.0 debs): https://github.com/NVIDIA/DIGITS/blob/v5.0.0/docs/UbuntuInstall.md.

In the meantime, please refer to these files to solve your problems:

* https://github.com/NVIDIA/DIGITS/releases/tag/v5.0.0-rc.1
* https://github.com/NVIDIA/DIGITS/blob/digits-5.0/docs/Configuration.md (don't try to use `python -m digits.config.edit` anymore because of https://github.com/NVIDIA/DIGITS/pull/1091)
* https://github.com/NVIDIA/DIGITS/blob/digits-5.0/packaging/deb/extras/custom_http_errors/502.html (you have SystemD on 16.04, not upstart)

lukeyeager:
@S4WRXTTCS will you take a look at #1546? Please let me know if it doesn't give you enough information to fix your problem.

Temporary link: https://github.com/lukeyeager/DIGITS/blob/updated-5.0-debs/docs/UbuntuInstall.md#troubleshooting

S4WRXTTCS:
These instructions work great.

Thanks for your help.

PR #1546, opened 2017-03-30
[Docs] Update for 5.0 debs and Ubuntu 16.04 support

lukeyeager:
Close https://github.com/NVIDIA/DIGITS/issues/1545.


lukeyeager:
@gheinrich I actually ran the same test and found the same issue - thanks for the careful review! I've reported the issue and it should be fixed with the next repo update. For now, it doesn't matter because it's just a warning and the cuda repo package installs the same key that the ml repo package was trying to install.

lukeyeager:
Rebased after merging https://github.com/NVIDIA/DIGITS/pull/1451.

PR #1555, opened 2017-04-03
PDF and latex versions of CLA

samikama:
Hi,

I added PDF version of CLA so that it can be filled directly and contributors don't have to deal with generating pdfs out of it or the formatting issues when text is copied to an editor.

Cheers,
Sami


lukeyeager:
Hi @samikama. Thanks for the PR which attempts to make it easier to contribute! Our CLA process is regrettably low-tech. Unfortunately, I'm not going to merge it for two reasons:

1. I still feel like I'd need to run it by our legal team (even though I wouldn't expect any issues)
2. I prefer not to check binary files into git repos (even though this PDF is pretty small)

We can revisit this if there's a bunch of clamor.

PR #1610, opened 2017-04-28
Previous network clone epoch

jmancewicz:
When cloning a job using a previous network, the epoch is not saved in the job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -121,6 +121,7 @@ before_install:
     - deactivate
     - virtualenv --system-site-packages ~/venv
     - source ~/venv/bin/activate
+    - sudo apt-get install libboost-filesystem1.55-dev libboost-python1.55-dev libboost-system1.55-dev libboost-thread1.55-dev
 
 install:
     - mkdir -p ~/.config/matplotlib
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -121,6 +121,7 @@ before_install:
     - deactivate
     - virtualenv --system-site-packages ~/venv
     - source ~/venv/bin/activate
+    - sudo apt-get install libboost-filesystem1.55-dev libboost-python1.55-dev libboost-system1.55-dev libboost-thread1.55-dev
```
lukeyeager: Why don't these lines do the trick?
https://github.com/NVIDIA/DIGITS/blob/776131dc746f3bd09b685ae32c6d4c4b11c03394/.travis.yml#L86-L89

digits/model/images/classification/views.py
```diff
@@ -204,8 +204,10 @@ def create():
                                     "Pretrained_model for the selected epoch doesn't exist. "
                                     "May be deleted by another user/process. "
                                     "Please restart the server to load the correct pretrained_model details.")
-                        break
 
+                            # terrible hack to save the epoch for cloning
```
lukeyeager: Why is a "terrible hack" necessary?

.travis.yml
lukeyeager: This build was fine, without the updated boost: https://travis-ci.org/NVIDIA/DIGITS/builds/217367405

lukeyeager: This is the change that broke things: https://github.com/BVLC/caffe/pull/5526

Issue #1645, opened 2017-05-19
Classification visualization: softmax normalization


Noiredd:
Is there any reason we don't normalize softmax layers? Every time I run a classification task with visualization, my softmax output is all blue (=zero). At first I thought that is some bug but then I dug into the code and found [these two lines](https://github.com/NVIDIA/DIGITS/blob/4bb688d2a35ee740f3ec6d5a26fa4dd2ae42cf7c/digits/model/tasks/caffe_train.py#L1367-L1368), which mean it is done on purpose. After I commented them out, my visualizations started to make sense (hot pixels are hot, zeros are blue).

Is this really desired behavior? I am ready to provide a pull request if it's just a mistake or overlook.

lukeyeager:
I think I just preferred that method of visualization. I care to see how confident the network is about its top prediction. When you normalize, you lose any information about whether the top value is actually high confidence or not. Plus, IIRC most of the other layers require normalization to fit into a visualize-able range, whereas softmax is guaranteed to output in 0-1.

lukeyeager:
So, maybe make the PR and try to drum up some upvotes? :wink:

Noiredd:
> softmax is guaranteed to output in 0-1

Therein lies the entire problem - softmax is guaranteed to output in 0-1, while the image range is 0-255. So no matter what the net spits out, everything's equally blue (well, at least for me).  
I forgot that simply normalizing will lose information about the absolute magnitudes, you're right that's not the right way to do it. But multiplying by a constant of 255 should be exactly it.

Before I make the PR - are there any other cases where we don't want to normalize? If not, I would assume that the `normalize` arg can be used in the sense that whenever it's false it's safe to assume we're dealing with the softmax layer (and we can skip normalization but do the scaling).

lukeyeager:
If you want to do that, how about this instead:
```py
                            if layer.type == 'Softmax':
                                vis = utils.image.get_layer_vis_square(data * 255,
                                                                       normalize=False,
                                                                       allow_heatmap=bool(top != 'data'),
                                                                       channel_order='BGR')
                            else:
                                vis = utils.image.get_layer_vis_square(data,
                                                                       normalize=normalize,
                                                                       allow_heatmap=bool(top != 'data'),
                                                                       channel_order='BGR')

```

PR #1647, opened 2017-05-22
Fix softmax visualization by scaling to image range

Noiredd:
Fixes issue #1645, where softmax output was left in 0-1 range (as compared to 0-255 image) which prevented proper visualization. See image for before&after comparison. Changes `digits/model/tasks/caffe_train.py` per @lukeyeager's suggestion.

before and after


ethantang95:
Validated and passes through all the tests.

PR #1680, opened 2017-06-14
Fix bounding box colors displayed, for multi class detections

JacoCronje:
Remove sorting of keys which caused a mismatch in colors with multi class detection networks.

The colors used to render the bounding boxes and the colors used to describe each key does not match because the keys get sorted. Removing the sort operation resolved the mismatching color issue.


TimZaman:
LGTM

lukeyeager:
I trust Tim's review :).  @JacoCronje can you send us a signed CLA so I can accept this?

PR #1715, opened 2017-07-07
Windows Installation Doc Updates

ahoehne:
some resources are only available in newer versions.
installation still failed for me, but atleast these modifications get you a step further, maybe someone else is able and willing to complete it.
for my part i will switch to ubuntu, because it seems to be way less painful.


prajy:
Also, the the launching DIGITS command must be changed from 
python digits-devserver 
to
python -m digits

Devserver is not a python script.
There are now also new versions for all the libs from the PR commit.

PR #1731, opened 2017-07-20
Bug fix changes

ethantang95:
Changes WRT to bug fixes.


digits/templates/home.html
```diff
@@ -52,8 +52,8 @@
     {[jc.model_jobs = (jobs | filter:is_model);'']}
     {[jc.pretrained_model_jobs = (jobs | filter:is_pretrained_model);'']}
     <section ng-controller="tab_controller as tab" ng-init="init({{request.args['tab']}})">
-        <div ng-controller="running_controller as c">
-            <div ng-if="jc.running_jobs.length">
+        <div ng-if="jc.running_jobs.length">
```
gheinrich: can you explain this change? Sorry I don't understand what it does.

ethantang95: This fixes the problem where you actually cannot select any running jobs on the home page to abort, delete, or group them.

PR #1732, opened 2017-07-21
Tensorflow: fix network customization bug

lukeyeager:
Without this fix to the regular expression, the `re` module will happily
match `lenet.pyc` in addition to `lenet.py`. If your filesystem happens
to return `lenet.pyc` first, then you get a cryptic unicode error when
you try to customize the LeNet network. Presumably you'd get a nasty
error if you tried to actually train with this file as well.


ethantang95:
Looks good.

PR #1767, opened 2017-08-07
Added instructions of how to visualize tensorflow in some examples

ethantang95:
Also generalized the gradient model for tensorflow.


docs/DockerInstall.md
```diff
@@ -23,7 +23,7 @@ sudo docker pull nvidia/digits.6-rc
 To run DIGITS with Nvidia Docker, use the following command:
 
 ```bash
-sudo docker run -v <path to data>:/data/ -p 5000:5000 nvidia/digits:latest
+sudo nvidia-docker run -v <path to data>:/data/ -p 5000:5000 nvidia/digits:latest
```
lukeyeager: I would remove the `sudo` here. Users could have added themselves to the `docker` group, or they could be running in a root terminal. `sudo` is not always necessary.

PR #1803, opened 2017-08-30
Add URL prefix support

IsaacYangSLA:


lukeyeager:
(was https://github.com/NVIDIA/DIGITS/pull/1397)

lukeyeager:
> Updating the test suite to allow DIGITS_URL_PREFIX=/custom-prefix ./digits-test would be a big undertaking, so I haven't done it yet.

Just a reminder - if you don't update the tests, someone is likely to break this feature without realizing it in the future.

IsaacYangSLA:
We will keep that in mind.

PR #1804, opened 2017-08-30
Prepare for 6.0.0-rc.2

IsaacYangSLA:
Point model store configuration to new URL and change __version__.


IsaacYangSLA:
Close this in favor of updating via direct commit.

PR #1819, opened 2017-09-08
Remove pypi upload portion in CI

IsaacYangSLA:
Pypi upload script is incompatible with current pypi API.  Remove it to avoid CI error.


lukeyeager:
If you're doing that, might as well remove `scripts/travis/pypi-upload.sh`, too.

IsaacYangSLA:
That's right.  That shell script will not be used after removing pypi part of CI codes.

Issue #1888, opened 2017-11-22
404 Error for inference datasets when using DIGITS_URL_PREFIX


mebersole:
When using the DIGITS_URL_PREFIX, you get an error related to /datasets/inference-form/...

This is because of a missing `{{config['URL_PREFIX']}}` in the show.html file

gheinrich:
Fixed by #1889

PR #1889, opened 2017-11-22
Fix 404 error for /datasets/inference-form/... from #1888

mebersole:
This is a fix for issue #1888


mebersole:
@gheinrich this is a fix for a recent issue we found in DIGITS v6.0.1 using DIGITS_URL_PREFIX feature.

gheinrich:
Thanks, I tested that it doesn't break the "normal" case where `DIGITS_URL_PREFIX` isn't set. Also I skimmed through templates and didn't find other cases like the one you fixed.

PR #1967, opened 2018-02-15
Make server host configurable with a cli argument

spMohanty:


spMohanty:
The failed checks seem to be because of errors in the build environment and not because of the code :-?

